PR #4, opened 2018-02-09
Added scheduled backup support

swapnilgm:
- Schedule backup on localhost and s3 bucket


cmd/snapshot/main.go
```diff
@@ -0,0 +1,59 @@
...
+
+	stopCh := setupSignalHandler()
+	command := app.NewCommandStartSnapshotter(stopCh)
+	if err := command.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+// SetupSignalHandler registered for SIGTERM and SIGINT. A stop channel is returned
```
georgekuruvillak: Why do we need 2 traps? One trap to enough to close, perform housekeeping tasks and exit gracefully.

cmd/snapshot/app/snapshot.go
```diff
@@ -0,0 +1,101 @@
...
+	etcdEndpoints   string
+	storageProvider string
+}
+
+// NewCommandStartSnapshotter create cobra command for snapshot
+func NewCommandStartSnapshotter(stopCh <-chan struct{}) *cobra.Command {
+	config := &configuration{}
+	var command = &cobra.Command{
+		Use:   "snapshot",
+		Short: "Backup the tcd periodically.",
```
georgekuruvillak: 'etcd'

cmd/snapshot/app/snapshot.go
```diff
@@ -0,0 +1,101 @@
...
+
+// configuration is struct to hold configuration for utility
+type configuration struct {
+	schedule        string
+	etcdEndpoints   string
+	storageProvider string
+}
+
+// NewCommandStartSnapshotter create cobra command for snapshot
+func NewCommandStartSnapshotter(stopCh <-chan struct{}) *cobra.Command {
```
georgekuruvillak: SnapshotCmd sounds better than NewCommandStartSnapshotter

cmd/snapshot/main.go
swapnilgm: This is standard approach followed in majority projects. In case somebody don't want to wait till cleanup work.

cmd/snapshot/app/snapshot.go
swapnilgm: i'll prefer `NewSnapshotCommand` then.

georgekuruvillak: cool. :)

cmd/snapshot/app/snapshot.go
```diff
@@ -0,0 +1,101 @@
...
+	config := &configuration{}
+	var command = &cobra.Command{
+		Use:   "snapshot",
```
georgekuruvillak: since this is root cmd, give the binary name.

cmd/snapshot/app/snapshot.go
```diff
@@ -0,0 +1,101 @@
...
+type configuration struct {
+	schedule        string
+	etcdEndpoints   string
+	storageProvider string
+}
+
+// NewCommandStartSnapshotter create cobra command for snapshot
+func NewCommandStartSnapshotter(stopCh <-chan struct{}) *cobra.Command {
+	config := &configuration{}
+	var command = &cobra.Command{
```
georgekuruvillak: why dont you keep this as a variable in the package and initialize the flags in init() and in the function return the instance?

swapnilgm: First concern was to pass stopCh.
Regarding init flags, I wanted to read the parameters directly  in configuration object. I'm expecting configuration object will have more members later on, mostly specific to storage providers authentication info. In that case, we might wanna go with global config file option in future. We can surely keep config object also public. But I preferred this way, untill we think of config file, environment variables and may be something like viber to manage it in future.

georgekuruvillak: 1. As with stopCh. We can create a channel in this command context as well instead of passing from main.go.
2. The configuration parameters are passed as flags to the command which makes it specific to this command ryt? So why do we need a global entitiy?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -0,0 +1,134 @@
...
+
+			}
+		}
+	}
+}
+
+// takeFullSnapshot will sotre full snpashot of etcd to snapstore.
+// It basically will connect to etcd. Then ask for snapshot. And finally
+// move store it underlying snapstore on the fly.
+func (ssr *Snapshotter) takeFUllSnapshot() error {
```
georgekuruvillak: takeFullSnapshot

PR #8, opened 2018-03-06
Expose etcd initialiation functionaliy over http

swapnilgm:
* Fixes the data directory validation related issues in last commit.
* Adds HTTP server to expose functionality over http.

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/pkg/types"
+	"github.com/gardener/etcd-backup-restore/pkg/errors"
 	"github.com/gardener/etcd-backup-restore/pkg/initializer"
 	"github.com/gardener/etcd-backup-restore/pkg/server"
 	"github.com/gardener/etcd-backup-restore/pkg/snapshot/restorer"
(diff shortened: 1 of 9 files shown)


cmd/server/server.go
```diff
@@ -0,0 +1,211 @@
...
+				ss,
+				logger,
+				config.maxBackups,
+				time.Duration(config.etcdConnectionTimeout))
+			if err != nil {
+				logger.Fatalf("Failed to create snapshotter: %v", err)
+			}
+			for {
+				logger.Infof("Probing etcd...")
+				select {
```
georgekuruvillak: We need to avoid busy loops.

https://www.reddit.com/r/golang/comments/6ec9o2/effect_of_infinite_loop_on_cpumem/

cmd/server/server.go
```diff
@@ -0,0 +1,211 @@
...
+				default:
+					err = probeEtcd(config.etcdEndpoints, time.Duration(config.etcdConnectionTimeout))
+				}
+				if err == nil {
+					break
+				}
+			}
+			err = ssr.Run(stopCh)
+			if err != nil {
+				logger.Fatalf("Snapshotter failed with error: %v", err)
```
georgekuruvillak: We have a cross container `/metrics` liveliness probe. That means if backup fails, etcd should fail. Ideally, backup should fail when etcd fails, right? So what about we get rid of the cross container liveliness probe and keep the liveliness probe of etcd same as the readiness probe. That way, if backup operation fails, sidecar will start probing etcd (which is basically its liveliness probe). When etcd fails the liveliness probe, the container is restarted and sidecar can act on the etcd failure if needed.

cmd/server/server.go
swapnilgm: As per my understanding, this is not a busy loop. This is polling mechanism. Since we don't get any event from etcd running as separate process, we have to poll for it to check whether it's up or not and also for safe termination we need `stopCh`.  If you are wondering about `default` case behaviour, then please understand the `probeEtcd` call one level deep. Internally, it's a socket or to be precise IO blocking call with timeout. CPU will hit the socket, and wait for interrupt from n/w device. So, indirectly speaking, CPU is ideal in both stopCh case and default case. You can also verify yourself using `top`  by running simple `etcdbrctl server` command (Don't start `etcd` otherwise it will exit out of for).

cmd/server/server.go
swapnilgm: Even i was little doubtful about this probe, we wanted that behaviour from business perspective, where etcd will also fail if backup fails, right? I agree that cause, container restarts, but we have to consider the worst case scenario as well where sidecar stuck in initial phase itself where it fail to start http server. Although i can't think of any sure shot scenario but just for consideration, rare case could be, just not enough CPU space to start sidecar container but the process. I think is call to make whether to have this probe or not. I'm ready to go without it considering it's rare occasion.

PR #12, opened 2018-03-14
GCS object store support.

georgekuruvillak:
This pull request consist of code changes to incorporate support for GCS object store.


pkg/snapstore/gcs_snapstore.go
```diff
@@ -0,0 +1,174 @@
...
+	prefix    string
+	client    *storage.Client
+	bucket    string
+	ctx       context.Context
+}
+
+// NewGCSSnapStore create new S3SnapStore from shared configuration with specified bucket
+func NewGCSSnapStore(bucket, prefix string) (*GCSSnapStore, error) {
+	ctx := context.Background()
+	projectID := os.Getenv(projectid)
```
swapnilgm: Please remove the projectID since we don't use it anywhere currently

example/etcd-statefulset-gcs.yaml
```diff
@@ -0,0 +1,110 @@
...
+        ports:
+        - containerPort: 8080
+          name: server
+          protocol: TCP
+        env:
+            - name: STORAGE_CONTAINER
+              value: <Container Name>      # change storage container name here
+            - name: GOOGLE_APPLICATION_CREDENTIALS
+              value: /root/.gcp/service-account-file.json
+            - name: PROJECT_ID
```
swapnilgm: Please remove the `projectID` since we don't use it anywhere currently. And anyways if we require it, we can read from `service-account-file.json`

pkg/snapstore/utils.go
```diff
@@ -40,6 +40,12 @@ func GetSnapstore(storageProvider, prefix string) (SnapStore, error) {
 			return nil, fmt.Errorf("storage container name not specified")
 		}
 		return NewS3SnapStore(container, prefix)
+	case SnapstoreProviderGCS:
+		container := os.Getenv(envStorageContainer)
```
swapnilgm: How about moving this before `switch`?

swapnilgm:
LGTM. We decided offline to add tests later on.

PR #15, opened 2018-03-22
Azure Blob storage support.

georgekuruvillak:
This PR introduces backup/restore support over Azure blobstorage.


pkg/snapstore/abs_snapshot.go
```diff
@@ -0,0 +1,143 @@
...
+	"fmt"
+	"io"
+	"path"
+	"sort"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+)
+
+const (
+	ABSStorageAccount = "STORAGE_ACCOUNT"
```
swapnilgm: Make it private.

PR #25, opened 2018-04-20
Fix to make sure etcd is not running when the backup container is failing

georgekuruvillak:
Doesnt need to restart the backup container but keep retrying instead keeping etcd down. Please


cmd/server.go
```diff
@@ -96,7 +88,41 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
+					// for this change is that the client and connection becomes in-
+					// valid and we need to re-initialize the snapshotter with a new
+					// client for the cloud provider.
+					logger.Infof("Re-initializing Snapshotter in case of error...")
+					select {
+					case <-stopCh:
+						logger.Info("Shutting down...")
+						return
+					default:
+						// TO DO: check if we need to do a exp. backoff.
```
swapnilgm: This will not crash process at all in case of error. And keep retrying continuously. This might work when it is used as  sidecar in statefulset since there we will get to know in case backup is failing because of liveness probe. But when used as systemd or simple command line process it won't work. So, we should definitely crash it after some retries. 
<br/>I think it should be small change to have exponential retries. Why not do it in same fix.  Something like:
```go
var exponentialRetryTimeout = 1 //(seconds)
for {
     for {
              if err== nil {
                 exponentialRetryTimeout=1
             } else {
                exponentialRetryTimeout << 1
             }
             
            if exponentialRetryTimeout > 16 { 
               fatalf("cloud error")
            } else {
                    select{
                    ...
                   }
            }
      }
      ...
}
```

pkg/server/httpAPI.go
```diff
@@ -68,8 +69,8 @@ func (h *HTTPHandler) Stop() error {
 // ServeHTTP serves the http re
 func (h *HTTPHandler) serveMetrics(rw http.ResponseWriter, req *http.Request) {
```
swapnilgm: Please rename endpoint to `Healthz` instead of `Metrics`.  Please mention it in `CHANGELOG.md` as well

cmd/server.go
```diff
@@ -113,14 +133,40 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
+						logger.Infof("Re-initializing Snapshotter in case of error...")
+						ssr, err = reinitializeSnapshotterWithSnapstore(ssr, snapstoreConfig)
+						if err != nil {
+							logger.Errorf("Re-initializing Snapshotter failed: %s", err.Error())
+							handler.Status = http.StatusInternalServerError
+							return err
+						}
+					}
+
+					err = ssr.Run(stopCh)
```
swapnilgm: Just move the `ssr.Run` at start([line 137](https://github.com/gardener/etcd-backup-restore/pull/25/files#diff-1427b112aa72f64adcf839007e954d56R137)) of function and move reinitialisation logic([line 143](https://github.com/gardener/etcd-backup-restore/pull/25/files#diff-1427b112aa72f64adcf839007e954d56R159) to [line 149](https://github.com/gardener/etcd-backup-restore/pull/25/files#diff-1427b112aa72f64adcf839007e954d56R149)) inside [ `else` case at line 159](https://github.com/gardener/etcd-backup-restore/pull/25/files#diff-1427b112aa72f64adcf839007e954d56R159)

pkg/server/httpAPI.go
```diff
@@ -95,18 +96,24 @@ func (h *HTTPHandler) serveInitialize(rw http.ResponseWriter, req *http.Request)
...
 // ServeInitializationStatus serves the etcd initialization progress status
 func (h *HTTPHandler) serveInitializationStatus(rw http.ResponseWriter, req *http.Request) {
-	//fmt.Fprintf(rw, "initialization request received.")
 	h.initializationStatusMutex.Lock()
 	defer h.initializationStatusMutex.Unlock()
 	h.Logger.Infof("Responding to status request with: %s", h.initializationStatus)
 
-	rw.WriteHeader(http.StatusOK)
-	rw.Write([]byte(h.initializationStatus))
+	rw.WriteHeader(http.StatusAccepted)
```
swapnilgm: It should be http.StatusOk

pkg/server/httpAPI.go
```diff
@@ -95,18 +96,24 @@ func (h *HTTPHandler) serveInitialize(rw http.ResponseWriter, req *http.Request)
...
+	// If the handler status is 202 then the Snapshotter is being initialized.
+	// Return in progress till the snapshotter is being intialized and then
+	// on successful initialization set the initializationStatus to Success.
+	if h.Status == http.StatusAccepted {
```
swapnilgm: Not required. Please do not bring any dependency on `handler.Status` in initialisation API. Initialisation API logic is all good.

pkg/server/httpAPI.go
```diff
@@ -38,14 +38,15 @@ type HTTPHandler struct {
...
+	Status                    int
 }
 
 // RegisterHandler registers the handler for different requests
 func (h *HTTPHandler) RegisterHandler() {
 	h.initializationStatus = "New"
 	http.HandleFunc("/initialization/start", h.serveInitialize)
 	http.HandleFunc("/initialization/status", h.serveInitializationStatus)
-	http.HandleFunc("/metrics", h.serveMetrics)
+	http.HandleFunc("/healthz", h.serveMetrics)
```
swapnilgm: Please rename h.serveMetrics to h.serveHealthz

cmd/server.go
```diff
@@ -76,51 +85,79 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
 				err = ssr.Run(stopCh)
 				if err != nil {
+					handler.Status = http.StatusServiceUnavailable
 					if etcdErr, ok := err.(*errors.EtcdError); ok == true {
-						logger.Errorf("Snapshotter failed with error: %v", etcdErr)
+						logger.Errorf("Snapshotter failed with etcd error: %v", etcdErr)
+
 					} else {
-						logger.Fatalf("Snapshotter failed with error: %v", err)
+						logger.Errorf("Snapshotter failed with error: %v", err)
```
swapnilgm: This should be `Fataf` right? Cause we retried with exponential backoff and still didn't succeeded. In that case it's probably because some config change, may be snapstore config like credentials or tls config and it should be reloaded. So container crash should happen.

pkg/retry/retry.go
```diff
@@ -0,0 +1,39 @@
...
+
+package retry
+
+import (
+	"time"
+)
+
+// Do retries the RetryFunc exponentially with backoff.
+// It is mutated from `retry.Do` function in package
+// [retry-go](https://github.com/avast/retry-go)
```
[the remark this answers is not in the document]
georgekuruvillak: No. I am not using the library. Just got inspired a bit and hacked up the unnecessary parts. :)

georgekuruvillak:
@swapnilgm Thanks for the review. I have made the required changes.

CHANGELOG.md
```diff
@@ -1,8 +1,35 @@
...
+### Docker Image: eu.gcr.io/gardener-project/gardener/etcdbrctl:0.1.0
+
+### Added
+
+- Take snapshot of etcd at periodic interval.
+- Save snapshot to object stores provided by AWS, Azure, GCS, Openstack and also to local disk.
+- Verify data directory of etcd for corruption before bootstrapping.
+- Restore the etcd data directory from previous snapshot. 
+
+[Unreleased]: https://github.com/gardener/etcd-backup-restore/compare/0.1.0...HEAD
```
swapnilgm: https://github.com/gardener/etcd-backup-restore/compare/0.2.2...HEAD

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -79,7 +96,10 @@ func (ssr *Snapshotter) Run(stopCh <-chan struct{}) error {
 			return nil
 		case <-time.After(effective.Sub(now)):
 			ssr.logger.Infof("Taking scheduled snapshot for time: %s", time.Now().Local())
-			err := ssr.takeFullSnapshot()
+			err := retry.Do(func() error {
+				err := ssr.TakeFullSnapshot()
+				return err
```
swapnilgm: Please log the error message here.

pkg/retry/retry.go
```diff
@@ -0,0 +1,39 @@
...
+func Do(retryFunc Func, config *Config) error {
+	var n uint
+	var err error
+	for n < config.Attempts {
+		delayTime := config.Delay * (1 << (n - 1))
```
swapnilgm: Please log here: "Job attempt %d",n

PR #51, opened 2018-08-13
Race condition fix.

georgekuruvillak:
This PR fixes race-conditions that arise when etcd backup-restore occurs.
Fixes #
This PR fixes race-conditions that arise when etcd backup-restore occurs.
- The http handler in the backup-restore should wait till snapshotter has stopped before starting data directory initialization.

**Release note**:

```improvement operator

```

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"sync/atomic"
 	"time"
 
 	"github.com/coreos/etcd/pkg/types"
@@ -38,7 +39,12 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 		Short: "start the http server with backup scheduler.",
 		Long:  `Server will keep listening for http request to deliver its functionality through http endpoins.`,
 		Run: func(cmd *cobra.Command, args []string) {
-
+			var (
+				snapstoreConfig *snapstore.Config
+				ssrStopCh       chan struct{}
+				ackCh           chan struct{}
+			)
+			ackCh = make(chan struct{})
 			clusterUrlsMap, err := types.NewURLsMap(restoreCluster)
 			if err != nil {
 				logger.Fatalf("failed creating url map for restore cluster: %v", err)
@@ -57,7 +63,6 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 				SkipHashCheck:  skipHashCheck,
 			}
 
-			var snapstoreConfig *snapstore.Config
 			if storageProvider != "" {
 				snapstoreConfig = &snapstore.Config{
 					Provider:  storageProvider,
@@ -77,60 +82,60 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 				Status:          http.StatusServiceUnavailable,
 				StopCh:          make(chan struct{}),
 				EnableProfiling: enableProfiling,
+				ReqCh:           make(chan struct{}),
+				AckCh:           make(chan struct{}),
 			}
-			logger.Info("Regsitering the http request handlers...")
+			logger.Info("Registering the http request handlers...")
 			handler.RegisterHandler()
 			logger.Info("Starting the http server...")
 			go handler.Start()
 			defer handler.Stop()
 
-			ssrStopCh := make(chan struct{})
-			go func() {
-				for {
-					var s struct{}
-					select {
-					case <-handler.StopCh:
-					case <-stopCh:
-					}
-					ssrStopCh <- s
-				}
-			}()
-
 			if snapstoreConfig == nil {
 				logger.Warnf("No snapstore storage provider configured. Will not start backup schedule.")
+				go handleNoSsrRequest(handler)
 				handler.Status = http.StatusOK
 				<-stopCh
+				logger.Infof("Received stop signal. Terminating !!")
 				return
 			}
 
+			tlsConfig := snapshotter.NewTLSConfig(
+				certFile,
+				keyFile,
+				caFile,
+				insecureTransport,
+				insecureSkipVerify,
+				etcdEndpoints)
+
+			ss, err := snapstore.GetSnapstore(snapstoreConfig)
+			if err != nil {
+				logger.Fatalf("Failed to create snapstore from configured storage provider: %v", err)
+			}
+			logger.Infof("Created snapstore from provider: %s", storageProvider)
+
+			snapshotterConfig, err := snapshotter.NewSnapshotterConfig(
+				schedule,
+				ss,
+				maxBackups,
+				deltaSnapshotIntervalSeconds,
+				time.Duration(etcdConnectionTimeout),
+				time.Duration(garbageCollectionPeriodSeconds),
+				garbageCollectionPolicy,
+				tlsConfig)
+			if err != nil {
+				logger.Fatalf("failed to create snapstore config: %v", err)
+			}
+			logger.Infof("Creating snapshotter...")
+			ssr := snapshotter.NewSnapshotter(
+				logger,
+				snapshotterConfig)
+
+			ssrStopCh = make(chan struct{})
+			go handleSsrRequest(handler, ssr, ackCh, ssrStopCh, stopCh)
+			go handleAckState(handler, ackCh)
+
 			for {
-				ss, err := snapstore.GetSnapstore(snapstoreConfig)
-				if err != nil {
-					logger.Fatalf("Failed to create snapstore from configured storage provider: %v", err)
-				}
-				logger.Infof("Created snapstore from provider: %s", storageProvider)
-
-				tlsConfig := snapshotter.NewTLSConfig(
-					certFile,
-					keyFile,
-					caFile,
-					insecureTransport,
-					insecureSkipVerify,
-					etcdEndpoints)
-				ssr, err := snapshotter.NewSnapshotter(
-					schedule,
-					ss,
-					logger,
-					maxBackups,
-					deltaSnapshotIntervalSeconds,
-					time.Duration(etcdConnectionTimeout),
-					time.Duration(garbageCollectionPeriodSeconds),
-					garbageCollectionPolicy,
-					tlsConfig)
-				if err != nil {
-					logger.Fatalf("Failed to create snapshotter from configured storage provider: %v", err)
-				}
-
 				logger.Infof("Probing etcd...")
 				select {
 				case <-stopCh:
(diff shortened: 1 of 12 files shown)


cmd/server.go
```diff
@@ -145,45 +181,43 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
+					snapshotterConfig,
+					&ackCh)
+				if err != nil {
+					logger.Fatalf("Failed to create snapshotter from configured storage provider: %v", err)
 				}
 
+				go func() {
+					for {
+						select {
+						case ssrHealth, ok := <-ssr.HealthCh:
```
swapnilgm: Do we really need HealthCh?  I think, you can convert this go routine to simple function `UpdateHealth(bool)` And call where ever required.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -293,33 +346,41 @@ func (ssr *Snapshotter) processWatch(client *clientv3.Client) {
...
 			return
 
 		case wr, ok := <-watchCh:
 			if !ok {
 				ssr.logger.Warnln("watch channel closed")
 				ssr.fullSnapshotCh <- time.Now()
-				return
-			}
-			if err := wr.Err(); err != nil {
+
```
swapnilgm: Remove empty line.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -293,33 +346,41 @@ func (ssr *Snapshotter) processWatch(client *clientv3.Client) {
...
 				ssr.fullSnapshotCh <- time.Now()
-				return
-			}
 
-			//aggregate events
-			for _, ev := range wr.Events {
-				currentRev = ev.Kv.ModRevision
-				events = append(events, newEvent(ev))
+			} else {
+
```
swapnilgm: Remove empty line.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -328,13 +389,17 @@ func (ssr *Snapshotter) processWatch(client *clientv3.Client) {
 				if err := ssr.saveDeltaSnapshot(events, currentRev); err != nil {
 					ssr.logger.Errorf("failed to take new delta snapshot: %s", err)
 					ssr.fullSnapshotCh <- time.Now()
-					return
+
```
swapnilgm: Remove empty line.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -328,13 +389,17 @@ func (ssr *Snapshotter) processWatch(client *clientv3.Client) {
...
+				} else {
+					timer.Reset(time.Second * time.Duration(ssr.config.deltaSnapshotIntervalSeconds))
+					events = []*event{}
 				}
-				events = []*event{}
+
```
swapnilgm: Remove empty line.

cmd/server.go
```diff
@@ -76,60 +76,96 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
 			logger.Info("Starting the http server...")
 			go handler.Start()
 			defer handler.Stop()
 
-			ssrStopCh := make(chan struct{})
+			if snapstoreConfig == nil {
+				logger.Warnf("No snapstore storage provider configured. Will not start backup schedule.")
+				handler.Status = http.StatusOK
+
+				select {
```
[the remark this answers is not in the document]
georgekuruvillak: We need to have the server running till it is stopped to cater to the health check. If we return, then the container will get restarted again and again.

cmd/server.go
georgekuruvillak: We need the health state of ssr in the handler. We are not passing the ssr object to the handler. If we need to update the handler of the health of the ssr object, we would either need to pass ssr to the handler or handler to the ssr object. Do you think thats a good idea? As for a global function, we cannot pass the ssr or handler instance in the same. How do you suggest we go about it?

swapnilgm: As we discussed, health status can be updated only in `server.go` before calling `ssr.Run` or after return from `ssr.Run`.  No need to update it from inside ssr routine.  Handler is just reading status field on /status request. So, no update call needed in handler. And since we are initializing handler in main thread of server command, there we can surely update the handler.status. If still not clear to you, lets sit together and refactor it.

georgekuruvillak: Are you suggesting we set the health status to OK and run the snapshotter. Change it when it fails? Earlier we could do it as we had the first snapshot run seperate. Now I have integrated that with ssr.Run(). I need to send out a response when the first snapshot succeeded to toggle the Ready state.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -269,108 +284,139 @@ func GetTLSClientForEtcd(tlsConfig *TLSConfig) (*clientv3.Client, error) {
...
 		StartRevision: ssr.prevSnapshot.LastRevision + 1,
-		LastRevision:  lastRevision,
+		LastRevision:  ssr.events[len(ssr.events)-1].EtcdEvent.Kv.ModRevision,
 		SnapDir:       ssr.prevSnapshot.SnapDir,
 	}
 	snap.GenerateSnapshotName()
-
-	data, err := json.Marshal(ops)
+	ssr.config.DeltaEventsMutex.Lock()
+	data, err := json.Marshal(ssr.events)
```
amshuman-kr: Future optimization: Can we think of using Readers and Writers and avoid keeping the whole data in memory?

pkg/server/httpAPI.go
```diff
@@ -84,8 +108,11 @@ func (h *HTTPHandler) serveInitialize(rw http.ResponseWriter, req *http.Request)
 		h.initializationStatus = initializationStatusProgress
 		go func() {
 			// This is needed to stop snapshotter.
-			var s struct{}
-			h.StopCh <- s
+			h.AckStatusMutex.Lock()
```
amshuman-kr: Same check and set func mentioned for the snapshotter might be helpful here as well?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -269,108 +284,139 @@ func GetTLSClientForEtcd(tlsConfig *TLSConfig) (*clientv3.Client, error) {
...
+		return
+	}
+	ssr.config.SsrStateMutex.Unlock()
+	ssr.config.SsrStateMutex.Lock()
+	if ssr.DeltaSnapshotTimer != nil {
+		ssr.DeltaSnapshotTimer.Stop()
+		ssr.DeltaSnapshotTimer = nil
+	}
+	ssr.config.SsrStateMutex.Unlock()
+	ssr.deltaStopCh <- stop
```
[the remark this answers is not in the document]
georgekuruvillak: I am creating the deltaStopSnap channel at snapshotter creation to indicate the delta snapshotter to halt till the next iteration. We need it because we cannot stop the watchCh (it is a receive only channel).  If we are to close the channel and use that as a signal to return from deltaSnapshot function, we will not be passing anything over the channel. We will need to create the channel on each deltasnapshot function invocation. Would that be a good way to do it? Should we use some other mechanism to handle this scenario? I just need to exit the for loop in `watchDeltaSnapshots` without polling overhead.

amshuman-kr:
These are pretty big changes. Let's write some tests to make sure the race scenarios are really covered.

cmd/server.go
```diff
@@ -209,3 +201,47 @@ func ProbeEtcd(tlsConfig *snapshotter.TLSConfig) error {
...
+}
+
+// handleSsrRequest responds to handlers reqeust and stop interrupt.
+func handleSsrRequest(handler *server.HTTPHandler, snapshotterConfig *snapshotter.Config, ackCh, ssrStopCh chan struct{}, stopCh <-chan struct{}) {
+	var handlerReq server.HandlerRequest
+	for {
+		select {
+		case handlerReq = <-handler.ReqCh:
+		case <-stopCh:
+			logger.Infof("Received stop signal. Terminating !!")
```
amshuman-kr: We are not terminating!

cmd/server.go
```diff
@@ -145,35 +146,26 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
-
-				go ssr.GarbageCollector(gcStopCh)
+				logger.Infof("Creating snapshotter...")
+				ssr := snapshotter.NewSnapshotter(
+					logger,
+					snapshotterConfig)
 
-				if err := ssr.Run(true, ssrStopCh); err != nil {
-					handler.Status = http.StatusServiceUnavailable
+				gcStopCh := make(chan struct{})
```
amshuman-kr: Who closes this?

cmd/server.go
```diff
@@ -145,35 +146,26 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
+						logger.Errorf("Snapshotter failed with error: %v", err)
 					}
-				} else {
-					handler.Status = http.StatusOK
 				}
-				gcStopCh <- true
+				logger.Infof("Snapshotter stopped.")
+				ackCh <- emptyStruct
+				handler.Status = http.StatusServiceUnavailable
+				gcStopCh <- emptyStruct
```
amshuman-kr: Can we not simply `Close()` here?

cmd/snapshot.go
```diff
@@ -48,25 +48,28 @@ storing snapshots on various cloud storage providers as well as local disk locat
...
+				logger.Fatalf("failed to create snapstore config: %v", err)
 			}
-			gcStopCh := make(chan bool)
-			go ssr.GarbageCollector(gcStopCh)
-			if err := ssr.Run(false, stopCh); err != nil {
+			ssr := snapshotter.NewSnapshotter(
+				logger,
+				snapshotterConfig)
+
+			gcStopCh := make(chan struct{})
```
amshuman-kr: Isn't section a duplication of code from `cmd/server.go`?

cmd/types.go
```diff
@@ -59,3 +59,5 @@ var (
 	storageContainer string
 	storagePrefix    string
 )
+
+var emptyStruct struct{}
```
amshuman-kr: Do we need this? Can we not simply `Close()` the stop channels instead?

cmd/server.go
```diff
@@ -209,3 +201,47 @@ func ProbeEtcd(tlsConfig *snapshotter.TLSConfig) error {
...
+func handleSsrRequest(handler *server.HTTPHandler, snapshotterConfig *snapshotter.Config, ackCh, ssrStopCh chan struct{}, stopCh <-chan struct{}) {
+	var handlerReq server.HandlerRequest
+	for {
+		select {
+		case handlerReq = <-handler.ReqCh:
+		case <-stopCh:
+			logger.Infof("Received stop signal. Terminating !!")
+		}
+		snapshotterConfig.SsrStateMutex.Lock()
+		if snapshotterConfig.SsrState == snapshotter.SnapshotterActive && handlerReq == server.HandlerSsrAbort {
```
amshuman-kr: Is this stuff still needed now that HTTPHandler has mutexes as well?

cmd/snapshot.go
swapnilgm: Actually yes. but `cmd/snapshot.go` and `cmd/server.go` are two different commands with different parameters to `ssr.Run`. So, we have to duplicate the code. Regarding `ssr.RunGarbageCollelctor`, we wanted it to separate from  `ssr.Run` so explicitly call it, instead of calling it from `ssr.Run`

PR #55, opened 2018-09-11
Add periodical defragmentation support for ETCD

swapnilgm:

Fixes #

**Release note**:

```improvement operator
Etcd-backup-restore utility when started in server mode will start defragmenting etcd member data directory periodically, so that db size will be under control. You can set defragmentation period in number of hours by setting `defragmentation-period-in-hours` flag on `etcdbrctl server` command.
```

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/coreos/etcd/pkg/types"
 	"github.com/gardener/etcd-backup-restore/pkg/errors"
+	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
 	"github.com/gardener/etcd-backup-restore/pkg/initializer"
 	"github.com/gardener/etcd-backup-restore/pkg/server"
 	"github.com/gardener/etcd-backup-restore/pkg/snapshot/restorer"
@@ -101,7 +102,7 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 				return
 			}
 
-			tlsConfig := snapshotter.NewTLSConfig(
+			tlsConfig := etcdutil.NewTLSConfig(
 				certFile,
 				keyFile,
 				caFile,
@@ -131,10 +132,10 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 			ssr := snapshotter.NewSnapshotter(
 				logger,
 				snapshotterConfig)
-
 			ssrStopCh = make(chan struct{})
 			go handleSsrRequest(handler, ssr, ackCh, ssrStopCh, stopCh)
 			go handleAckState(handler, ackCh)
+			go etcdutil.DefragDataPeriodically(stopCh, tlsConfig, time.Duration(defragmentationPeriodHours)*time.Hour, time.Duration(etcdConnectionTimeout)*time.Second, ssr.TriggerFullSnapshot)
 
 			for {
 				logger.Infof("Probing etcd...")
@@ -192,8 +193,8 @@ func initializeServerFlags(serverCmd *cobra.Command) {
 
 // ProbeEtcd will make the snapshotter probe for etcd endpoint to be available
 // before it starts taking regular snapshots.
-func ProbeEtcd(tlsConfig *snapshotter.TLSConfig) error {
-	client, err := snapshotter.GetTLSClientForEtcd(tlsConfig)
+func ProbeEtcd(tlsConfig *etcdutil.TLSConfig) error {
+	client, err := etcdutil.GetTLSClientForEtcd(tlsConfig)
 	if err != nil {
 		return &errors.EtcdError{
 			Message: fmt.Sprintf("failed to create etcd client: %v", err),
@@ -218,7 +219,7 @@ func handleAckState(handler *server.HTTPHandler, ackCh chan struct{}) {
 	}
 }
 
-// handleNoSsrRequest responds to handlers reqeust with acknowledgment when snapshotter is not running.
+// handleNoSsrRequest responds to handlers request with acknowledgment when snapshotter is not running.
 func handleNoSsrRequest(handler *server.HTTPHandler) {
 	for {
 		_, ok := <-handler.ReqCh
--- a/pkg/etcdutil/defrag.go
+++ b/pkg/etcdutil/defrag.go
@@ -16,47 +16,51 @@ package etcdutil
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/sirupsen/logrus"
 )
 
 // defragData defragments the data directory of each etcd member.
-func defragData(tlsConfig *TLSConfig, etcdConnectionTimeout time.Duration) {
+func defragData(tlsConfig *TLSConfig, etcdConnectionTimeout time.Duration) error {
 	client, err := GetTLSClientForEtcd(tlsConfig)
 	if err != nil {
-		logrus.Warnf("failed to create etcd client for defragmentation: %v", err)
-		return
+		return fmt.Errorf("failed to create etcd client for defragmentation: %v", err)
 	}
+	defer client.Close()
 
 	for _, ep := range tlsConfig.endpoints {
+		var dbSizeBeforeDefrag, dbSizeAfterDefrag int64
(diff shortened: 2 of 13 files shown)


pkg/etcdutil/defrag.go
```diff
@@ -0,0 +1,58 @@
...
+		if err != nil {
+			logrus.Warnf("Failed to defragment etcd member[%s] with error: %v", ep, err)
+		} else {
+			logrus.Infof("Finished defragmenting etcd member[%s]", ep)
+		}
+	}
+}
+
+// DefragDataPeriodically defragments the data directory of each etcd member
+func DefragDataPeriodically(stopCh <-chan struct{}, tlsConfig *TLSConfig, d time.Duration) {
```
amshuman-kr: This function is directly invoked as a goroutine in https://github.com/swapnilgm/etcd-backup-restore/blob/defrag/cmd/server.go#L107.

But I don't see any coordination with full and delta snapshots. Shouldn't we pause those before defragmenting? Also, would a full snapshot after defrag make sense?

cmd/server.go
```diff
@@ -200,12 +210,13 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 // initializeServerFlags adds the flags to <cmd>
 func initializeServerFlags(serverCmd *cobra.Command) {
 	serverCmd.Flags().IntVarP(&port, "server-port", "p", defaultServerPort, "port on which server should listen")
+	serverCmd.Flags().IntVar(&defragmentationPeriodInHour, "defragmentation-period-in-hours", 24, "period after which we should defragment etcd data directory")
```
amshuman-kr: Should we have some validation about the full snapshot period and the defragmentation period? For example, full snapshot period should be less than the defragment period?

pkg/etcdutil/defrag.go
swapnilgm: Snapshotting is independent of defragmentation. We won't need to worry about any coordination among them. Etcd takes care of it. At worst, etcd might not respond to snapshot request in case defragmentation takes too much time. And backup-restore sidecar will retry it afterwords.

cmd/server.go
swapnilgm: > full snapshot period should be less than the defragment period

Not necessary. If there are too many `put` and `del` request in between to full snapshots, there  are huge chances of internal fragmentation leading to larger db disk size. Defragmentation in between will help in reduction of size.

swapnilgm:
@amshuman-kr  Thanks for the review.

> Also, some tests please. It would be good to be sure that defragmentation does not affect the regular operation ;-)

Yes sure. I will add it before merging.

pkg/etcdutil/defrag.go
amshuman-kr: What will the healthcheck response be if snapshot takes too much time and times out?

cmd/server.go
amshuman-kr: 1. What happens to delta snapshots if defragmentation is done in between two full snapshots?
1. What happens to restoration in such a case?

pkg/etcdutil/defrag.go
amshuman-kr: Also, see the comment below regarding coordination during defragmentation.

swapnilgm: After time out, the health will be set to `false`. And set back to `true` on next successful retry.

cmd/server.go
swapnilgm: > What happens to delta snapshots if defragmentation is done in between two full snapshots?

Defragmentation won't loose the revision number. So, for next update events, `watch` will still send the event, delta snapshot will be taken as usual. So, object store state i.e. last full snapshot + delta snapshots over it will still be consistent.

And anyways  independent of defragmentation period smaller or larger than full snapshot period, defragmentation  step will coincide with one of the window between two consecutive full snapshot.

> What happens to restoration in such a case?

Since, restoration time there wont be any defragmentation since etcd isn''t running. Defrag request will simply fail. Sicne objects store state is stable all the time, restoration will continue as it is.

amshuman-kr: Now I am confused. Are we talking about compaction or defragmentation?

swapnilgm: We are talking about defragmentation only. What is confusing you? May be you wan't to go through https://coreos.com/etcd/docs/latest/op-guide/maintenance.html once again.

amshuman-kr: As discussed, it might be worth considering to take a full snapshot after every scheduled defragmentation (similar to what we do after a restore).

This can help avoid the edge case of restore happening between a scheduled defragmentation and the next scheduled full snapshot which can in effect nullify the defragmentation.

swapnilgm:
As per discussion, we agreed to merge it after PR, https://github.com/gardener/etcd-backup-restore/pull/51 just to minimise time resolving probable conflicts later on.

pkg/etcdutil/defrag.go
```diff
@@ -0,0 +1,79 @@
...
+			logrus.Warnf("Failed to defragment etcd member[%s] with error: %v", ep, err)
+		} else {
+			logrus.Infof("Finished defragmenting etcd member[%s]", ep)
+		}
+		ctx, cancel = context.WithTimeout(context.TODO(), etcdConnectionTimeout)
+		status, err = client.Status(ctx, ep)
+		cancel()
+		if err != nil {
+			logrus.Warnf("Failed to get status of etcd member[%s] with error: %v", ep, err)
+			continue
```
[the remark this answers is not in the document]
amshuman-kr: Also, do we need to worry about timeout issues since defrag might be a bit longer running operation?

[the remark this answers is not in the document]
swapnilgm: >Is some cleanup required for failed defrag?

Internally defragmentation create `db.tmp` file and take care of deleting it in case of failure. So, no cleanup required cause defragmentation happens at server end.  And anyways, unlike restore call,  defrag is client call on live etcd server. So, it is safe to assume that client need not have any access to server's  etcd data directory.

> Also, do we need to worry about timeout issues since defrag might be a bit longer running operation?
Its already configurable. Do you want me to keep it separate from `etcdConnectionTimeout`?

pkg/etcdutil/defrag.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, ep := range tlsConfig.endpoints {
+		logrus.Infof("Defragmenting etcd member[%s]", ep)
+		ctx, cancel := context.WithTimeout(context.TODO(), etcdConnectionTimeout)
+		status, err := client.Status(ctx, ep)
+		cancel()
+		if err != nil {
+			logrus.Warnf("Failed to get status of etcd member[%s] with error: %v", ep, err)
+			continue
+		}
+		dbSizeBeforeDefrag := status.DbSize
```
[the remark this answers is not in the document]
swapnilgm: May be not. Since because of IOPS issue we already assign large disk space to etcd data directory.

pkg/etcdutil/defrag.go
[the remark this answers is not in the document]
amshuman-kr: > Its already configurable. Do you want me to keep it separate from etcdConnectionTimeout?

 I was more asking about what happens if the timeout occurs? Would the defrag be completed still? How do we know if the operation times out, if it actually succeeded on the server side or not?

pkg/etcdutil/defrag_test.go
```diff
@@ -0,0 +1,59 @@
...
+				client.Put(ctx, fmt.Sprintf("%s%d", keyPrefix, index), valuePrefix)
+				cancel()
+			}
+			for index := 0; index <= 50; index++ {
+				ctx, cancel := context.WithTimeout(context.TODO(), etcdConnectionTimeout)
+				client.Delete(ctx, fmt.Sprintf("%s%d", keyPrefix, index))
+				cancel()
+			}
+		})
+		It("should defragment and reduce size of DB", func() {
```
amshuman-kr: Maybe also add a negative test case? Say, configure timeout in such a way the it would be guaranteed to timeout? Or nothing to defrag?

swapnilgm:
@amshuman-kr  Thanks you for the review. I have added the tests cases as per discussion. Do you mind having another look at it?

PR #57, opened 2018-10-10
Optimise the restoration process

### LINKED ISSUE #41: Optimise the restoration process
Currently, while restoring etcd from available incremental/delta snapshots, it fetch the snapshot and applies the snapshot sequentially one by one. For reducing the snapshot fetch time from object store, we can parallelise the fetch call. This can be done at two level: 
Lets we have delta snapshot in order as A, B, C ... then
- Once A is fetched, one go routine will start unmarsheling and applying events from it to embedded etcd and simultaneously another go routine will start fetching next delta snapshot i.e. B. We can think of it as kind of two step pipeline parallelism implementation.
- Another way is trigger fetch call for all (i.e. A,B as well as C) the snapshots in parallel initially itself. And apply the sequentially to embedded etcd.

cc: @georgekuruvillak @amshuman-kr

shreyas-s-rao:
This PR optimises the restoration process by parallelizing the fetching of delta snapshots.

Fixes #41 

**Release note**:

```improvement operator
Parallelization options such as maximum number of fetcher routines (maxNumFetchers), fetcher routine start delay (subsequentFetchDelay), waiting period for applier to check for next fetched snapshot (waitDuration) can be tweaked as constants, and cannot be passed as flags   ##yet.
```


amshuman-kr:
As discussed last week, can we include some tests please? ;-)

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		restorationTime := endTime.Sub(startTime)
+		r.logger.Warnf("RESTORATION TIME: %v", restorationTime)
+		return nil
+	case err := <-errCh:
+		r.logger.Infof("--- APPLY DELTA SNAPSHOTS - select errCh")
+		return err
+	}
+}
+
+// coordinator coordinates the actions of the fetcher routines and the applier routine. It ensures that events are applied in the right order, regardless of the order in which snaps are fetched
```
georgekuruvillak: The events are persisted in the files. I have one concern, though, with the implementation. Say I have 5 incr snaps and I fetch the 5 snaps in 5 threads. Now if say, 3 failed and others succeeded, applier will apply 1 and 2. Files are created for 4 and 5. 3 returned an err so files 4 and 5 are not cleaned up. 

Another problem I see is that, after the initializer starts after the err and does the data validation check, it will see that data directory is valid, though we have not applied 3, 4 and 5. That will be catastrophic. We need to remove contents from the data directory if restore fails.

georgekuruvillak: mismatch in function name. `persistIncrSnapshots` would be a better name.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		r.logger.Infof("--- FETCHLISTENER - filePath written to snapLocations[%d]", eventsInfo.SnapIndex)
+	}
+}
+
+// startFetchers starts a number of fetcher goroutines pseudo-parallelly, with time delays between the goroutine starts
+func (r *Restorer) startFetchers(remainingSnaps snapstore.SnapList, snapLocations []string, fetchedEventsInfo chan eventsInfo, errCh chan error) {
+	r.logger.Infof("--- START FETCHERS")
+	numSnaps := len(remainingSnaps)
+
+	if numMaxFetchers < 1 {
```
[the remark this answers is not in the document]
shreyas-s-rao: This was in case we allowed numFetchers to be a configurable value later on.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Indeed.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		errCh <- fmt.Errorf("Invalid number of fetcher routines")
 	}
 
-	return nil
+	numFetchers := numSnaps
+	if numFetchers < numMaxFetchers {
```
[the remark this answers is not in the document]
shreyas-s-rao: Yes. Thanks for letting me know!

pkg/snapshot/restorer/restorer.go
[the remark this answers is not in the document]
swapnilgm: Make it configurable in same PR.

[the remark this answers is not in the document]
shreyas-s-rao: Sure.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
 func (r *Restorer) applyDeltaSnapshots(client *clientv3.Client, snapList snapstore.SnapList) error {
+	startTime := time.Now()
+	r.logger.Infof("--- APPLY DELTA SNAPSHOTS")
+
+	for _, snap := range snapList {
+		r.logger.Infof("--- APPLY DELTA SNAPSHOTS - snap revision = %d, %d", snap.StartRevision, snap.LastRevision)
+	}
+
 	firstDeltaSnap := snapList[0]
 	if err := r.applyFirstDeltaSnapshot(client, *firstDeltaSnap); err != nil {
```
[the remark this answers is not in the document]
shreyas-s-rao: There seems to be an issue with etcd itself (https://github.com/coreos/etcd/issues/9037), and a workaround to that while taking delta snapshots, causes the latest revision from full snapshot and the first few revisions from the first delta snapshot to overlap, as mentioned as a note here: https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L349

@georgekuruvillak @swapnilgm Correct me if I'm wrong.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		r.logger.Warnf("RESTORATION TIME: %v", restorationTime)
+		return nil
+	case err := <-errCh:
+		r.logger.Infof("--- APPLY DELTA SNAPSHOTS - select errCh")
+		return err
+	}
+}
+
+// coordinator coordinates the actions of the fetcher routines and the applier routine. It ensures that events are applied in the right order, regardless of the order in which snaps are fetched
+func (r *Restorer) fetchListener(snapLocations []string, fetchedEventsInfo chan eventsInfo, errCh chan error) {
```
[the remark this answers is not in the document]
shreyas-s-rao: I'm using an array `snapLocations` to store the locations of the temp files into which the fetched events are being stored. Allowing individual fetcher routines to modify this array might give rise to write conflicts, which I altogether avoided by having a separate listener routine that solely writes to the `snapLocations` array. Even though each `fetcher` thread would be able to write to a different element of the array (and hence, different memory location) simultaneously without using locks, I chose the safe path by using a dedicated goroutine to handle all writes to the array. There isn't too much of an overhead in using channels for transferring information between goroutines, plus it provides code clarity for future maintainability, so I felt the pros outweighed the cons.

The `fetchedEventsInfo` channel serves to pass the events data from the fetched snapshot to the listener routine. It has nothing to do with the snapshot revision verification as such, as the verification occurs after applying the events to the embedded etcd, and as the goal is to decouple the fetch and apply logic, it made sense to club the verification within the `applier` thread.

pkg/snapshot/restorer/restorer.go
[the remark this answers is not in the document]
amshuman-kr: Understood.

pkg/snapshot/restorer/restorer.go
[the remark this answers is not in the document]
amshuman-kr: See my comments about the worker-pool pattern? Perhaps you can avoid this array business.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
-	return nil
+	numFetchers := numSnaps
+	if numFetchers < numMaxFetchers {
+		numFetchers = numMaxFetchers
+	}
+
+	snapsPerFetcher, snapIndicesPerFetcher := assignSnapsToFetchers(numFetchers, remainingSnaps)
+
+	for i := 0; i < numFetchers; i++ {
+		go r.fetcher(snapsPerFetcher[i], snapIndicesPerFetcher[i], fetchedEventsInfo, errCh)
```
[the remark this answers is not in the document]
shreyas-s-rao: Sure!

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		r.logger.Infof("--- APPLY DELTA SNAPSHOTS - select errCh")
+		return err
+	}
+}
+
+// coordinator coordinates the actions of the fetcher routines and the applier routine. It ensures that events are applied in the right order, regardless of the order in which snaps are fetched
+func (r *Restorer) fetchListener(snapLocations []string, fetchedEventsInfo chan eventsInfo, errCh chan error) {
+	r.logger.Infof("--- FETCHLISTENER")
+	for {
+		eventsInfo := <-fetchedEventsInfo
```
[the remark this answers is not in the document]
shreyas-s-rao: I never felt the need for a stop channel here till now, so I only had a success channel and an error channel in place. Now that I'll also be looking into the file cleanup during failed fetches as suggested by George, I will be using a stop channel as well.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+	snapsPerFetcher, snapIndicesPerFetcher := assignSnapsToFetchers(numFetchers, remainingSnaps)
+
+	for i := 0; i < numFetchers; i++ {
+		go r.fetcher(snapsPerFetcher[i], snapIndicesPerFetcher[i], fetchedEventsInfo, errCh)
+		time.Sleep(subsequentFetchDelay)
+	}
+}
+
+// assignSnapsToFetchers assigns delta snapshots to fetchers
+func assignSnapsToFetchers(numFetchers int, remainingSnaps snapstore.SnapList) ([][]snapstore.Snapshot, [][]int) {
```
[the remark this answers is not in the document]
shreyas-s-rao: Thanks for letting me know about this! Internally, I've been running the `fetcher` threads similar to a worker pool, and `assignSnapsToFetchers` is used to create a job-queue-like assignment of snaps for each of the fetchers. I think using the worker pool will simplify this process and increase code readability as well. Thanks! 😄

pkg/snapshot/restorer/restorer.go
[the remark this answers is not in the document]
shreyas-s-rao: I'll definitely look into this. Indeed looks like a promising alternative to the current approach. 
I'll look into using a channel for passing the snapshot file locations from fetchers to the applier. That should help do away with arrays!  :)

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+				break
+			}
+
+			if err := applyEventsToEtcd(client, events); err != nil {
+				errCh <- fmt.Errorf("failed to apply events to etcd for delta snapshot %s : %v", currSnap.SnapName, err)
+				break
+			}
+			r.logger.Infof("--- APPLIER - applied events to etcd")
+
+			if err := verifySnapshotRevision(client, currSnap); err != nil {
```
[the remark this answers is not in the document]
georgekuruvillak: @amshuman-kr After applying the delta, it checks if etcd has the revision number of the delta. I think its fine here.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		r.logger.Infof("--- FETCHER - %s - fetchedEventsInfo data sent on channel", snap.SnapName)
+	}
+}
+
+func (r *Restorer) applier(client *clientv3.Client, remainingSnaps snapstore.SnapList, snapLocations []string, restoreFinished chan bool, errCh chan error) {
+	r.logger.Infof("--- APPLIER")
+
+	currSnapIndex := 0
+	for {
+		if snapLocations[currSnapIndex] != "" {
```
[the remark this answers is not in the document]
shreyas-s-rao: Will address this.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,25 +326,192 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+
+			if currSnapIndex == len(remainingSnaps) {
+				r.logger.Infof("--- APPLIER - currSnapIndex = len(remainingSnaps)")
+				restoreFinished <- true
+				r.logger.Infof("--- APPLIER - true sent on restoreFinished")
+				break
+			}
+		} else {
+			r.logger.Infof("--- APPLIER - waiting for snapIndex %d", currSnapIndex)
+			time.Sleep(waitDuration)
```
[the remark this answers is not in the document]
shreyas-s-rao: In my original implementation, the goroutine `fetchListeners` updated the `snapLocations` array, and `applier` thread would poll this array to check for array updations, with a `waitDuration` to avoid hogging up CPU. But with the new worker pool implementation, a channel would be handling the synchronization.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Thanks for the heads up about file cleanup! I completely agree!

I see that you've raised a separate PR for the data directory cleanup for failed restores. I hope there's nothing to do for that from my side, right? Just confirming.

shreyas-s-rao:
Code changes are finished and ready for review. Will push the unit tests shortly.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+	if len(snapList) == 1 {
+		return nil
+	}
+
+	var (
+		remainingSnaps         = snapList[1:]
+		numSnaps               = len(remainingSnaps)
+		snapLocations          = make([]string, numSnaps)
+		errCh                  = make(chan error, 1)
+		snapInfoCh             = make(chan snapInfo, numSnaps)
```
swapnilgm: Keep the `snapInfoCh, fetchedEventsInfoCh,applyInfoCh` channel length to `numFetchers`.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		fetchedEventsInfoCh    = make(chan eventsInfo, numSnaps)
+		applyInfoCh            = make(chan applyInfo, numSnaps)
+		numFetchers            = int(math.Min(float64(numMaxFetchers), float64(numSnaps)))
+		stopCh                 = make(chan bool, numFetchers+2) // used for stopping all routines (fetchers, persister and applier)
+		fetchersStoppedCh      = make(chan bool, numFetchers)   // used to indicate that fetchers have stopped
```
swapnilgm: It could be simplified with use of [`WaitGroup`](https://golang.org/pkg/sync/#WaitGroup). What do you think?

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
 			return err
+		default:
+			time.Sleep(time.Millisecond * 10)
+		}
+	}
+}
+
+// cleanup stops all running goroutines and removes the persisted snapshot files from disk
+func (r *Restorer) cleanup(numFetchers int, snapLocations []string, stopCh chan<- bool, fetchersStoppedCh <-chan bool, persistDeltasStoppedCh <-chan bool, applyStoppedCh <-chan bool) {
+	for i := 0; i < numFetchers+2; i++ {
```
swapnilgm: Use `waitGroup`. 
FYI,  for such case where you wan't send stop signal to everybody who are sharing same channel and the value in channel doesn't matter for them, just use `close(stopCh)`. No need of buffered `stopCh` and for loop sending individual stop channel to each fetcher.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		stopCh <- true
+	}
+
+	for i := 0; i < numFetchers; i++ {
```
swapnilgm: Use `waitGroup`.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+				Data:      eventsData,
+				Snap:      snapInfo.Snapshot,
+				SnapIndex: snapInfo.SnapIndex,
+			}
+			fetchedEventsInfoCh <- eventsInfo
+		}
+	}
+}
+
+// persistDeltaSnapshots receives delta snapshot events from fetch routines and persists them to disk, while informing the apply function about the same
```
swapnilgm: Full stop.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+
+			tmpFile, err := ioutil.TempFile(tmpDir, tmpEventsDataFilePrefix)
+			if err != nil {
+				errCh <- fmt.Errorf("failed to create temp file for delta snapshot %s : %v", eventsInfo.Snap.SnapName, err)
+				applyInfoCh <- applyInfo{SnapIndex: -1}
+				return
+			}
+
+			_, err = tmpFile.Write(eventsInfo.Data)
+			if err != nil {
```
swapnilgm: Refactor it.
```go
if _, err := tmpFile.Write(eventsInfo.Data);  err != nil {
```

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+// persistDeltaSnapshots receives delta snapshot events from fetch routines and persists them to disk, while informing the apply function about the same
+func (r *Restorer) persistDeltaSnapshots(numSnaps int, snapLocations []string, fetchedEventsInfoCh <-chan eventsInfo, applyInfoCh chan<- applyInfo, errCh chan<- error, stopCh <-chan bool, stoppedCh chan<- bool) {
+	defer func() {
+		stoppedCh <- true
+	}()
+
+	for i := 0; i < numSnaps; i++ {
+		select {
+		case stop := <-stopCh:
+			if stop {
```
swapnilgm: See comment above.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+				return
+			}
+		default:
+			eventsInfo := <-fetchedEventsInfoCh
```
swapnilgm: Refactor:
```go
case eventsInfo := <-fetchedEventsInfoCh:
```

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		fetchedEventsInfoCh    = make(chan eventsInfo, numSnaps)
+		applyInfoCh            = make(chan applyInfo, numSnaps)
+		numFetchers            = int(math.Min(float64(numMaxFetchers), float64(numSnaps)))
+		stopCh                 = make(chan bool, numFetchers+2) // used for stopping all routines (fetchers, persister and applier)
+		fetchersStoppedCh      = make(chan bool, numFetchers)   // used to indicate that fetchers have stopped
+		persistDeltasStoppedCh = make(chan bool, numFetchers)   // used to indicate that persister has stopped
+		applyStoppedCh         = make(chan bool, numFetchers)   // used to indicate that applier has stopped
+	)
+
+	go r.persistDeltaSnapshots(numSnaps, snapLocations, fetchedEventsInfoCh, applyInfoCh, errCh, stopCh, persistDeltasStoppedCh)
```
swapnilgm: Number of persister should be equal to number of fetchers, shouldn't it? Cause persister is also, doing diskIO task of disk which can be parallelised.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+			applyInfoCh <- applyInfo{
+				EventsFilePath: filePath,
+				SnapIndex:      eventsInfo.SnapIndex,
+			}
+			tmpFile.Close() // Never use defer within a loop; defer is stack-hungry
 		}
 	}
+}
+
+// apply applies delta snapshot events to the embedded etcd sequentially, in the right order of snapshots, regardless of the order in which they were fetched
```
swapnilgm: Full stop.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+	nextSnapIndexToApply := 0
 
+	for i := 0; i < len(remainingSnaps); i++ {
+		select {
+		case stop := <-stopCh:
+			if stop {
+				return
+			}
+		default:
+			applyInfo := <-applyInfoCh
```
swapnilgm: Refactor:
```go
case applyInfo := <-applyInfoCh:
```

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+			filePath := applyInfo.EventsFilePath
+			fetchedSnapIndex := applyInfo.SnapIndex
+			snapName := remainingSnaps[fetchedSnapIndex].SnapName
+
+			eventsData, err := ioutil.ReadFile(filePath)
+			if err != nil {
+				errCh <- fmt.Errorf("failed to read events data from file for delta snapshot %s : %v", snapName, err)
+				return
+			}
+			os.Remove(filePath)
```
swapnilgm:
```go
if err:= os.Remove(filePath); err !=nil && !os.IsNotExists(err) {
 r.looger.Warnf("Failed to remove file, name: %s, err: %v")
}

File: pkg/snapshot/restorer/restorer.go [hunk 24]
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+			}
+			if fetchedSnapIndex == nextSnapIndexToApply {
+				for j := fetchedSnapIndex; j < len(remainingSnaps); j++ {
+					if !snapFetchStatus[j] {
+						break
+					}
+
+					r.logger.Infof("Applying delta snapshot %s", path.Join(remainingSnaps[j].SnapDir, remainingSnaps[j].SnapName))
+
+					err := applyEventsAndVerify(client, eventsList[j], remainingSnaps[j], errCh)
```
swapnilgm: Refactor:
```go
if err := applyEventsAndVerify(client, eventsList[j], remainingSnaps[j], errCh); err !=nil {
```

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+						errCh <- nil // restore finished
+						return
+					}
+				}
+			}
+		}
+	}
+}
+
+// applyEventsAndVerify applies events from one snapshot to the embedded etcd and verifies the correctness of the sequence of snapshot applied
```
swapnilgm: Add full stop.

pkg/snapshot/restorer/restorer.go
```diff
@@ -48,6 +49,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
```
swapnilgm: Move this to `pkg/snapshot/restorer/types.go`.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+		select {
+		case err := <-errCh:
+			r.cleanup(numFetchers, snapLocations, stopCh, fetchersStoppedCh, persistDeltasStoppedCh, applyStoppedCh)
+			if err == nil {
+				r.logger.Infof("Restoration complete.")
+				return nil
+			}
+			r.logger.Warnf("Restoration failed.")
 			return err
+		default:
```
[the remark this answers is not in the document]
shreyas-s-rao: Eliminated it

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+// fetch fetches delta snapshots as events and passes them on to be persisted onto disk; multiple instances of this function are run in parallel
+func (r *Restorer) fetch(fetcherIndex int, snapInfoCh <-chan snapInfo, fetchedEventsInfoCh chan<- eventsInfo, errCh chan<- error, stopCh <-chan bool, stoppedCh chan<- bool) {
+	defer func() {
+		stoppedCh <- true
+	}()
+
+	for snapInfo := range snapInfoCh {
+		select {
+		case stop := <-stopCh:
+			if stop {
```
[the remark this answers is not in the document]
shreyas-s-rao: Sure. Addressed.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,24 +326,260 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+			continue
+		}
+		if _, err := os.Stat(filePath); !os.IsNotExist(err) {
+			os.Remove(filePath)
+		}
+	}
+	r.logger.Infof("Cleanup complete")
+}
+
+// fetch fetches delta snapshots as events and passes them on to be persisted onto disk; multiple instances of this function are run in parallel
```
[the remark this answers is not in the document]
shreyas-s-rao: Have removed it from the comment.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: `snapInfoCh` needs to have buffer length of `numSnaps`, as we don't want the main function `applyDeltaSnapshots` to wait for fetchers, because in case of failure of one of the fetchers, `applyDeltaSnapshots` will not be able to receive on the `errCh`, as it is still waiting to send on `snapInfoCh` but fetcher(s) have exited.

In case `persister` is taking longer than fetchers, fetchers would have to wait to send on `fetchedEventsInfoCh` if there isn't enough buffer for it. Hence, buffer length is `numSnaps`.

The same logic applies to `applyInfoCh` as well, and usually applying snaps takes longer than persisting them, and we don't want persister to wait for applier. Hence, buffer length is `numSnaps`.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Yes. I've implemented this now, and eliminated `fetchersStoppedCh`, `persistDeltasStoppedCh` and `applyStoppedCh`. Thanks.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Sure. Thanks! Have implemented this now.  :)

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Done.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: During restoration, fetching and applying snaps usually take more time than persisting them. Applier cannot be parallelised, and fetcher is now parallelised. I didn't see the need to parallelise persister as well, as it wouldn't lead to any real performance improvement during restoration, but would still add the overhead of more go routines and synchronisation. What do you think?

shreyas-s-rao:
Will push unit tests shortly.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Fetcher and persister functionalities have now been combined to help curb memory over-utilisation, and also keeps clean, maintainable code.

pkg/snapshot/restorer/restorer.go
```diff
@@ -317,28 +322,214 @@ func startEmbeddedEtcd(ro RestoreOptions) (*embed.Etcd, error) {
...
+			}
+
+			eventsList[fetchedSnapIndex] = events
+			snapFetchStatus[fetchedSnapIndex] = true
+
+			if fetchedSnapIndex < nextSnapIndexToApply {
+				errCh <- fmt.Errorf("snap index mismatch for delta snapshot %d; expected snap index to be atleast %d", fetchedSnapIndex, nextSnapIndexToApply)
+				return
+			}
+			if fetchedSnapIndex == nextSnapIndexToApply {
```
amshuman-kr: What if `fetchedSnapIndex > nextSnapIndexToApply`?

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -0,0 +1,240 @@
...
+				RestoreDataDir: restoreDataDir,
+				PeerURLs:       peerUrls,
+				SkipHashCheck:  skipHashCheck,
+				Name:           restoreName,
+				MaxFetchers:    maxFetchers,
+				BaseSnapshot:   *baseSnapshot,
+				DeltaSnapList:  deltaSnapList,
+			}
+			err = rstr.Restore(restoreOptions)
+			Expect(err).Should(HaveOccurred())
```
amshuman-kr: Might be good to validate the error message also.

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: It will just continue to the next iteration. The logic is: when `fetchedSnapIndex` == `nextSnapIndexToApply`, it applies that snap as well as any existing subsequent snaps till an unfetched snap is encountered, and it moves to the next iteration. This way, all snaps are applied in the right order.

pkg/snapshot/restorer/restorer_test.go
shreyas-s-rao: Sure, will definitely look into it.

amshuman-kr:
LGTM

PR #58, opened 2018-10-11
Remove the data directory in case of a failed restore.

georgekuruvillak:
This PR is a hotfix that fixes the bug where etcd can start with partially applied snapshot. If the incremental snapshot is partially applied due to failure in between, the etcd data is consistent though not up-to-date. This PR fixes this issue.
Fixes #

**Release note**:

```improvement operator
At the time of restoration, the etcd data directory will be restored to temporary directory with suffix `.part` i.e.`<path-to-etcd-data-dir>.part`. On successful restoration we will replace actual etcd data directory with this. This brings standard and more cleaner approach to restoration. 
```

### DIFF
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -78,7 +78,7 @@ func (e *EtcdInitializer) restoreCorruptData() error {
 	dataDir := e.Config.RestoreOptions.RestoreDataDir
 
 	if e.Config.SnapstoreConfig == nil {
-		logger.Warnf("No snapstore storage provider configured. Will only clean the directory.")
+		logger.Warnf("No snapstore storage provider configured.")
 		return nil
 	}
 	store, err := snapstore.GetSnapstore(e.Config.SnapstoreConfig)
@@ -99,10 +99,16 @@ func (e *EtcdInitializer) restoreCorruptData() error {
 
 	e.Config.RestoreOptions.BaseSnapshot = *baseSnap
 	e.Config.RestoreOptions.DeltaSnapList = deltaSnapList
+	tempRestoreOptions := *e.Config.RestoreOptions
+	tempRestoreOptions.RestoreDataDir = fmt.Sprintf("%s.%s", tempRestoreOptions.RestoreDataDir, "part")
+
+	logger.Infof("Removing data directory(%s) for snapshot restoration.", tempRestoreOptions.RestoreDataDir)
+	if err := os.RemoveAll(filepath.Join(tempRestoreOptions.RestoreDataDir)); err != nil {
+		return fmt.Errorf("failed to delete previous temporary data directory %s with err: %v", tempRestoreOptions.RestoreDataDir, err)
+	}
 
 	rs := restorer.NewRestorer(store, logger)
-
-	if err := rs.Restore(*e.Config.RestoreOptions); err != nil {
+	if err := rs.Restore(tempRestoreOptions); err != nil {
 		err = fmt.Errorf("Failed to restore snapshot: %v", err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/initializer/initializer.go
```diff
@@ -46,8 +46,12 @@ func (e *EtcdInitializer) Initialize() error {
 		return err
 	}
 	if dataDirStatus != validator.DataDirectoryValid {
-		if err := e.restoreCorruptData(); err != nil {
-			return fmt.Errorf("error while restoring corrupt data: %v", err)
+		if rerr := e.restoreCorruptData(); rerr != nil {
+			dataDir := e.Config.RestoreOptions.RestoreDataDir
+			if err := removeContents(filepath.Join(dataDir)); err != nil {
```
amshuman-kr: Why not restore to a different directory and rename/move the directory once restore is complete?
This is a widely used practice and for good reasons. For example, it doesn't create open questions about what happens if the delete fails, etc.

georgekuruvillak: Ah. Yes. Thats better.

pkg/initializer/initializer.go
```diff
@@ -104,32 +99,36 @@ func (e *EtcdInitializer) restoreCorruptData() error {
 
 	e.Config.RestoreOptions.BaseSnapshot = *baseSnap
 	e.Config.RestoreOptions.DeltaSnapList = deltaSnapList
+	tempRestoreOptions := *e.Config.RestoreOptions
+	tempRestoreOptions.RestoreDataDir = fmt.Sprintf("%s.%s", tempRestoreOptions.RestoreDataDir, "part")
 
-	rs := restorer.NewRestorer(store, logger)
+	logger.Infof("Removing data directory(%s) for snapshot restoration.", tempRestoreOptions.RestoreDataDir)
+	if err := os.RemoveAll(filepath.Join(tempRestoreOptions.RestoreDataDir)); err != nil {
+		return fmt.Errorf("failed to delete member directory: %v", err)
```
swapnilgm: Change the error message: `fmt.Errorf("failed to remove temporary data directory %s with err: %v",dataDir, err)`

pkg/initializer/initializer.go
```diff
@@ -104,32 +99,36 @@ func (e *EtcdInitializer) restoreCorruptData() error {
...
-		return err
+	logger.Infoln("Successfully restored the etcd data directory.")
+	return nil
+}
+
+func (e *EtcdInitializer) removeContents(dataDir string) error {
+	logger := e.Logger
+	logger.Infof("Removing data directory(%s) for snapshot restoration.", dataDir)
+	if err := os.RemoveAll(filepath.Join(dataDir)); err != nil {
+		return fmt.Errorf("failed to delete member directory: %v", err)
```
swapnilgm: Change the error message. `fmt.Errorf("failed to delete data directory %s with err: %v",dataDir, err)`

pkg/initializer/initializer.go
```diff
@@ -104,32 +99,36 @@ func (e *EtcdInitializer) restoreCorruptData() error {
...
 	}
-	for _, name := range names {
-		err = os.RemoveAll(filepath.Join(dir, name))
-		if err != nil {
-			return err
-		}
+
+	if err := os.Rename(filepath.Join(fmt.Sprintf("%s.%s", dataDir, "part")), filepath.Join(dataDir)); err != nil {
+		return fmt.Errorf("Failed to rename temp snap directory to member directory: %v", err)
```
swapnilgm: Change the error message: `fmt.Errorf("failed to rename data directory %s with err: %v",dataDir, err)`

PR #80, opened 2018-12-06
Change AllowReauth to be set to true.

georgekuruvillak:
The token from the initial authentication expires the sidecar does not acquire a new token leaving the sidecar broken. `AllowReauth` flag in the auth options has to be set to `true` explicitly so that gophercloud can re-authenticate the token.
Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/snapstore/swift_snapstore.go
+++ b/pkg/snapstore/swift_snapstore.go
@@ -54,6 +54,10 @@ func NewSwiftSnapStore(bucket, prefix, tempDir string, maxParallelChunkUploads i
 	if err != nil {
 		return nil, err
 	}
+	// AllowReauth should be set to true if you grant permission for Gophercloud to
+	// cache your credentials in memory, and to allow Gophercloud to attempt to
+	// re-authenticate automatically if/when your token expires.
+	authOpts.AllowReauth = true
 	provider, err := openstack.AuthenticatedClient(authOpts)
 	if err != nil {
 		return nil, err


pkg/snapstore/swift_snapstore.go
```diff
@@ -54,6 +54,10 @@ func NewSwiftSnapStore(bucket, prefix, tempDir string, maxParallelChunkUploads i
 	if err != nil {
 		return nil, err
 	}
+	// AllowReauth should be set to true if you grant permission for Gophercloud to
+	// cache your credentials in memory, and to allow Gophercloud to attempt to
+	// re-authenticate automatically if/when your token expires.
+	authOpts.AllowReauth = true
```
shreyas-s-rao: According to the documentation at https://godoc.org/github.com/gophercloud/gophercloud#AuthOptions.AllowReauth :

> NOTE: The reauth function will try to re-authenticate endlessly if left unchecked. The way to limit the number of attempts is to provide a custom HTTP client to the provider client and provide a transport that implements the RoundTripper interface and stores the number of failed retries. For an example of this, see here https://github.com/rackspace/rack/blob/1.0.0/auth/clients.go#L311

I feel we should limit the number of retries, and maybe keep this configurable as well? Just a suggestion.

shreyas-s-rao: Ah ok. You're right, the re-authentication retry happens only once. Thanks for the clarification.  😀

PR #84, opened 2018-12-12
[WIP] Add memory limit on delta snapshot events aggregation

### LINKED ISSUE #81: Parameter to limit maximum number of delta changes before a delta snapshot is triggered
### Problem
The frequency of delta snapshots is only controlled by the parameter [`delta-snapshot-period-seconds`](https://github.com/gardener/etcd-backup-restore/blob/master/cmd/snapshot.go#L95).
In case of very frequent changes to ETCD, a large number of changes might be accumulated before the next delta snapshot is triggered according to the configured `delta-snapshot-period-seconds`.

Since, this is related to the load, the behaviour of the backup/restore sidecar could be unpredictable including a potential crash of the backup/restore sidecar if sufficient memory is not available to the sidecar container.

### Solution
To make things more robust, we must introduce a parameter (`max-changes-per-delta-snapshot`?) to limit the number of changes that are accumulated before a delta snapshot is triggered.

Delta snapshot could be triggered based on the new parameter (`max-changes-per-delta-snapshot`) or the older parameter (delta-snapshot-period-seconds`) whichever happens earlier.

### Alternative
Use maximum limit to the size of the delta change as a parameter. Could this be harder to compute than a simple number of changes?

swapnilgm:

For delta snapshots we aggregate event in memory over delta snapshot period. For sudden load, this may lead to memory consumption. So, to avoid this we will trigger delta snapshot irrespective to delta snapshot period once the memory threshold is reached. 

Fixes #81 

**Release note**:

```noteworthy user
Added configurable flag `delta-snapshot-memory-limit` to restrict memory consumption due to periodic delta snapshots.
```


swapnilgm:
@shreyas-s-rao @georgekuruvillak Please take review of this PR on priority. Since this is blocker for performance team.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -234,10 +246,12 @@ func (ssr *Snapshotter) takeDeltaSnapshot() error {
 	}
 	// compute hash
 	hash := sha256.New()
-	hash.Write(data)
+	if _, err := io.Copy(hash, bytes.NewReader(data)); err != nil {
```
shreyas-s-rao: Does `io.Copy()` provide any significant advantage over `Write()` ?

swapnilgm: Hmm, right. It won't in this case. Since `hash.Write` is actually does in memory computation over byte array and there isn't network/disk IO involved.  I Will rollback but do add missing error handling.

swapnilgm:
@shreyas-s-rao  Thank you for the review. I have addressed your suggestion. PTAL.

PR #96, opened 2019-01-21
Allow disabling delta snapshots

### LINKED ISSUE #94: Add options to disable delta snpashots
For performance testing, we want to do the tests without any possible effects from delta snapshots or watch mechanism we use. So, there is a need for a configurable parameter to disable delta snapshots. Currently we correct delta snapshot time interval less than 1  to 1 seconds in code with warning log. We should change this behaviour and use that check to disable delta snapshot and watch.

shreyas-s-rao:
Added the option to disable delta snapshots, by setting the `--delta-snapshot-period-seconds` flag to any value lesser than 1.

Fixes #94 

**Release note**:

```improvement operator
Added the option to disable delta snapshots, by setting the 'delta-snapshot-period-seconds' flag to any value lesser than 1.
```


shreyas-s-rao:
Will push tests shortly.

shreyas-s-rao:
@georgekuruvillak @swapnilgm I've updated the PR. PTAL.

PR #105, opened 2019-01-30
Update issue templates

swapnilgm:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,28 @@
...
+assignees: ''
+
+---
+
+**Describe the bug:**
+A clear and concise description of what the bug is.
+
+**Expected behaviour:**
+A clear and concise description of what you expected to happen.
+
```
shreyas-s-rao: Shall we have a "Current behaviour" section to let the issue-creator tell us exactly what problem they are experiencing, as opposed to what they expected to happen..?

swapnilgm: First section i.e. `Describe the bug` is exactly where issue raiser will right the current behaviour. And i don't think need to rename the section.

georgekuruvillak:
lgtm

PR #106, opened 2019-01-30
[WIP] Negative tests for network interruption

### LINKED ISSUE #73: Negative tests for single-node backup restoration
The tests should cover the following scenarios.

- [ ] Parallel backups (full and delta)
  - [ ] If backups are already being taken when restoration starts, then the backups should be stopped, discarded and any future scheduling of the backups should be suspended until after restoration.
  - [ ] If backups start while restoration is in progress then such backups should fail and should not be retried until after the restoration is completed.
- [ ] Parallel Defragmentation/compaction
  - [ ] If defragmentation/compaction is already being done when restoration starts, then such operations should be stopped and discarded and any future scheduling of such operations should be suspended until after restoration.
  - [ ] If defragmentation/compaction starts while restoration is in progress then such operations should fail and should not be retried until after the restoration is completed.
- [ ] Download interruption
The interruption could be anything from broken connections to the process being restarted.
  - [ ] If full snapshot download is interrupted then it should be retried, preferably without a container restart or at least on the container restart.
  - [ ] If any of the delta snapshot download is interrupted then it should be retried, preferably without a container restart or at least on the container restart.
    - [ ] In either case, sequence of applying the delta backups should not be affected.
- [x] Restoration interruption
The interruption could be anything from broken connections to the process being restarted.
  - [x] If full snapshot restoration is interrupted, then it should be retried, preferably without a container restart or at least on the container restart.
For example, what happens if the process is restarted [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L73) or [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L89)?
  - [ ] If a delta snapshot restoration is interrupted, then it should be retried, preferably without a container restart or at least on the container restart.
For example, what happens if restoration is interrupted [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L336), [anywhere here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L406-L437) or [anywhere here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L450-L505)?
    - [ ] In either case, sequence of applying the delta backups should not be affected.
- [x] Backup corruption
  - [x] If any of the backup files (full or delta) are corrupted then the restoration should fail and ETCD container should never be available/ready.
- [ ] ETCD availability
  - [ ] Regardless of how much time or how many iterations are needed (based on the above scenarios) for a restoration to complete, ETCD should not be available/ready until restoration is successfully completed.
- [ ] Data Sanity
  - [ ] Regardless of how much time or how many iterations are needed (based on the above scenarios) for a restoration to complete, ETCD should never be restored to any state except the latest backup. If any part of the latest backup is corrupted then the ETCD should never be restored successfully.

shreyas-s-rao:
Negative tests for network interruption scenarios.

 fixes partially #73 and #104 

This is only a first-cut version of setting up negative tests, and includes only network interruption cases, which might occur during calls to the object store.

**Release note**:

```improvement operator
Negative tests for network interruption scenarios.
```

### DIFF
--- a/pkg/snapstore/snapstore_suite_test.go
+++ b/pkg/snapstore/snapstore_suite_test.go
@@ -27,8 +27,9 @@ import (
 )
 
 var (
-	testObj    *testing.T
-	swiftStore snapstore.SnapStore
+	testObj            *testing.T
+	swiftStore         snapstore.SnapStore
+	networkTimeoutFlag bool = false
 )
 
 func TestSnapstore(t *testing.T) {
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -171,7 +171,7 @@ func (ssr *Snapshotter) TakeFullSnapshotAndResetTimer() error {
 	} else {
 		ssr.logger.Infof("Stopping full snapshot...")
 		ssr.fullSnapshotTimer.Stop()
-		ssr.logger.Infof("Reseting full snapshot to run after %d secs.", effective.Sub(now))
+		ssr.logger.Infof("Reseting full snapshot to run after %f secs.", effective.Sub(now).Seconds())
 		ssr.fullSnapshotTimer.Reset(effective.Sub(now))
 	}
 	ssr.logger.Infof("Will take next full snapshot at time: %s", effective)
--- a/pkg/snapstore/abs_snapstore_test.go
+++ b/pkg/snapstore/abs_snapstore_test.go
@@ -136,6 +136,12 @@ func (p *fakePolicy) handleContainerGetOperation(w *http.Response) error {
 
 // handleListObjectNames responds with a blob `List` response.
 func (p *fakePolicy) handleListObjects(w *http.Response) error {
+	if networkTimeoutFlag {
+		w.StatusCode = http.StatusRequestTimeout
+		w.Body = http.NoBody
+		return nil
+	}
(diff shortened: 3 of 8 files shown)


pkg/snapstore/snapstore_suite_test.go
```diff
@@ -27,8 +27,9 @@ import (
 )
 
 var (
-	testObj    *testing.T
-	swiftStore snapstore.SnapStore
+	testObj            *testing.T
+	swiftStore         snapstore.SnapStore
+	networkTimeoutFlag bool = false
```
amshuman-kr: We will need eventually simulate other errors as well such as different HTTP errors, latency etc.

This flag is better kept in the fakePolicy or the errorPolicy I proposed rather than as a global flag.

swapnilgm:
What is the status here? if it is matching with https://github.com/gardener/etcd-backup-restore/pull/173 can we close this in favor of other PR.?

shreyas-s-rao:
@swapnilgm We can close this PR for now and reopen when necessary. Would be easier to run such negative test scenarios from infra perspective once we have etcd controller. WDYT?

Issue #102, opened 2019-01-29
Possible memory leak by Backup sidecar during delta snapshot 


nikhilsap:
We did few performance based test in the cluster, we noticed memory consumption by backup side car during Delta snapshot is more. 
Even after delta snapshot, memory usage is high and it is not coming down.

![image](https://user-images.githubusercontent.com/41104107/51895895-425c6800-23d1-11e9-93ee-03759502e07e.png)

Here in this case, 
We have loaded the db using benchmark tool. The DB size was 512 mb and delta snapshot period is 5 mins. we can see that, after Delta snapshot (13:14) the memory used by backup raised to more then 3GB where as ETCD memory consumption is around 350mb.

Here is other metrics during the same load.

![image](https://user-images.githubusercontent.com/41104107/51897768-7df93100-23d5-11e9-993a-116e2bcce40b.png)

We have trigged the load around 13.09 and it has taken full snapshot at 13.20. During full snapshot , it has a small rise in memory for both etcd and backup sidecar. 

etcd-backup-restore version : 0.5.0-dev-pref-test-3-memory and perf-test-2-with-metrics

swapnilgm:
@nikhilsap  Willl you please mention the etcd-backup-restore version or commit id  on which you observed the issue?

nikhilsap:
@swapnilgm i have updated the issue with etcd-backup-restore version.

databus23:
We are seeing quite a lot of backup container (version 0.4.1) constantly hitting their memory limit in CCEE. The usage graph seems to suggest its some sort of memory leak:
![image](https://user-images.githubusercontent.com/439223/52711817-11fcf800-2f93-11e9-91bf-98da3341f068.png)

amshuman-kr:
@databus23's [PR#116](https://github.com/gardener/etcd-backup-restore/pull/116) seems to have nailed it. 🥇👏 

A comparison is below.
1. [https://github.com/gardener/etcd-backup-restore/tree/master/head](https://github.com/gardener/etcd-backup-restore/commit/b173f994ce8b1a3c9e3f97c049d2cf3b5c887658)
<img width="1241" alt="gcvis-only-delta-master" src="https://user-images.githubusercontent.com/32157254/52798471-00ecdd80-309e-11e9-9548-6c9319668ca8.png">

[...]

PR #116, opened 2019-02-14
Fix memory leak during full snapshot

databus23:
This fixes a memory/goroutine leak that caused several gouroutines and memory to leak every time a full snapshot was execucted because the old watch was never closed and continued to receive updates that were not processed.

From the etcd client godocs:
> The context "ctx" MUST be canceled, as soon as watcher is no longer being used, to release the associated resources.

https://github.com/etcd-io/etcd/blob/4cd0bf8ea846a0d158761d55899f631eb2a423cf/clientv3/watch.go#L52-L53

It fixes a serious memory leak

Not sure maybe #102

From Converged Cloud Enterprise Edition with ❤️ 

**Release note**:

```noteworthy operator
Fixed memory/goroutine leak: close previous Etcd watches
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -95,6 +95,7 @@ func NewSnapshotter(logger *logrus.Logger, config *Config) *Snapshotter {
 		SsrState:       SnapshotterInactive,
 		SsrStateMutex:  &sync.Mutex{},
 		fullSnapshotCh: make(chan struct{}),
+		cancelWatch:    func() {},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


databus23:
I just found #99 which also talks about this memory leak but the PR seems to be changing a lot more than just this memory leak.
/cc @swapnilgm

amshuman-kr:
@databus23 Thanks a lot. This seems to have [nailed it](https://github.com/gardener/etcd-backup-restore/issues/102#issuecomment-463679537). ❤️

@swapnilgm Can we take this up for validation and merge?

databus23:
@amshuman-kr Thanks for responding so quickly. I would love a quick 0.4.2 release. we have 30 instances of the latest (0.4.1) backup sidecar constantly crashing (~6h) because of this memory leak.

amshuman-kr:
@databus23 We will take the release on priority tomorrow.

databus23:
a little follow up. I'm still pretty satisfied with the results for far:
image

The graph shows the last 2 days of a crashing backup sidecar. The newest two graphs are with the patch applied.

amshuman-kr:
@databus23 Thanks for the follow up.

I verified the combination of full and delta snapshot too. The memory leaks seems to be resolved in that scenario too. But full snapshot is still taking up [more memory](https://github.com/gardener/etcd-backup-restore/issues/102#issuecomment-463905222) than I think it should. But it is not a memory leak anymore. So, it is not a blocker for a quick release.

swapnilgm:
@amshuman-kr Thanks you for comparing results and verifying it.

PR #131, opened 2019-02-25
Fix snapstore sorting for snapshots with same timestamp

shreyas-s-rao:
This PR allows for correct sorting of snapstore in cases where snapshots have same `CreatedOn` timestamp.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/snapstore/snapshot_test.go
+++ b/pkg/snapstore/snapshot_test.go
@@ -76,6 +76,55 @@ var _ = Describe("Snapshot", func() {
 			})
 		})
 
+		Context("when provied with list of snapshots where few have the same timestamp", func() {
+			It("sorts snapshot by creation time as well as start revision", func() {
+				interval := int64(5)
+				now := time.Now().Unix()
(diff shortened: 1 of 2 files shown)


pkg/snapstore/snapshot_test.go
```diff
@@ -76,6 +76,55 @@ var _ = Describe("Snapshot", func() {
 			})
 		})
 
+		Context("when provied with list of snapshots where few have the same timestamp", func() {
```
swapnilgm: Why not merge with https://github.com/gardener/etcd-backup-restore/pull/131/files#diff-cd87350645939a896c52eacbb73c2729R32.  We can handle it under same case.

shreyas-s-rao: I felt it would be better to separate levels of the test cases where first one handles vanilla case while the second one handles cases with same timestamps. WDYT?

swapnilgm: Make sense 👍

PR #136, opened 2019-03-13
Cherry pick fixes from master

swapnilgm:
This PR simply cherry-picks the recent fixes from master.
Fixes #

As usual like pre-release step, please create a shoot using gardener. And check if etcd-events and etcd-main are working fine there.

**Release note**:

```noteworthy user
Added the `embedded-etcd-quota-bytes` flag to allow configuring the backend quota size of the embedded etcd instance used during restoration of data.
```

### DIFF
--- a/chart/templates/etcd-statefulset.yaml
+++ b/chart/templates/etcd-statefulset.yaml
@@ -91,6 +91,9 @@ spec:
         - --data-dir=/var/etcd/data/new.etcd
         - --storage-provider={{ .Values.backup.storageProvider }}
         - --store-prefix=etcd-{{ .Values.role }}
+{{- if .Values.backup.embeddedEtcdQuotaBytes }}
+        - --embeddedEtcdQuotaBytes={{ .Values.backup.embeddedEtcdQuotaBytes }}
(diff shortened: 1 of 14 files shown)


chart/templates/etcd-statefulset.yaml
```diff
@@ -91,6 +91,9 @@ spec:
         - --data-dir=/var/etcd/data/new.etcd
         - --storage-provider={{ .Values.backup.storageProvider }}
         - --store-prefix=etcd-{{ .Values.role }}
+{{- if .Values.backup.embeddedEtcdQuotaBytes }}
+        - --embeddedEtcdQuotaBytes={{ .Values.backup.embeddedEtcdQuotaBytes }}
```
rfranzke: Shouldn't this be `embedded-etcd-quota-bytes`?

swapnilgm: Nice catch. Thank you.

swapnilgm: I'll do the correction with helm chart upgrade tomorrow.

PR #159, opened 2019-05-09
Skip taking initial delta snapshot if no new events are present

shreyas-s-rao:

This PR allows to skip taking the initial delta snapshot if no new events have been accumulated on etcd since the previous snapshot.

Fixes #

Previously, even if no new events are accumulated on etcd, initial delta snapshot would be attempted but later skipped since no events are present. This PR skips attempting to take the delta snapshot itself if no new events are present, so an unnecessary delta snapshot attempt is avoided.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -185,7 +185,8 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 					}
 				} else {
 					startWithFullSnapshot = true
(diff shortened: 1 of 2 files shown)


swapnilgm:
This case is already handled at https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/snapshotter/snapshotter.go#L294. No need to double check it. Closing it. Reopen it if there is some other use case i've missed.

PR #161, opened 2019-05-14
Consider prefix while the sorting for snapshots

swapnilgm:

Fixes #

**Release note**:

```improvement user
Fixed the sorting of snapshots. 
```

### DIFF
--- a/pkg/snapstore/snapshot.go
+++ b/pkg/snapstore/snapshot.go
@@ -44,7 +44,13 @@ func (s *Snapshot) GenerateSnapshotName() {
 
 // GenerateSnapshotDirectory prepares the snapshot directory name from metadata
 func (s *Snapshot) GenerateSnapshotDirectory() {
-	s.SnapDir = fmt.Sprintf("%s-%d", "Backup", s.CreatedOn.Unix())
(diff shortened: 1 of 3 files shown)


pkg/snapstore/snapshot.go
```diff
@@ -106,6 +112,21 @@ func ParseSnapshot(snapPath string) (*Snapshot, error) {
 func (s SnapList) Len() int      { return len(s) }
 func (s SnapList) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 func (s SnapList) Less(i, j int) bool {
+	// Ignoring errors here, as we assume at this stage the error won't happen.
+	iCreationTime, err := s[i].GetSnapshotDirectoryCreationTimeInUnix()
+	if err != nil {
+		logrus.Errorf("Failed to get snapshot directory creation time for snapshot: %s, with error: %v", path.Join(s[i].SnapDir, s[i].SnapName), err)
+	}
+	jCreationTime, _ := s[j].GetSnapshotDirectoryCreationTimeInUnix()
```
shreyas-s-rao: Why aren't we using the error here? The `err` on the next line will end up referring to the previous error from `iCreationTime` right?

swapnilgm: done.

PR #165, opened 2019-06-07
Fix error handling for initial delta snapshot

shreyas-s-rao:

This PR fixes the error handling while taking initial delta snapshot. Currently, if the initial delta snapshot fails, we retry the etcd probe and try taking the initial delta snapshot again, repeatedly until it succeeds. In the case where backup sidecar had already taken some snapshots and then stopped for some time while etcd is still running and is also compacted in that interval, when sidecar starts again, there's a chance the watch for initial delta snapshot will fail because the latest revision from the snapstore is not available anymore on etcd (as it was compacted). This throws backup sidecar into an infinite failure loop.
This PR fixes this behavior by taking a full snapshot on an initial delta snapshot error, instead of retrying delta snapshot again. Error handling for the full snapshot is as expected, ie, it will be retried after subsequent etcd probe.

Fixes #

**Release note**:

```improvement user
In the case that initial delta snapshot fails, a full snapshot is tried instead.
```

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -172,19 +172,12 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 				// the delta snapshot memory limit), after which a full snapshot
 				// is taken and the regular snapshot schedule comes into effect.
 
-				var startWithFullSnapshot bool
+				var initialDeltaSnapshotTaken bool
 
 				// TODO: write code to find out if prev full snapshot is older than it is
 				// supposed to be, according to the given cron schedule, instead of the
 				// hard-coded "24 hours" full snapshot interval
-				if ssr.PrevFullSnapshot == nil || time.Now().Sub(ssr.PrevFullSnapshot.CreatedOn).Hours() > 24 {
-					startWithFullSnapshot = false
-					if err = ssr.TakeFullSnapshotAndResetTimer(); err != nil {
-						logger.Errorf("Failed to take first full snapshot: %v", err)
-						continue
-					}
-				} else {
-					startWithFullSnapshot = true
+				if ssr.PrevFullSnapshot != nil && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= 24 {
 					ssrStopped, err := ssr.CollectEventsSincePrevSnapshot(ssrStopCh)
 					if ssrStopped {
 						logger.Info("Snapshotter stopped.")
@@ -193,16 +186,19 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 						logger.Info("Shutting down...")
 						return
 					}
-					if err != nil {
-						if etcdErr, ok := err.(*errors.EtcdError); ok == true {
-							logger.Errorf("Failed to take first delta snapshot: snapshotter failed with etcd error: %v", etcdErr)
-						} else {
-							logger.Errorf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
+					if err == nil {
+						if err := ssr.TakeDeltaSnapshot(); err != nil {
+							logger.Warnf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
+							continue
 						}
-						continue
+						initialDeltaSnapshotTaken = true
+					} else {
+						logger.Warnf("Failed to collect events for first delta snapshot: %v", err)
 					}
-					if err = ssr.TakeDeltaSnapshot(); err != nil {
-						logger.Errorf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
+				}
+				if !initialDeltaSnapshotTaken {
+					if err := ssr.TakeFullSnapshotAndResetTimer(); err != nil {
+						logger.Errorf("Failed to take substitute first full snapshot: %v", err)
 						continue
 					}
 				}
@@ -217,11 +213,11 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 				gcStopCh := make(chan struct{})
 				go ssr.RunGarbageCollector(gcStopCh)
 				logger.Infof("Starting snapshotter...")
-				if err := ssr.Run(ssrStopCh, startWithFullSnapshot); err != nil {
+				if err := ssr.Run(ssrStopCh, initialDeltaSnapshotTaken); err != nil {
 					if etcdErr, ok := err.(*errors.EtcdError); ok == true {
 						logger.Errorf("Snapshotter failed with etcd error: %v", etcdErr)
 					} else {
-						logger.Errorf("Snapshotter failed with error: %v", err)
+						logger.Fatalf("Snapshotter failed with error: %v", err)
 					}
 				}
 				logger.Infof("Snapshotter stopped.")


cmd/server.go
```diff
@@ -195,15 +195,23 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
 					if err != nil {
 						if etcdErr, ok := err.(*errors.EtcdError); ok == true {
-							logger.Errorf("Failed to take first delta snapshot: snapshotter failed with etcd error: %v", etcdErr)
+							logger.Warnf("Failed to take first delta snapshot: snapshotter failed with etcd error: %v", etcdErr)
 						} else {
-							logger.Errorf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
+							logger.Warnf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
+						}
+						startWithFullSnapshot = false
+						if err := ssr.TakeFullSnapshotAndResetTimer(); err != nil {
```
amshuman-kr: Why not `continue` only on delta-snapshot success and fall-through here?

cmd/server.go
```diff
@@ -195,15 +195,23 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
+							logger.Errorf("Failed to take substitute first full snapshot: %v", err)
+							continue
 						}
-						continue
 					}
 					if err = ssr.TakeDeltaSnapshot(); err != nil {
```
amshuman-kr: Why not a `TakeFullSnapshotAndResetTimer` here and skip the second call on error?

cmd/server.go
shreyas-s-rao: That would mean we wait for the snapshotter take the first snapshot (which will be a full snapshot) and set `handler.Status = http.StatusOK` before that, without actually having taken a snapshot before. We want to take atleast one snapshot (delta or full) before setting the readiness probe to true, so that we know that sidecar is able to take snapshots and is actually ready.

cmd/server.go
```diff
@@ -195,15 +195,23 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
 					}
 					if err != nil {
 						if etcdErr, ok := err.(*errors.EtcdError); ok == true {
```
swapnilgm: I don't find it making sense to have different log messages here. Let's just combine it.  Remove typecasting and if-else and have single log message on error.

cmd/server.go
swapnilgm: @shreyas-s-rao Agreed. But I guess Amshu's concern if more about complex if-else ladder here. If not atleaset i find it little complex. I think we can simplify it by reversing the if checks and have ssr.TakeFullSnapshotAndResetTimer() call at single place with minimum if-else conditions. e.g.  check at https://github.com/gardener/etcd-backup-restore/pull/165/files#diff-1427b112aa72f64adcf839007e954d56R180 can be reversed to evaluate else case first. Also, later on https://github.com/gardener/etcd-backup-restore/pull/165/files#diff-1427b112aa72f64adcf839007e954d56R196 this kind of checks can be changed to `err == nil`.  Please give it try.

cmd/server.go
```diff
@@ -195,15 +195,23 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
+							continue
 						}
-						continue
 					}
 					if err = ssr.TakeDeltaSnapshot(); err != nil {
-						logger.Errorf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
-						continue
+						logger.Warnf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
+						startWithFullSnapshot = false
+						if err := ssr.TakeFullSnapshotAndResetTimer(); err != nil {
```
swapnilgm: If taking delta snapshot failed means, there is issue with snapshot upload. And ssr.TakeFullSnapshotAndResetTimer() will also fail with same reason. So, it's better to fail here and retry for delta snapshot.

cmd/server.go
shreyas-s-rao: Yes, @amshuman-kr had the same suggestion in mind. I'll make this change. 👍🏼

cmd/server.go
shreyas-s-rao: Sure. I kept the error messages distinct because the same convention was followed here https://github.com/gardener/etcd-backup-restore/blob/8f32a9a72405589ae6a5d4d4192d002e1adb64f6/cmd/server.go#L221

cmd/server.go
shreyas-s-rao: Not necessarily. The primary motivation for this PR was to handle cases where the watch revision is not present in etcd due to compaction before backup sidecar starts. If delta snapshot fails due to a bucket error or network issue, then even on the next attempt there's a good chance the delta snapshot upload will fail. So I see no difference between retrying a delta snapshot and trying a full snapshot, because anyway full snapshot is done as chunk upload, so there's no issue of eating up network bandwidth as well.

cmd/server.go
swapnilgm: Agreed. But earlier we used to call logger.Fatalf for non etcderr / snapshot upload related error. I don;t remeber why we changed that. But now since you have pointed that out, it should be `Fatalf` at https://github.com/gardener/etcd-backup-restore/blob/8f32a9a72405589ae6a5d4d4192d002e1adb64f6/cmd/server.go#L224.

cmd/server.go
swapnilgm: Yes. but the time difference is more with full snapshot. That's the main point behind having this optimisation with first delta snapshot. On failure to take initial delta snapshot due cloud provider/network error, situation boil down to same point before etcd-backup-restore restart.

cmd/server.go
shreyas-s-rao: Ah ok. I'll make that change then.

cmd/server.go
shreyas-s-rao: Ah ok. There's one thing I can do. We know that the revision compaction issue is actually encountered by the `CollectEventsSincePrevSnapshot` call. If this call fails, we can attempt a full snapshot (because this call has nothing to do with the bucket itself, so provider/network errors won't occur here). But if it's an error with `TakeDeltaSnapshot`, then we re-attempt delta snapshot. WDYT?

swapnilgm: Yup, by re-attempt means - continue with the for loop; don't call TakeDeltaSnapshot again.

shreyas-s-rao:
@swapnilgm @amshuman-kr I have addressed the review comments. PTAL.

cmd/server.go
shreyas-s-rao: Yes, by re-attempt delta snapshot even I meant re-enter the loop.  :)

cmd/server.go
```diff
@@ -172,19 +172,12 @@ func NewServerCommand(stopCh <-chan struct{}) *cobra.Command {
...
 				// hard-coded "24 hours" full snapshot interval
-				if ssr.PrevFullSnapshot == nil || time.Now().Sub(ssr.PrevFullSnapshot.CreatedOn).Hours() > 24 {
-					startWithFullSnapshot = false
-					if err = ssr.TakeFullSnapshotAndResetTimer(); err != nil {
-						logger.Errorf("Failed to take first full snapshot: %v", err)
-						continue
-					}
-				} else {
-					startWithFullSnapshot = true
+				if ssr.PrevFullSnapshot != nil && time.Now().Sub(ssr.PrevFullSnapshot.CreatedOn).Hours() <= 24 {
```
swapnilgm: Please change this to `time.Since(ssr.PrevFullSnapshot.CreatedOn)`.

shreyas-s-rao: Done

shreyas-s-rao:
@swapnilgm I've addressed the review comment. PTAL.

PR #168, opened 2019-06-14
Proposal to handle high watch event rate.

### LINKED ISSUE #152: Memory optimizations for larger value sizes of etcd in backup-sidecar.
For large key-value sizes of etcd and watch on etcd yielding responses at a high rates, we see that the backup sidecar memory shoots up drastically but linearly.

profile015

georgekuruvillak:
This PR is a proposal to fix #152. 
Fixes #
#152

**Release note**:

```improvement operator

```


doc/proposals/high_watch_event_ingress_rate.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Problem statement
+For large key-value sizes of etcd and watch on etcd yielding responses at a high rates, we see that the backup sidecar memory shoots up drastically but linearly.
+
+![profile](../images/profile-high-rate-events.png)
+
+## Reason
+
+The memory consumption surge is associated with the watch implementation in the client library. 
+As can be seen [here](https://github.com/etcd-io/etcd/blob/master/clientv3/watch.go#L814), when the ingress rate of the events are more than the egress rate from the watch response channel the buffers holding the events tend to bloat. There is no throttling done between etcd and the client. Even if throttling was done, we would not be able to depend on it. As the revision of etcd would have moved ahead and due to throttling the watch events would be far behind. We would have to implement a different approach to watches if we stumble upon this issue in realtime.
```
amshuman-kr:
```suggestion
As can be seen [here](https://github.com/etcd-io/etcd/blob/master/clientv3/watch.go#L814), when the ingress rate of the events are more than the egress rate from the watch response channel the buffers holding the events tend to bloat. There is no throttling done between etcd and the client. Even if throttling was done, we would not be able to depend on it. As the revision of etcd would have moved ahead and due to throttling the watch events would be far behind. This lag could get progressively worse if the load on etcd persists. We would have to implement a different approach to watches if we stumble upon this issue in realtime.
```

PR #173, opened 2019-07-03
Added negative tests for restore of snapshots

ashwani2k:
Added more negative tests for restore of snapshots. 
Following test been added - 
1. Restore with embedded etcd quota not set 
2. Restore with invalid cluster URL
3. Restore with invalid restore directory
4. Restore with snapdir and snapname set to ""
5. Restore with invalid snapdir and snapname
6. Restore with no delta snapshots
7. Restore with corrupted snapstore
8. Restore with etcd client is unavailable
9. Restore with etcd data dir not cleaned up before restore
10. Restore with snapshotter running in parallel

Fixes partially #73 #104 

Enhanced the populate script to capture delete of keys.
Some refactoring is done to segregate the tests into 2 categories - 
1. Running restore with  pre-loaded snapstore  
2. Running restore with dynamic load for specific negative scenarios 

**Release note**:

```improvement operator
Added negative tests for restoration of snapshots.
```

### DIFF
--- a/pkg/snapshot/restorer/restorer_suite_test.go
+++ b/pkg/snapshot/restorer/restorer_suite_test.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"path"
 	"strconv"
 	"sync"
 	"testing"
@@ -101,6 +102,21 @@ func cleanUp() {
 
 	err = os.RemoveAll(snapstoreDir)
 	Expect(err).ShouldNot(HaveOccurred())
+
+	//for the negative scenario for invalid restoredir set to "" we need to cleanup the member folder in the working directory
+	restoreDir := path.Clean("")
+	info, err := os.Stat(fmt.Sprintf("%s/%s", restoreDir, "member"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return
+		} else {
+			Expect(err).ShouldNot(HaveOccurred())
+		}
+	}
+	if info.IsDir() {
+		err = os.RemoveAll(info.Name())
+		Expect(err).ShouldNot(HaveOccurred())
+	}
 }
 
 // startEmbeddedEtcd starts an embedded etcd server
--- a/pkg/snapshot/restorer/restorer_suite_test.go
+++ b/pkg/snapshot/restorer/restorer_suite_test.go
@@ -17,6 +17,7 @@ package restorer_test
 import (
 	"context"
 	"fmt"
+	"math"
 	"os"
 	"strconv"
 	"sync"
@@ -38,7 +39,7 @@ const (
 	etcdDir                    = outputDir + "/default.etcd"
 	snapstoreDir               = outputDir + "/snapshotter.bkp"
 	etcdEndpoint               = "http://localhost:2379"
-	snapshotterDurationSeconds = 100
+	snapshotterDurationSeconds = 20
 	keyPrefix                  = "key-"
 	valuePrefix                = "val-"
 	keyFrom                    = 1
@@ -72,7 +73,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	etcd, err = startEmbeddedEtcd(etcdDir, logger)
 	Expect(err).ShouldNot(HaveOccurred())
 	wg := &sync.WaitGroup{}
-	deltaSnapshotPeriod := 5
+	deltaSnapshotPeriod := 1
 	wg.Add(1)
 	go populateEtcd(wg, logger, endpoints, errCh, populatorStopCh)
 	go func() {
@@ -83,21 +84,24 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 		close(ssrStopCh)
 	}()
 
-	err = runSnapshotter(logger, deltaSnapshotPeriod, endpoints, ssrStopCh)
+	err = runSnapshotter(logger, deltaSnapshotPeriod, endpoints, ssrStopCh, true)
 	Expect(err).ShouldNot(HaveOccurred())
 
 	etcd.Server.Stop()
 	etcd.Close()
 	return data
+
 }, func(data []byte) {})
 
-var _ = SynchronizedAfterSuite(func() {}, func() {
+var _ = SynchronizedAfterSuite(func() {}, cleanUp)
+
+func cleanUp() {
 	err = os.RemoveAll(etcdDir)
 	Expect(err).ShouldNot(HaveOccurred())
 
 	err = os.RemoveAll(snapstoreDir)
 	Expect(err).ShouldNot(HaveOccurred())
-})
+}
 
 // startEmbeddedEtcd starts an embedded etcd server
 func startEmbeddedEtcd(dir string, logger *logrus.Logger) (*embed.Etcd, error) {
@@ -124,7 +128,7 @@ func startEmbeddedEtcd(dir string, logger *logrus.Logger) (*embed.Etcd, error) {
 }
 
 // runSnapshotter creates a snapshotter object and runs it for a duration specified by 'snapshotterDurationSeconds'
-func runSnapshotter(logger *logrus.Logger, deltaSnapshotPeriod int, endpoints []string, stopCh chan struct{}) error {
+func runSnapshotter(logger *logrus.Logger, deltaSnapshotPeriod int, endpoints []string, stopCh chan struct{}, startWithFullSnapshot bool) error {
 	var (
 		store                          snapstore.SnapStore
 		certFile                       string
@@ -177,7 +181,7 @@ func runSnapshotter(logger *logrus.Logger, deltaSnapshotPeriod int, endpoints []
 		snapshotterConfig,
 	)
 
-	return ssr.Run(stopCh, true)
+	return ssr.Run(stopCh, startWithFullSnapshot)
 }
 
 // populateEtcd sequentially puts key-value pairs into the embedded etcd, until stopped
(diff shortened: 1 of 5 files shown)


pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
+				err = rstr.Restore(restoreOptions)
+				Expect(err).Should(HaveOccurred())
+			})
+		})
 
-			err = checkDataConsistency(restoreDataDir, logger)
-			Expect(err).ShouldNot(HaveOccurred())
+		Context("with embedded etcd quota not set", func() {
+			fmt.Println("Testing for default embedded  etcd quota")
+			It("should set to defalut value of 8 GB and restore", func() {
```
amshuman-kr:
```suggestion
			It("should be set to defalut value of 8 GB and restore", func() {
```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
+				}
+				err = rstr.Restore(restoreOptions)
+				Expect(err).Should(HaveOccurred())
+
+			})
+		})
+
+		Context("with null snapdir and snapname", func() {
+			fmt.Println("Testing for nil snapdir and snapname")
+			It("should not fail and return nil as error", func() {
```
amshuman-kr: Is this a test for the default values for snapdir and snapname?

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
+
+				err = checkDataConsistency(restoreDataDir, logger)
+				Expect(err).ShouldNot(HaveOccurred())
+			})
+		})
+
+		Context("with maximum of hundred fetchers allowed", func() {
+			fmt.Println("Testing for max-fetchers=100")
+			It("should restore etcd data directory", func() {
+				maxFetchers = 100
```
amshuman-kr: Do we have 100 delta snapshots to be restored?

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
-			err = rstr.Restore(restoreOptions)
+			etcd, err = startEmbeddedEtcd(etcdDir, logger)
 			Expect(err).ShouldNot(HaveOccurred())
+			wg = &sync.WaitGroup{}
 
-			err = checkDataConsistency(restoreDataDir, logger)
-			Expect(err).ShouldNot(HaveOccurred())
+		})
+
+		AfterEach(func() {
```
amshuman-kr: How about `AfterEach(cleanUp)`? 😉

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
+					SkipHashCheck:          skipHashCheck,
+					Name:                   restoreName,
+					MaxFetchers:            maxFetchers,
+					EmbeddedEtcdQuotaBytes: embeddedEtcdQuotaBytes,
+					BaseSnapshot:           *baseSnapshot,
+					DeltaSnapList:          deltaSnapList,
+				}
+				logger.Infoln("starting restore restore directory exists already")
+				err = rstr.Restore(RestoreOptions)
+				Expect(err).Should(HaveOccurred())
```
amshuman-kr: Should we add a check that the existing directory is left untouched?

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
+					SkipHashCheck:          skipHashCheck,
+					Name:                   restoreName,
+					MaxFetchers:            maxFetchers,
+					EmbeddedEtcdQuotaBytes: embeddedEtcdQuotaBytes,
+					BaseSnapshot:           *baseSnapshot,
+					DeltaSnapList:          deltaSnapList,
+				}
+				logger.Infoln("starting restore while snapshotter is running")
+				err = rstr.Restore(RestoreOptions)
+				Expect(err).Should(HaveOccurred())
```
amshuman-kr: Should we add a check that the existing etcd data is left untouched?

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -215,17 +629,24 @@ func checkDataConsistency(dir string, logger *logrus.Logger) error {
...
-
 		resp, err := cli.Get(context.TODO(), key, opts...)
 		if err != nil {
 			return fmt.Errorf("unable to get value from etcd: %v", err)
 		}
 		if len(resp.Kvs) == 0 {
-			return fmt.Errorf("entry not found for key %s", resKey)
+			//handles deleted keys as every 10th key is deleted during populate etcd call
+			if math.Mod(float64(currKey), 10) == 0 {
+				continue //it should continue as key was put for action delete
```
amshuman-kr: Should we check that it is indeed deleted/not there?

pkg/snapshot/restorer/restorer_test.go
ashwani2k: the default duration to run snapshotter is 100s but when but as the delta snapshot period is 5s so in total this creates around 21 snapshots. we can change the deltaSnapshotPeriod to 1s and we will have 100 delta snapshots to apply. @shreyas-s-rao  -- wdyt?

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -215,17 +629,24 @@ func checkDataConsistency(dir string, logger *logrus.Logger) error {
...
+			} else {
```
shreyas-s-rao: Unnecessary `else`

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -215,17 +629,24 @@ func checkDataConsistency(dir string, logger *logrus.Logger) error {
...
+				res := resp.Kvs[0]
```
shreyas-s-rao: Will throw `index out of range` as length of `resp.Kvs` is 0 here.
```suggestion

```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -215,17 +629,24 @@ func checkDataConsistency(dir string, logger *logrus.Logger) error {
...
+				resKey = string(res.Key)
```
shreyas-s-rao:
```suggestion

```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -215,17 +629,24 @@ func checkDataConsistency(dir string, logger *logrus.Logger) error {
...
+				return fmt.Errorf("entry not found for key %s", resKey)
```
shreyas-s-rao:
```suggestion
				return fmt.Errorf("entry not found for key %s", key)
```
I agree there was a bug previously where we used to pass `resKey` here instead of `key`.

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -71,116 +79,522 @@ var _ = Describe("Running Restorer", func() {
...
+					DeltaSnapList:          deltaSnapList,
+				}
+				logger.Infoln("starting restore restore directory exists already")
+				err = rstr.Restore(RestoreOptions)
+				Expect(err).Should(HaveOccurred())
+
+			})
+		})
+
+		Context("while snapshooter is running ", func() {
```
shreyas-s-rao: Given that this test falls under the scope of server command rather than unit tests for restoration, because the server takes care of the coordination between snapshotter and restorer, I would advise moving this test to a different scope, maybe integration tests.

shreyas-s-rao:
```suggestion
		Context("while snapshotter is running ", func() {
```

pkg/snapshot/restorer/restorer_test.go
shreyas-s-rao: Yes. But as the behavior of restorer is different (restoration can succeed even without base snapshot, given the snapshotter was run without taking a base snapshot), this test would have to be modified so that snapshotter takes only delta snapshots for this test spec. @ashwani2k

ashwani2k: this test has now been moved under negative test - also a fix is included now in the GetLatestFullSnapshotAndDeltaSnapList() to return sorted the delta snapshots.

pkg/snapshot/restorer/restorer_test.go
ashwani2k: I've now made detlaSnapshotDuration to 20s as default with  deltaSnapshotPeriod  now set to 1s.

pkg/snapshot/restorer/restorer_test.go
ashwani2k: the test is now marked with 'X' and is excluded during the run. I've kept it for reference for refactoring at a later stage once agreed and aligned for handling such scenarios.

pkg/snapshot/restorer/restorer_suite_test.go
```diff
@@ -101,6 +102,21 @@ func cleanUp() {
 
 	err = os.RemoveAll(snapstoreDir)
 	Expect(err).ShouldNot(HaveOccurred())
+
+	//for the negative scenario for invalid restoredir set to "" we need to cleanup the member folder in the working directory
+	restoreDir := path.Clean("")
+	info, err := os.Stat(fmt.Sprintf("%s/%s", restoreDir, "member"))
```
shreyas-s-rao: No need to check for existence of the directory. You can simply run `os.RemoveAll`. If the directory doesn't exist, it'll return nil error anyway.

Also, use `path.Join(restoreDir, "member")` for joining `restoreDir` and `member`.

pkg/snapshot/restorer/restorer_suite_test.go
```diff
@@ -105,18 +105,19 @@ func cleanUp() {
 
 	//for the negative scenario for invalid restoredir set to "" we need to cleanup the member folder in the working directory
 	restoreDir := path.Clean("")
-	info, err := os.Stat(fmt.Sprintf("%s/%s", restoreDir, "member"))
+	// info, err := os.Stat(fmt.Sprintf("%s/%s", restoreDir, "member"))
+	info, err := os.Stat(path.Join(restoreDir, "member"))
```
shreyas-s-rao: You don't need `os.Stat()` itself. You can directly run `os.RemoveAll(path.Join(restoreDir, "member"))`.

pkg/snapshot/restorer/restorer_suite_test.go
```diff
@@ -105,18 +105,19 @@ func cleanUp() {
 
 	//for the negative scenario for invalid restoredir set to "" we need to cleanup the member folder in the working directory
 	restoreDir := path.Clean("")
-	info, err := os.Stat(fmt.Sprintf("%s/%s", restoreDir, "member"))
+	// info, err := os.Stat(fmt.Sprintf("%s/%s", restoreDir, "member"))
```
shreyas-s-rao: Also, please remove this commented line. Thanks.

PR #176, opened 2019-07-11
 Fix the OS_USER_DOMAIN_NAME auth issue

kayrus:

If the user was created in a different domain, then this required to auth against the keystone, otherwise auth will fail with the `Failed to create snapstore from configured storage provider: Authentication failed`

Fixes partially https://github.com/gardener/gardener/issues/1096

**Release note**:

```improvement operator
Etcd-backup-restore now uses the go modules for its dependecy management.
```


swapnilgm:
LGTM. Thank you for the PR.

PR #180, opened 2019-07-29
Initialize Prometheus metrics before exposing

### LINKED ISSUE #167: [BUG] Some metrics return no values
**Describe the bug:**
Prometheus [best practices](https://www.robustperception.io/existential-issues-with-metrics)  is to avoid missing metrics whenever possible. Some of the metrics that are exposed by the sidecar return no values if no event has occurred that would generate a value for this metric. These metrics should be initialized to 0 to prevent this problem. Some offending metrics are:
* `etcdbr_snapshot_gc_total`
* `etcdbr_restoration_duration_seconds_*`
* `etcdbr_defragmentation_duration_seconds_*`

**Expected behaviour:**
There should be no "missing metrics." All metrics should be initialized to 0 (or another value if 0 doesn't make sense).

shreyas-s-rao:

This PR initializes all Prometheus metrics to their respective zero-values before exposing them.

Fixes #167 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -15,6 +15,7 @@
 package metrics
 
 import (
+	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
@@ -34,6 +35,18 @@ const (
 )
 
 var (
+	labels = map[string][]string{
+		LabelKind: {
+			snapstore.SnapshotKindFull,
+			snapstore.SnapshotKindDelta,
+			snapstore.SnapshotKindChunk,
+		},
+		LabelSucceeded: {
+			ValueSucceededFalse,
+			ValueSucceededTrue,
+		},
+	}
+
 	// GCSnapshotCounter is metric to count the garbage collected snapshots.
 	GCSnapshotCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -105,7 +118,128 @@ var (
...
+	for i := 0; i < len(a); i++ {
+		for j := 0; j < len(b); j++ {
+			output = append(output, append(a[i], b[j]...))
+		}
+	}
+	return output
+}
+
+// wrapInSlice is a helper function to wrap a slice of strings within
+// a slice of slices of strings
```
swapnilgm:
```suggestion
// a slice of slices of strings
// e.g. [a,b,c] -> [[a],[b],[c]]
```

pkg/metrics/metrics.go
```diff
@@ -105,7 +118,128 @@ var (
...
+		return [][]string{}
+	} else if len(valuesList) == 1 {
+		return wrapInSlice(valuesList[0])
+	}
+
+	return cartesianProduct(wrapInSlice(valuesList[0]), getCombinations(valuesList[1:]))
+}
+
+// cartesianProduct combines two slices of slice of strings while also
+// combining the sub-slices of strings into a single string
```
swapnilgm:
```suggestion
// combining the sub-slices of strings into a single string
// e.g.
// a => [[p,q],[r,s]]
// b => [[1,2],[3,4]]
//  Output <= [[p,q,1,2],[p,q,3,4],[r,s,1,2],[r,s,3,4]]

pkg/metrics/metrics.go
shreyas-s-rao: Done.

pkg/metrics/metrics.go
shreyas-s-rao: Done.

shreyas-s-rao:
@swapnilgm Thanks for the review. I have addressed the changes and also added tests for the helper functions. PTAL.

PR #193, opened 2019-08-30
Expose etcdbr_snapshot_required metric

shreyas-s-rao:

This PR exposes new metric `etcdbr_snapshot_required` that exposes a Gauge metric to indicate whether a snapshot is required to be taken, based on whether a new etcd revision is observed.

Fixes #

Also contains minor linting corrections.

**Release note**:

```improvement developer
Expose new metric `etcdbr_snapshot_required`.
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -249,6 +249,8 @@ func (ssr *Snapshotter) takeFullSnapshot() error {
 
 		metrics.LatestSnapshotRevision.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.LastRevision))
 		metrics.LatestSnapshotTimestamp.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.CreatedOn.Unix()))
+		metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindFull}).Set(0)
+		metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindDelta}).Set(0)
 
 		ssr.logger.Infof("Successfully saved full snapshot at: %s", path.Join(s.SnapDir, s.SnapName))
 	}
@@ -325,6 +327,7 @@ func (ssr *Snapshotter) TakeDeltaSnapshot() error {
 	ssr.prevSnapshot = snap
 	metrics.LatestSnapshotRevision.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.LastRevision))
 	metrics.LatestSnapshotTimestamp.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.CreatedOn.Unix()))
+	metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindDelta}).Set(0)
 	ssr.logger.Infof("Successfully saved delta snapshot at: %s", path.Join(snap.SnapDir, snap.SnapName))
 	return nil
 }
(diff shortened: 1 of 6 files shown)


shreyas-s-rao:
@swapnilgm That check is only to see if a delta snapshot is required, so it doesn't check whether a full snapshot is required there (atleast at the moment). And again, this code path is taken only when we decide to start with delta snapshot, but not for the case where we start with full snapshot.

shreyas-s-rao:
@swapnilgm Thanks for your comments. I've updated the PR. PTAL.

PadmaB:
How about post restoration, if triggered? Shouldn't the metric be reset appropriately?

shreyas-s-rao:
@PadmaB Once restoration is complete, the snapshotter will start again, and we set the appropriate metric values there. So there's no need to explicitly reset the metrics after restoration.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -249,6 +249,7 @@ func (ssr *Snapshotter) takeFullSnapshot() error {
 
 		metrics.LatestSnapshotRevision.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.LastRevision))
 		metrics.LatestSnapshotTimestamp.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.CreatedOn.Unix()))
+		metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindFull}).Set(0)
```
swapnilgm: Shouldn't we also reset delta snapshot required metrics here?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -351,11 +353,23 @@ func (ssr *Snapshotter) CollectEventsSincePrevSnapshot(stopCh <-chan struct{}) (
...
+	}
 	if ssr.prevSnapshot.LastRevision == lastEtcdRevision {
 		ssr.logger.Infof("No new events since last snapshot. Skipping initial delta snapshot.")
 		return false, nil
 	}
 
+	// need to take a delta snapshot here, because etcd revision is
+	// newer than latest snapshot revision
+	metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindDelta}).Set(1)
+
```
swapnilgm: Also, mark full snapshot required to true right?

pkg/snapshot/snapshotter/snapshotter.go
shreyas-s-rao: You're right. Thanks for the catch

pkg/snapshot/snapshotter/snapshotter.go
shreyas-s-rao: Yes, thanks

shreyas-s-rao:
@swapnilgm Thanks for your review. I've addressed your comments. PTAL.

pkg/snapshot/snapshotter/snapshotter.go
swapnilgm: Actually move it out of after `else` block. It should be reset after both `if` and `else` path.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -351,11 +354,25 @@ func (ssr *Snapshotter) CollectEventsSincePrevSnapshot(stopCh <-chan struct{}) (
 	}
 	lastEtcdRevision := resp.Header.Revision
 
+	metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindFull}).Set(0)
+	metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindDelta}).Set(0)
+
+	// if etcd revision newer than latest full snapshot revision,
+	// set `required` metric for full snapshot to 1
+	if ssr.PrevFullSnapshot == nil || ssr.PrevFullSnapshot.LastRevision != lastEtcdRevision {
```
swapnilgm: We don't need this `if` block right? Check is already done on  next line.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -325,6 +327,7 @@ func (ssr *Snapshotter) TakeDeltaSnapshot() error {
 	ssr.prevSnapshot = snap
 	metrics.LatestSnapshotRevision.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.LastRevision))
 	metrics.LatestSnapshotTimestamp.With(prometheus.Labels{metrics.LabelKind: ssr.prevSnapshot.Kind}).Set(float64(ssr.prevSnapshot.CreatedOn.Unix()))
+	metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: snapstore.SnapshotKindDelta}).Set(0)
```
swapnilgm: Please add it at https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/snapshotter/snapshotter.go#L301 as well.

pkg/snapshot/snapshotter/snapshotter.go
shreyas-s-rao: Not necessary, because full snapshot required will be 0 if there are no revision changes. So there's no need to set it to 0 again in the `if` case.

pkg/snapshot/snapshotter/snapshotter.go
shreyas-s-rao: This check is for previous full snapshot, whereas next check is for previous snapshot (it can be full or delta). Even if there are no new events since previous delta snapshot, there could be new events since previous full snapshot (this is the case most of the time). So we need a separate check for previous full snapshot.

pkg/snapshot/snapshotter/snapshotter.go
shreyas-s-rao: Not necessary, because we know that the delta snapshot required will be 0 if there are no new events.

PR #195, opened 2019-09-02
Added credentials file and dir handling

afritzler:
This PR introduces 2 additional flags 
```
snapstore-credentials-file
snapstore-credentials-dir
```
This allows to load credentials from a file or a directory (e.g. mounted secret in `podSpec`).

This PR is related to the PR in the extension repo https://github.com/gardener/gardener-extensions/pull/287 and the umbrella issue https://github.com/gardener/gardener-extensions/issues/154

**Release note**:
```improvement operator
Added snapstore-credentials-file and snapstore-credentials-dir flags to support file and directory based credentials
```


swapnilgm:
Thank you for the PR and efforts.
But this PR looks more specific to reading openstack swift credentials from file, right?.
In that case, make use of environment variable combination `OS_CLIENT_CONFIG_FILE` and `OS_CLOUD`. By default it looks for `cloud.yaml` in current directory and choose first cloud spec from list. You can refer  [example file](https://github.com/gophercloud/utils/blob/94e6842399e53620efe002c58b46ab18f6b160bc/openstack/clientconfig/testing/clouds.yaml) and [go struct](https://github.com/gophercloud/utils/blob/94e6842399e53620efe002c58b46ab18f6b160bc/openstack/clientconfig/results.go) for file structure. 

I think PR https://github.com/gardener/gardener-extensions/pull/287 can easily adopt the proposed solution. 

For rest of provider also, we use common files supported by provider SDK or env variables. And we didn't wanted/felt necessary to introduce new credentials file format to keep it simple and generic. At least i prefer it the way it is for now.
WDYS? 

Or else this feature needs to well thought and those flags should be supported by all provider implementation. In that case please open up new feature request issue and we can have discussion there.
If you agree, is it ok to close this PR then?

afritzler:
That is actually an even better idea!

afritzler:
Thinking about it: how does the `cloud.yaml` file gets into the container? The only way I can think of is using a config map. Is that what we want?

swapnilgm:
First thing its should be secret not configmap since it has creds.
I think we can change  the source secret templated for openstack. while creating shoot and seed itself. cause otherwise inside code we will have to duplicate. CHanging it in source secret will lets allow simple code like gcp where we can map cloud.yaml to serviceaccount.json .

PR #200, opened 2019-09-09
Fix the revision consitency check error handling

swapnilgm:

- Because of incorrect error handling at revision consistency check, the etcd-backup-restore used to fail the validation and trigger restore for valid data directory. This was noticed when network was down at validation time but available at restore time. This PR fixes the issue.
- Additionally uses etcd API for etcd snap directory validation. We don't have to maintain code here anymore.
- Improves the test coverage for validator package.
Fixes #

**Release note**:

```improvement user
Fix the error handling in revision consistency check leading to restoration on valid etcd data directory.
```

### DIFF
--- a/pkg/initializer/validator/datavalidator_test.go
+++ b/pkg/initializer/validator/datavalidator_test.go
@@ -2,15 +2,16 @@ package validator_test
 
 import (
 	"fmt"
+	"math"
 	"os"
 	"path"
 
 	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 	"github.com/sirupsen/logrus"
 
 	. "github.com/gardener/etcd-backup-restore/pkg/initializer/validator"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Running Datavalidator", func() {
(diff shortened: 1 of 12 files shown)


pkg/initializer/validator/datavalidator_test.go
```diff
@@ -41,170 +42,244 @@ var _ = Describe("Running Datavalidator", func() {
...
-							validator.Config.SnapstoreConfig.Container = path.Join(snapstoreBackupDir, "tmp")
-							dataDirStatus, err := validator.Validate(Full, 1000000)
-							Expect(err).ShouldNot(HaveOccurred())
-							Expect(int(dataDirStatus)).Should(Equal(FailBelowRevisionConsistencyError))
-						})
-					})
-				})
+		})
+
+		Context("without providing snapstore config", func() {
```
shreyas-s-rao: This should be kept outside the `with fail below revision configured to high value` context, because whenever snapstore config is nil, validation will always return `DataDirectoryValid`, irrespective of other values (like failBelowRevision)

pkg/initializer/validator/datavalidator_test.go
```diff
@@ -41,170 +42,244 @@ var _ = Describe("Running Datavalidator", func() {
...
+			It("should return DataDirStatus as DataDirectoryValid and nil error", func() {
+				validator.Config.SnapstoreConfig = nil
+				dataDirStatus, err := validator.Validate(Sanity, failBelowRevision)
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(int(dataDirStatus)).Should(Equal(DataDirectoryValid))
 			})
 		})
 	})
+
+	Context("with failure on snapstore call", func() {
```
shreyas-s-rao:
```suggestion
	Context("with failure on snapstore call due to unknown snapstore provider", func() {
```
So that it'll be easy to differentiate between this and the next test context.

pkg/initializer/validator/datavalidator_test.go
```diff
@@ -41,170 +42,244 @@ var _ = Describe("Running Datavalidator", func() {
...
+		It("should return DataDirStatus as DataDirectoryStatusUnknown and error", func() {
+			//this is to fake the failure the snapstore call.
+			validator.Config.SnapstoreConfig.Provider = "unknown"
+			dataDirStatus, err := validator.Validate(Full, 0)
+			Expect(err).Should(HaveOccurred())
+			Expect(int(dataDirStatus)).Should(Equal(DataDirectoryStatusUnknown))
+		})
+	})
+
+	Context("with failure on snapstore call", func() {
```
shreyas-s-rao:
```suggestion
	Context("with failure on snapstore call due to fake failing snapstore provider", func() {
```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -460,55 +460,6 @@ var _ = Describe("Running Restorer", func() {
 			})
 		})
 
-		Context("with etcd client is unavailable ", func() {
```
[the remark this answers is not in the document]
swapnilgm: Etcd client up or no, it shouldn't matter for restore right. Restore just know that it has to restore to some directory. It doesn't matter anybody else should use it or not. The test is appropriate for server command or intialization request handler http API.

pkg/initializer/validator/datavalidator_test.go
swapnilgm: Done

swapnilgm:
Thank you for the review. I have updated the PR. PTAL.

shreyas-s-rao:
@swapnilgm Thanks for making the changes. LGTM.

PR #202, opened 2019-09-16
Fix restoration metric

shreyas-s-rao:

This PR fixes the bug where for a fresh etcd dir, restoration metric would be observed initially even though a restoration didn't actually happen. Because of this, the metric `etcdbr_restoration_duration_seconds_count` would be set to 1 instead of 0 when `etcdbrctl` process starts on a fresh data directory.

Fixes #

**Release note**:

```improvement operator
NONE
```


shreyas-s-rao:
@swapnilgm Thanks for your review. I've addressed the comments. PTAL.

PR #215, opened 2020-04-04
Fix integration test setup script

shreyas-s-rao:

This PR fixes recent apt issue for installing pip tool during integration test setup.

Fixes #

**Release note**:

```improvement developer
Fix integration test setup script.
```

### DIFF
--- a/.ci/integration_test
+++ b/.ci/integration_test
@@ -82,7 +82,12 @@ function setup_etcdbrctl(){
 
 function setup_awscli() {
     echo "Installing awscli..."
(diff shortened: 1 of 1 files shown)


shreyas-s-rao:
@swapnilgm I tried that out. But the unzipping step itself is taking around 2 minutes, whereas using the pip method takes less than 30 seconds. Keeping time constraint in mind, I would suggest let's keep the pip approach itself. WDYT?

swapnilgm:
Ahh ok. 👍

Issue #221, opened 2020-04-21
[Improvement] Skip first full snapshot if last full snapshot is recent enough


swapnilgm:
**Feature (What you would like to be added):**
Skip first full snapshot if last full snapshot is recent enough

**Motivation (Why is this needed?):**
Currently, once initial delta snapshot it taken, and backup sidecar marked healthy, we take full snapshot irrespective of last snapshot is recent enough or not.

**Approach/Hint to the implement solution (optional):**

PR #222, opened 2020-04-21
Skip the full snapshot if initial delta snapshot is taken and

swapnilgm:
last full snapshot is within 24hr

Fixes https://github.com/gardener/etcd-backup-restore/issues/221

**Release note**:

```improvement operator
Skip the first full snapshot on start,  if initial delta snapshot is taken and last full snapshot is recent enough than 24hr.
```

### DIFF
--- a/doc/usage/getting_started.md
+++ b/doc/usage/getting_started.md
@@ -16,7 +16,7 @@ The procedure to provide credentials to access the cloud provider object store v
 
 * For `Azure Blob storage`, `STORAGE_ACCOUNT` and `STORAGE_KEY` should be made available as environment variables.
 
-* For `Openstack Swift`, `OS_USERNAME`, `OS_PASSWORD`, `OS_AUTH_URL`, `OS_TENANT_ID` and `OS_DOMAIN_ID` should be made available as environment variables.
+* For `Openstack Swift`, `OS_USERNAME`, `OS_PASSWORD`, `OS_AUTH_URL`, `OS_TENANT_ID` and `OS_DOMAIN_ID` or `OS_CLOUD` should be made available as environment variables.
 
 * For `Alicloud OSS`, `ALICLOUD_ENDPOINT`, `ALICLOUD_ACCESS_KEY_ID`, `ALICLOUD_ACCESS_KEY_SECRET` should be made available as environment variables.
 
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -199,7 +199,9 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 		// TODO: write code to find out if prev full snapshot is older than it is
 		// supposed to be, according to the given cron schedule, instead of the
 		// hard-coded "24 hours" full snapshot interval
-		if ssr.PrevFullSnapshot != nil && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= 24 {
+		const recentFullSnapshotPeriodInHours = 24
+		initialDeltaSnapshotTaken = false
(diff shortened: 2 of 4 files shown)


shreyas-s-rao:
Can we reduce the full snapshot "recent" duration to something like 15 or 30 minutes? So that sidecar can will take full snapshot only if last full snapshot was more than 30 minutes ago. This is because when etcd starts restarting, there's a higher chance of a data corruption and restoration, and this restoration will take longer if the last full snapshot was taken too long ago. A full snapshot taken 30 minutes ago will have far lesser number of delta snapshots, so the restoration will be quick and there's a smaller downtime there. WDYT?

swapnilgm:
I think restoration time is separate problem. Let's not try to solve it here. Currently we are find with whatever RPO its provides. The restoration time in generic issue and should be solved with compaction idea that we have in backlog.

swapnilgm:
@shreyas-s-rao @georgekuruvillak If you don't have any further concerns, can we merge this?

swapnilgm:
I have updated the bootstrap script. Please go through it once and merge this.

doc/usage/getting_started.md
```diff
@@ -16,7 +16,7 @@ The procedure to provide credentials to access the cloud provider object store v
 
 * For `Azure Blob storage`, `STORAGE_ACCOUNT` and `STORAGE_KEY` should be made available as environment variables.
 
-* For `Openstack Swift`, `OS_USERNAME`, `OS_PASSWORD`, `OS_AUTH_URL`, `OS_TENANT_ID` and `OS_DOMAIN_ID` should be made available as environment variables.
+* For `Openstack Swift`, `OS_USERNAME`, `OS_PASSWORD`, `OS_AUTH_URL`, `OS_TENANT_ID` and `OS_DOMAIN_ID` or `OS_CLOUD` should be made available as environment variables.
```
shreyas-s-rao, line 19: Who picks up `OS_CLOUD` env var? I don't see gophercloud API supporting this, atleast in our vendored files. Am I missing something?

swapnilgm: It does support it. In-fact we were using it before etcd-druid integration in gardener. In webhook for openstack.

pkg/server/backuprestoreserver.go
```diff
@@ -238,7 +240,8 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 		gcStopCh := make(chan struct{})
 		go ssr.RunGarbageCollector(gcStopCh)
 		b.logger.Infof("Starting snapshotter...")
-		if err := ssr.Run(ssrStopCh, initialDeltaSnapshotTaken); err != nil {
+		startWithFullSnapshot := !(initialDeltaSnapshotTaken && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= recentFullSnapshotPeriodInHours)
```
georgekuruvillak, line 243: I have a query:
Suppose there was no change in the revision since the last full snapshot, then in this function it takes a full snapshot and proceeds taking delta snapshots. Say, etcd gets OOM killed here. Wont the cycle reappear as we wont take delta snapshots as [here](https://github.com/gardener/etcd-backup-restore/pull/222/files#diff-63ebe6acc9f30c180e7bf89c61edbc97R223) we are not checking for recent full snapshot. If deltaSnaps are not there and if full snapshots are recent then how do we proceed?

swapnilgm: Nope.
On next restart, it will attempt delta snap first. So, `initialDeltaSnapshotTaken` will be true and last full snap will be within 24hr. So, it won't take full snapshot immediately but as per the full snap schedule.

swapnilgm:
if your queries has been resolved. Will you merge this PR.

PR #248, opened 2020-07-16
Use channel id for posting slack release notes

petersutter:
Use the channel id for posting slack release notes so that the definition does not need to be adapted in case the channel will be renamed

Fixes #

**Release note**:

```improvement operator

```


petersutter:
yes `scp_workspace` still works and it's api_token is the same as for `ti_workspace`. So you could also change `scp_workspace` to `ti_workspace`

PR #249, opened 2020-07-23
Integrate with Test Machinery

### LINKED ISSUE #107: Integrate etcd tests with TestMachinery
# Story
As etcd backup-restore developer I want to integrate the existing tests with the TestMachinery to run the tests in an integrated environment and also be able automate the same

# Motivation
TestMachinery provides a robust test environment, automates cluster creation and such operations making it is easy for developers to run their tests in an integrated environment, publishing the results.

# Acceptance Criteria
- [ ] Integrate the existing tests into TestMachinery
- [ ] Ensure the end-to-end automation
- [ ] Validate the results

# Definition of Done
- [ ] Knowledge is distributed: Have you spread your knowledge in pair programming/code review?
- [ ] Unit tests are provided: Have you written automated unit tests?
- [ ] Integration tests are provided: Have you written automated integration tests?
- [ ] Minimum API exposure: If you have added/changed public API, was it really necessary/is it minimal?
- [ ] Operations guide: Have you updated the operations guide about ops-relevant changes?
- [ ] User documentation: Have you updated the READMEs/docs/how-tos about user-relevant changes?

shreyas-s-rao:

This PR adds support for running integration tests on [TestMachinery](https://github.com/gardener/test-infra) framework. The tests can be triggered from the TM Concourse dashboard, so that integration tests are run on a shoot cluster spawned by TestMachinery.

Fixes #107 

**Release note**:

```improvement developer
Add TestMachinery integration.
```

### DIFF
--- a/.ci/integration_test
+++ b/.ci/integration_test
@@ -27,8 +27,10 @@ VCS="github.com"
 ORGANIZATION="gardener"
 PROJECT="etcd-backup-restore"
 REPOSITORY=${VCS}/${ORGANIZATION}/${PROJECT}
+URL=https://${REPOSITORY}.git
 VERSION_FILE="$(readlink -f "${SOURCE_PATH}/VERSION")"
 VERSION="$(cat "${VERSION_FILE}")"
+TEST_ID_PREFIX="etcdbr-test"
 
 export GOBIN="${SOURCE_PATH}/bin"
 export PATH="${GOBIN}:${PATH}"
@@ -39,6 +41,7 @@ cd "${SOURCE_PATH}"
 # Declare global variables
 TEST_ID=
 ETCD_VER=
+ETCDBR_VER=
 ETCD_DATA_DIR=
 TEST_DIR=
 TEST_RESULT=
@@ -96,7 +99,14 @@ function setup_awscli() {
 
 function get_test_id() {
   git_commit=`git show -s --format="%H"`
-  export TEST_ID=etcdbr-test-${git_commit}
+  export TEST_ID=${TEST_ID_PREFIX}-${git_commit}
+  echo "Test id: ${TEST_ID}"
+}
+
+function get_latest_commit_test_id() {
+  git_commit_remote=`echo \`git ls-remote $URL HEAD\` | cut -d' ' -f 1`
+  export ETCDBR_VER=${VERSION}-${git_commit_remote}
+  export TEST_ID=${TEST_ID_PREFIX}-${git_commit_remote}
   echo "Test id: ${TEST_ID}"
 }
 
@@ -109,6 +119,20 @@ function create_etcd_data_directory() {
 #        AWS Setup          #
 #############################
 
+function write_aws_secret() {
+  echo "Creating aws credentials for API access..."
+  mkdir ${HOME}/.aws
+  cat << EOF > ${HOME}/.aws/credentials
+[default]
+aws_access_key_id = $1
+aws_secret_access_key = $2
+EOF
+  cat << EOF > ${HOME}/.aws/config
+[default]
+region = $3
+EOF
+}
+
 function create_aws_secret() {
   echo "Fetching aws credentials from secret server..."
   export ACCESS_KEY_ID=`/cc/utils/cli.py config attribute --cfg-type aws --cfg-name etcd-backup-restore --key access_key_id`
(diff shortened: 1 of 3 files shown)


.ci/integration_test
```diff
@@ -96,7 +99,14 @@ function setup_awscli() {
 
 function get_test_id() {
   git_commit=`git show -s --format="%H"`
-  export TEST_ID=etcdbr-test-${git_commit}
+  export TEST_ID=${TEST_ID_PREFIX}-${git_commit}
+  echo "Test id: ${TEST_ID}"
+}
+
+function get_latest_commit_test_id() {
+  git_commit_remote=`echo \`git ls-remote $URL HEAD\` | cut -d' ' -f 1`
```
swapnilgm, line 107: Won't this list all PR commits? 
What's the difference between `get_test_id` function and this one?

shreyas-s-rao: It doesn't list PR commits. Just the latest commit on master branch on the remote repo. Reason for this is, we won't be able to run TM tests on every PR because it requires the image to be built from the current PR build. And for this to happen, the CI pipeline needs to execute the `publish` trait. After playing around with the CI definition, I realised there's no way to execute a step post `publish` trait (I may be wrong here, so would love your help on this). So I decided to populate test_id based on latest commit on remote repo's master branch. With the next PR for CI integration, I'll modify the test script to use the latest release version for the etcdbrctl image.

The other approach is to build a new docker image from the PR repo and push to registry, but that's quite tedious and unnecessary I felt. Because anyway we won't be running TM tests on every PR build, but rather on a head-update or even a cron basis. WDYS?

swapnilgm: > It doesn't list PR commits.
right. sorry my mistake. 

Ok got the intention behind running TM only on head update. That's totally fine.

but i still didn't understand, since the head will be checked out locally,  `get_test_id` function will give expected test_id and git_commit sha. isn't this function redundant then?

shreyas-s-rao: Actually, I'm thinking we can't run it on head update as well, since in the pipeline we can't run a `step` *after* the `publish` trait. So I think it might be better to run the TM tests periodically, outside the scope of backup-restore CI pipeline, ie, the TM framework itself would trigger the etcd-backup-restore tests on a cron basis. In that case, since the test would have to pull the latest image from registry (which was built using the latest commit on the remote repo), I'm using a separate `get_latest_commit_test_id` for that. In fact, the `test_id` that we generate is only used to specify the bucket name. We could even generate a random ID for it, and since the test always cleans up the bucket, we wouldn't have to worry about `test_id` at all. WDYS?

swapnilgm: > Actually, I'm thinking we can't run it on head update as well, since in the pipeline we can't run a step after the publish trait.

We can run `step` after `publish`. You can make use of `depedes` condition mentioned at https://gardener.github.io/cc-utils/pipeline_step.html .

>  I'm using a separate get_latest_commit_test_id for that.

The test script is available locally means the repository is checked out locally. No, please remove   `get_latest_commit_test_id` and use `get_test_id`. Output of both is same in all scenario. But get_latest_commit_test_id unnecessary calls remote.

> the test_id that we generate is only used to specify the bucket name. We could even generate a random ID for it, and since the test always cleans up the bucket, we wouldn't have to worry about test_id at all.

Yes. But when the test scripts fails because of some bug or random event. Since test_id will map to commit, it will be easy to debug. That's why we didn't and should not go for random id in first place.

shreyas-s-rao: > We can run step after publish. You can make use of depedes condition mentioned at https://gardener.github.io/cc-utils/pipeline_step.html .

`depends` only makes sure that `publish` trait runs after the dependency is satisfied, right?

.ci/integration_test
```diff
@@ -248,10 +272,38 @@ function run_test_on_cluster() {
...
 
-if [ "$1" == "cluster" ]; then
+function run_test_on_tm() {
+  if [ "$ACCESS_KEY_ID" == "" ] || [ "$SECRET_ACCESS_KEY" == "" ] || [ "$REGION" == "" ] ; then
+    echo "AWS S3 credentials unavailable. Exiting."
+    exit 1
+  fi
+
+  # TODO: remove
+  echo "Access ID: ${ACCESS_KEY_ID}, Access key: ${SECRET_ACCESS_KEY}, Region: ${REGION}"
```
swapnilgm, line 282: Remove this.

.ci/integration_test
```diff
@@ -136,7 +150,15 @@ function delete_aws_secret() {
 }
 
 function create_s3_bucket() {
-    aws s3api create-bucket --bucket ${TEST_ID} --region ${REGION} --create-bucket-configuration LocationConstraint=${REGION}
+  echo "Creating S3 bucket ${TEST_ID} in region ${REGION}"
+  # TODO: remove
+  echo "Access ID: ${ACCESS_KEY_ID}, Access key: ${SECRET_ACCESS_KEY}"
```
swapnilgm, line 155: Remove this.

.ci/integration_test
shreyas-s-rao: I just added for testing something. Not to worry, I've removed it.

shreyas-s-rao:
@swapnilgm PR is rebased and ready now.

PR #261, opened 2020-09-24
Add support for OCS S3 snapstore

stoyanr:
Adds support for OpenShift Container Storage (OCS) S3-compatible snapstore. Also refactors the existing ECS snapstore to avoid code duplication since the 2 snapstores are quite similar.

Fixes #

See also https://github.com/gardener/etcd-druid/pull/98.

**Release note**:
```improvement operator
Added support for OpenShift Container Storage (OCS) S3 storage type.
```

### DIFF
--- a/pkg/snapstore/generic_s3_snapstore.go
+++ b/pkg/snapstore/generic_s3_snapstore.go
@@ -1,0 +1,60 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


pkg/snapstore/generic_s3_snapstore.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// s3AuthOptions contains all needed options to authenticate against a S3-compatible store.
+type s3AuthOptions struct {
+	endpoint           string
+	region             string
+	disableSSL         bool
+	insecureSkipVerify bool
+	accessKeyID        string
+	secretAccessKey    string
```
rootfs, line 35: need an `enableEncrytion` option?

stoyanr: No, this has been removed, see gardener/gardener-extension-provider-kubevirt#59. Encryption is not supported by OCS, correct? It only made sense in the context of a "generic" S3-compatible store, but this is gone now.

rootfs (approved):

amshuman-kr:
@stoyanr Thanks a lot for the PR and the new provider support! We were held up with some landscape issues and review already this week. I apologize.

shreyas-s-rao:
Hi @stoyanr. Thanks for your PR. Will review it today, since I was busy the past few days.

amshuman-kr (approved):
@stoyanr Thanks a lot for the PR and sorry for the delay in the review! The PR looks good. Would it be possible for you to give access to an OCS environment for us to test before merging?

PR #270, opened 2020-10-08
Enabled check script to run on mac and for normal user.

abdasgupta:
Check script was showing error when the user was not root. This
commit fixes that. Also, the check script was not running on mac
due to readlink error. The error is fixed as well.

Check script was showing error when the user was not root. This
commit fixes that. Also, the check script was not running on mac
due to readlink error. The error is fixed as well.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/.ci/check
+++ b/.ci/check
@@ -16,11 +16,16 @@ set -e
 
 # For the check step concourse will set the following environment variables:
 # SOURCE_PATH - path to component repository root directory.
+if [[ $(uname) == 'Darwin' ]]; then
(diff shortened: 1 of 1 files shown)


abdasgupta:
cc @amshuman-kr

amshuman-kr (approved):
Thanks! LGTM. Could you please sign the CLA?

abdasgupta:
cc @shreyas-s-rao

PR #272, opened 2020-10-14
Fix missing alternate full snapshots

### LINKED ISSUE #258: [Feature] Retry full backups to avoid false positives in alerts
If regular full snapshot backup fails, they are retried only in the next interval (typically, `24h`). We should retry the full snapshot backup in shorter time frame (say, `10m`, `15m` or `20m`).

Alerts are configured to fire if the latest full backups are more than `24h` old. These typically get resolved automatically in the next interval when the full backup goes through. So, most of the alerts are false positives that get resolved automatically. This makes it hard to automate the follow up process using ticketing systems (mandated by audit).

Retrying within a range of `10m-20m` might resolve the issue automatically earlier so that alerts fire only if retries also fail.

shreyas-s-rao:

This PR fixes missing full snapshots on shoot unhibernation. Setting the previous full snapshot existence check to 23.5 hours provides a buffer of 30 minutes for the pod to become ready to take the first full snapshot everyday in the case where full snapshot schedule lies within the shoot's hibernation period.

Fixes #258 

**Release note**:

```improvement operator
Fix missing alternate full snapshots for some unhibernating shoots.
```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -199,7 +199,11 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 		// TODO: write code to find out if prev full snapshot is older than it is
 		// supposed to be, according to the given cron schedule, instead of the
 		// hard-coded "24 hours" full snapshot interval
-		const recentFullSnapshotPeriodInHours = 24
+
+		// Temporary fix for missing alternate full snapshots for Gardener shoots
+		// with hibernation schedule set: change value from 24 ot 23.5 to
+		// accommodate for slight pod spin-up delays on shoot wake-up
+		const recentFullSnapshotPeriodInHours = 23.5
 		initialDeltaSnapshotTaken = false
 		if ssr.PrevFullSnapshot != nil && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= recentFullSnapshotPeriodInHours {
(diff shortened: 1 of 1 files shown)


amshuman-kr:
Thanks for the PR, @shreyas-s-rao! The change itself looks good. But as discussed elsewhere, how about addressing this by tweaking the gardener to consider hibernation schedule while computing the backup schedule string? It is already considering maintenance schedule.

shreyas-s-rao:
@amshuman-kr as discussed, we can go ahead with this PR as it's a quick win, while we do acknowledge that it's only a temporary solution, and we need a more generic approach to determine whether a "recent enough" previous full snapshot was taken, according to the given cron schedule.

amshuman-kr (approved):
LGTM

shreyas-s-rao:
@amshuman-kr I have tested this locally with a helm chart, where I simulated a delayed startup (by 10 minutes) of etcd on day 1. Without this PR, the day 2 etcdbr startup missed the full snapshot. With the PR, it was able to account for that 10-minute delay the previous day, and took the full snapshot on etcdbr startup. Not sure if we can have an automated test case for this, since the config values are more Gardener-specific (like 24 hours). Until we automate that, I don't see a way to add a test case for this scenario. I'll open an issue to move away from the hard-coded 24-hour value [here](https://github.com/gardener/etcd-backup-restore/blob/6b8bf4d501195b2790e47ad333ac340cfdc3f368/pkg/server/backuprestoreserver.go#L202) towards a more cron-schedule-aware method to calculate the period between two consecutive full snapshots. For the time-being, I think we can go ahead and merge this PR. WDYT?

PR #273, opened 2020-10-14
Fixed an error in unit test cases that appears due to use of latest golang.

abdasgupta:
An error is appearing printing this: conversion from int to string yields a
string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
This error is appearing while running unit tests with golang version
1.15 . This error is fixed in this PR.

Fixes #

**Release note**:

```improvement operator

```


abdasgupta:
cc @shreyas-s-rao @amshuman-kr

shreyas-s-rao:
@abdasgupta Thanks for the PR. Nice catch. Since etcdbrctl is compiled with golang 1.14 right now, it was never tested on v1.15. Thanks for finding this. Did you also try building the binary (`make build` and `make build-local`) using golang 1.15?

abdasgupta:
@shreyas-s-rao Yeah.. `make build` and `make build-local` built the binary successfully using golang 1.15

amshuman-kr (approved):
LGTM

PR #275, opened 2020-10-23
Double check latest revision by starting an embedded etcd if revision check based on data dir fails during data validation 

### LINKED ISSUE #260: [BUG] Database verification should check WALs also for latest revision
**Describe the bug:**
The database validation logic currently [only checks the bolt DB data](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/initializer/validator/datavalidator.go#L265-L301) for the latest revision in the etcd directory.

It is possible that  etcd has terminates abnormally and fails to flush the changes from the WALs to the bolt DB but if the latest backup (full or delta) already have backed up some of the changes that are still in the WALs. If this happens, the data validation might see a revision in the etcd data directory (based only on the bold DB) that is older than the revision that has already been backed up. This will lead to an unnecessary data restoration (with possible data loss if not all the WALs where backed up).

**Expected behavior:**
The WALs should also be checked while calculating the latest revision in the etcd directory so that the situation where the latest revision in the backup is more recent than the revision that the validation logic sees in the etcd data directory.

Hard to reproduce.

**Logs:**
Logs in the backup-restore sidecar like below.
```
current etcd revision (2314180238) is less than latest snapshot revision (2314180239): possible data loss
```

 - Etcd version/commit ID :

ishan16696:
WALs file can have data revisions which are ahead of data revisions present in DB file, if etcd terminates abnormally and fails to flush the changes from WALs to Bolt DB ,this lead to  unnecessary data restoration.
This PR eliminates this unnecessary data restoration. Starting the embedded etcd on data dir, ping the embedded etcd for 60s to get the latest revision from etcd, after that compare this latest revision and latest revision of back-up data.

Fixes #260 

**Release note**:

```improvement operator
Validator now double checks latest revision by starting an embedded etcd if DB-based revision check fails. This can potentially avoid unnecessary data restoration when etcd terminates abnormally.
```

### DIFF
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -15,6 +15,7 @@
 package validator
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -22,11 +23,14 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/gardener/etcd-backup-restore/pkg/miscellaneous"
+	"github.com/gardener/etcd-backup-restore/pkg/snapshot/restorer"
 	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
 	"github.com/sirupsen/logrus"
 	bolt "go.etcd.io/bbolt"
+	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/etcdserver/api/snap"
 	"go.etcd.io/etcd/raft/raftpb"
 	"go.etcd.io/etcd/wal"
@@ -114,8 +118,19 @@ func (d *DataValidator) sanityCheck(failBelowRevision int64) (DataDirStatus, err
 		d.Logger.Infof("unable to get current etcd revision from backend db file: %v", err)
 		return DataDirectoryCorrupt, nil
 	}
-	d.Logger.Info("Checking for revision consistency...")
-	return d.checkRevisionConsistency(etcdRevision, failBelowRevision)
+
+	d.Logger.Info("Checking for etcd revision consistency...")
+	etcdRevisionStatus, latestSnapshotRevision, err := d.checkEtcdDataRevisionConsistency(etcdRevision, failBelowRevision)
+
+	// if etcd revision is inconsistent with latest snapshot revision then
+	//   check the etcd revision consistency by starting an embedded etcd since the WALs file can have uncommited data which it was unable to flush to Bolt DB.
+	if etcdRevisionStatus == RevisionConsistencyError {
+		d.Logger.Info("Checking for Full revision consistency...")
+		fullRevisionConsistencyStatus, err := d.checkFullRevisionConsistency(dataDir, latestSnapshotRevision)
+		return fullRevisionConsistencyStatus, err
+	}
+
+	return etcdRevisionStatus, err
 }
 
 // checkForDataCorruption will check for corruption of different files used by etcd.
@@ -229,18 +244,20 @@ func verifyDB(path string) error {
 	})
 }
 
-// checkRevisionConsistency compares the latest revisions on the etcd db file and the latest snapshot to verify that the etcd revision is not lesser than snapshot revision.
-// Return DataDirStatus indicating whether it is due to failBelowRevision or latest snapshot revision for snapstore.
-func (d *DataValidator) checkRevisionConsistency(etcdRevision, failBelowRevision int64) (DataDirStatus, error) {
+// checkEtcdDataRevisionConsistency compares the latest revision of the etcd db file and the latest snapshot revision to verify that the etcd revision is not lesser than snapshot revision.
+// Return DataDirStatus indicating whether it is due to failBelowRevision or latest snapshot revision for snapstore and also return the latest snapshot revision.
+func (d *DataValidator) checkEtcdDataRevisionConsistency(etcdRevision, failBelowRevision int64) (DataDirStatus, int64, error) {
+	var latestSnapshotRevision int64
+	latestSnapshotRevision = 0
+
 	store, err := snapstore.GetSnapstore(d.Config.SnapstoreConfig)
 	if err != nil {
-		return DataDirectoryStatusUnknown, fmt.Errorf("unable to fetch snapstore: %v", err)
+		return DataDirectoryStatusUnknown, latestSnapshotRevision, fmt.Errorf("unable to fetch snapstore: %v", err)
 	}
 
-	var latestSnapshotRevision int64
 	fullSnap, deltaSnaps, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(store)
 	if err != nil {
-		return DataDirectoryStatusUnknown, fmt.Errorf("unable to get snapshots from store: %v", err)
+		return DataDirectoryStatusUnknown, latestSnapshotRevision, fmt.Errorf("unable to get snapshots from store: %v", err)
 	}
 	if len(deltaSnaps) != 0 {
 		latestSnapshotRevision = deltaSnaps[len(deltaSnaps)-1].LastRevision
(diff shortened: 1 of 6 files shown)


pkg/initializer/validator/datavalidator.go
```diff
@@ -250,17 +265,73 @@ func (d *DataValidator) checkRevisionConsistency(etcdRevision, failBelowRevision
...
+		e.Close()
+	}()
+	client, err := clientv3.NewFromURL(e.Clients[0].Addr().String())
+	if err != nil {
+		d.Logger.Infof("unable to get the embedded etcd client: %v", err)
+		return DataDirectoryCorrupt, err
+	}
+
+	start := time.Now()
+	timer := time.NewTimer(60 * time.Second)
```
amshuman-kr, line 304: Can you please a constant for the timeout?

pkg/initializer/validator/datavalidator.go
```diff
@@ -250,17 +265,73 @@ func (d *DataValidator) checkRevisionConsistency(etcdRevision, failBelowRevision
...
+	for {
```
amshuman-kr, line 305: How about simplifying the loop like this?
```go
       defer timer.Stop()
       waitLoop: for {
		select {
		case <-timer.C:
			break waitLoop
		default:
			latestSyncedEtcdRevision, _ = getLatestSyncedRevision(client)
			if latestSyncedEtcdRevision >= latestSnapshotRevision {
				d.Logger.Infof("After starting embeddedEtcd backend DB file revision (%d) is greater than or equal to latest snapshot revision (%d): no data loss", latestSyncedEtcdRevision, latestSnapshotRevision)
				break waitLoop
			}
			time.Sleep(1 * time.Second)
		}
	}
```

amshuman-kr (changes requested):
Thanks a lot for the PR @ishan16696! Just some small suggestions.

Also, do you plan to include the test (copying old data directory and restoring it)?

BTW Could you please sign the CLA?

pkg/initializer/validator/datavalidator.go
ishan16696: I have made all changes which you suggested 😊

pkg/initializer/validator/datavalidator.go
```diff
@@ -250,17 +265,68 @@ func (d *DataValidator) checkRevisionConsistency(etcdRevision, failBelowRevision
...
+		e.Server.Stop()
+		e.Close()
+	}()
+	client, err := clientv3.NewFromURL(e.Clients[0].Addr().String())
+	if err != nil {
+		d.Logger.Infof("unable to get the embedded etcd client: %v", err)
+		return DataDirectoryCorrupt, err
+	}
+
+	timer := time.NewTimer(embeddedEtcdPingLimit * time.Second)
```
amshuman-kr, line 303:
```suggestion
	timer := time.NewTimer(embeddedEtcdPingLimitDuration)
```

pkg/initializer/validator/types.go
```diff
@@ -34,14 +34,18 @@ const (
...
 	RevisionConsistencyError
 	//FailBelowRevisionConsistencyError indicates the current etcd revision is inconsistent with failBelowRevision.
 	FailBelowRevisionConsistencyError
 )
 
 const (
-	snapSuffix = ".snap"
+	snapSuffix            = ".snap"
+	connectionTimeout     = 10
+	embeddedEtcdPingLimit = 60
```
amshuman-kr, line 48:
```suggestion
	embeddedEtcdPingLimitDuration = 60 * time.Second
```

amshuman-kr (changes requested):
Thanks for the changes! One small change again :-)

pkg/initializer/validator/types.go
ishan16696: should i also change the `connectionTimeout = 10 * time.Second`?

pkg/initializer/validator/datavalidator.go
```diff
@@ -300,3 +366,19 @@ func getLatestEtcdRevision(path string) (int64, error) {
 
 	return rev, nil
 }
+
+// getLatestSyncedRevision finds out the latest DB revision when WAL have some uncommitted data and when embedded etcd is started.
+func getLatestSyncedRevision(client *clientv3.Client) (int64, error) {
+	var latestSyncRevision int64
+
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Duration(connectionTimeout))
```
amshuman-kr, line 374:
```suggestion
	ctx, cancel := context.WithTimeout(context.TODO(), connectionTimeout)
```

amshuman-kr (changes requested):
One small change again 😉

amshuman-kr:
> Hi @ishan16696, I feel the commits are not properly named and divided. Ideally, the commit names should tell what are the changes about. Also, if the changes in same file are doing same work, the changes should come under same commit.

Yes. I was waiting for the tests and then going to request to [squash the commits into a single commit](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) and [force push](https://git-scm.com/docs/git-push#Documentation/git-push.txt---force) to the [PR branch](https://github.com/ishan16696/etcd-backup-restore) before we merge this PR.

Typically, most PRs have only a single commit. If the PR includes multiple logically separable functionality then there can be separate commits as @abdasgupta mentioned. This is rare. One example, could be revendoring changes as part of some other dependency changes in the PR.

pkg/initializer/validator/types.go
```diff
@@ -34,14 +36,18 @@ const (
 	DataDirectoryCorrupt
 	// DataDirectoryStatusUnknown indicates validator failed to check the data directory status.
 	DataDirectoryStatusUnknown
-	// RevisionConsistencyError indicates current etcd revision is inconsistent with latest snapshot revision.
+	// EtcdRevisionConsistencyError indicates current etcd revision is inconsistent with latest snapshot revision.
+	EtcdRevisionConsistencyError
```
shreyas-s-rao, line 40: If I remember right, we discussed that we use only `RevisionConsistencyError` as the returned status, so there's no need for a new error type here.

pkg/initializer/validator/datavalidator_test.go
```diff
@@ -293,6 +298,57 @@ var _ = Describe("Running Datavalidator", func() {
 			Expect(int(dataDirStatus)).Should(Equal(DataDirectoryStatusUnknown))
 		})
 	})
+
+	Context("with inconsistent revision numbers between etcd and latest snapshot and WALs file have some uncommitted data", func() {
```
shreyas-s-rao, line 302: Please move this test to line 200, before the `with inconsistent revision numbers between etcd and latest snapshot` test

pkg/initializer/validator/datavalidator.go
```diff
@@ -250,13 +267,61 @@ func (d *DataValidator) checkRevisionConsistency(etcdRevision, failBelowRevision
...
+			return FailBelowRevisionConsistencyError, latestSnapshotRevision, nil
 		}
-		return DataDirectoryValid, nil
+		return DataDirectoryValid, latestSnapshotRevision, nil
 	}
 
 	if etcdRevision < latestSnapshotRevision {
-		d.Logger.Infof("current etcd revision (%d) is less than latest snapshot revision (%d): possible data loss", etcdRevision, latestSnapshotRevision)
+		d.Logger.Infof("current etcd revision (%d) is less than latest snapshot revision (%d)", etcdRevision, latestSnapshotRevision)
+		return EtcdRevisionConsistencyError, latestSnapshotRevision, nil
```
shreyas-s-rao, line 277: As already discussed, return `RevisionConsistencyError` here

pkg/initializer/validator/datavalidator.go
```diff
@@ -114,8 +118,19 @@ func (d *DataValidator) sanityCheck(failBelowRevision int64) (DataDirStatus, err
...
 	}
-	d.Logger.Info("Checking for revision consistency...")
-	return d.checkRevisionConsistency(etcdRevision, failBelowRevision)
+
+	d.Logger.Info("Checking for etcd revision consistency...")
+	etcdRevisionStatus, latestSnapshotRevision, err := d.checkEtcdDataRevisionConsistency(etcdRevision, failBelowRevision)
+
+	// if etcd revision is inconsistent with latest snapshot revision then
+	//   check the etcd revision consistency by starting an embedded etcd since the WALs file can have uncommited data which it was unable to flush to Bolt DB.
+	if etcdRevisionStatus == EtcdRevisionConsistencyError {
```
shreyas-s-rao, line 127: As already discussed, change this to `RevisionConsistencyError`

shreyas-s-rao (changes requested):
@ishan16696 Thanks for adding tests. I have added few comments. Please address them. Thanks.

pkg/initializer/validator/types.go
ishan16696: Yes, but I thought a new error will be helpful in future if somebody want to debug .

amshuman-kr: I agree with Shreyas to re-use the existing `RevisionConsistencyError` rather than create a new error type. The distinction between DB file revision being older than backup and the WAL file revision being older than backup is only a _technical_ distinction because of the way we are optimizing the consistency check (checking DB file for a quick check and doing the WAL check only to be doubly sure that the inconsistency is real). It is not a functional or real distinction.

pkg/initializer/validator/types.go
```diff
@@ -34,14 +36,16 @@ const (
 	DataDirectoryCorrupt
 	// DataDirectoryStatusUnknown indicates validator failed to check the data directory status.
 	DataDirectoryStatusUnknown
-	// RevisionConsistencyError indicates current etcd revision is inconsistent with latest snapshot revision.
+	// RevisionConsistencyError indicates current etcd revision is inconsistent with latest snapshot revision even after WALs file flushes the data to Bolt DB,if any.
```
shreyas-s-rao, line 39: `even after WALs file flushes the data to Bolt DB,if any` is an implementation detail, and ideally shouldn't be specified in the error description. You can remove this.

pkg/initializer/validator/datavalidator_test.go
```diff
@@ -136,7 +138,10 @@ var _ = Describe("Running Datavalidator", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 
 			tempFile := path.Join(outputDir, "temp", "db")
-			err = copyFile(dbFile, tempFile)
+
+			fileInfo, err := os.Stat(dbFile)
```
shreyas-s-rao, line 142: We are already running `os.Stat()` on the `dbFile` 3 lines before this. You can simply reuse the output from that to populate `fileInfo`.

shreyas-s-rao (changes requested):
Thanks @ishan16696 for making the changes. Looks like the change in `copyFile` function is breaking one of the other test cases. Please look into this and fix it. You can run unit tests locally before pushing changes, by running `make verify`.

amshuman-kr (approved):
LGTM pending unit test results.

pkg/initializer/validator/datavalidator_test.go
```diff
@@ -132,11 +134,13 @@ var _ = Describe("Running Datavalidator", func() {
...
 		It("should return DataDirStatus as DataDirectoryCorrupt, and nil error", func() {
 			dbFile := path.Join(restoreDataDir, "member", "snap", "db")
-			_, err = os.Stat(dbFile)
+			dbFileInfo, err := os.Stat(dbFile)
 			Expect(err).ShouldNot(HaveOccurred())
 
 			tempFile := path.Join(outputDir, "temp", "db")
-			err = copyFile(dbFile, tempFile)
+
+			Expect(err).ShouldNot(HaveOccurred())
```
shreyas-s-rao, line 142: Please remove this line, as it's redundant.

shreyas-s-rao (approved):
/lgtm

PR #276, opened 2020-10-23
Changed the chart to store local backup in the PV.

amshuman-kr:
Changed the chart to store local backup in the PV. Also, updated the image version.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
@@ -309,4 +309,4 @@ spec:
       - "ReadWriteOnce"
       resources:
         requests:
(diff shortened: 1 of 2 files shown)


shreyas-s-rao:
@amshuman-kr Thanks for the PR. Are we also considering the size of the PV, since backups will have a significant impact on the volume space usage?

amshuman-kr:
> @amshuman-kr Thanks for the PR. Are we also considering the size of the PV, since backups will have a significant impact on the volume space usage?

Thanks for the suggestion. I have made the storage capacity configurable via values. Can you PTAL?

shreyas-s-rao (approved):
/lgtm

PR #282, opened 2020-11-04
Added CLI-flags to enable restoration for delta snapshots with large amount of data

### LINKED ISSUE #263: [BUG] Snapshot restoration failed due to ETCD max request bytes size
**Describe the bug:**

In our production environment, restoration of an ETCD backup potentially fails with the error message `trying to send message larger than max (2541915 vs. 2097152)`. The restoration process will be stuck in an endless loop.

**Expected behavior:**

Restoration succeeds.

I don't know. :(

But I can provide a 4GB backup that is not restorable.

**Logs:**

```
...
etcd-main-0 backup-restore time="2020-09-29T11:09:55Z" level=info msg="Applying delta snapshot Backup-1601328720/Incr-28989619-29005178-1601360349" actor=restorer                                                 
etcd-main-0 backup-restore 2020-09-29 11:10:03.367348 I | wal: segmented wal file /var/etcd/data/new.etcd.part/member/wal/0000000000000012-000000000009d057.wal is created                                         
etcd-main-0 backup-restore {"level":"warn","ts":"2020-09-29T11:10:15.219Z","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"endpoint://client-8df1b049-bb7d-4a2e-8a67
-1384108e55a5/127.0.0.1:44391","attempt":0,"error":"rpc error: code = ResourceExhausted desc = trying to send message larger than max (2541915 vs. 2097152)"}                                                      
etcd-main-0 backup-restore time="2020-09-29T11:10:15Z" level=info msg="Cleanup complete" actor=restorer                                                                                                            
etcd-main-0 backup-restore time="2020-09-29T11:10:15Z" level=error msg="Restoration failed." actor=restorer                                                                                                        
etcd-main-0 backup-restore 2020-09-29 11:10:15.314956 I | etcdserver: skipped leadership transfer for single member cluster
etcd-main-0 backup-restore 2020-09-29 11:10:15.349119 I | etcdserver: skipped leadership transfer for single member cluster
etcd-main-0 backup-restore time="2020-09-29T11:10:15Z" level=error msg="Failed initialization: error while restoring corrupt data: Failed to restore snapshot: failed to apply events to etcd for delta snapshot In
cr-28989619-29005178-1601360349 : rpc error: code = ResourceExhausted desc = trying to send message larger than max (2541915 vs. 2097152)" actor=backup-restore-server
...
```

 - Etcd version/commit ID : 3.3.17 (client in `go.mod` has 3.3.15?)
 - Etcd-backup-restore version/commit ID: v0.9.1
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]: GCS

We were experimenting to resolve this error with a self-built image of the etcd-backup-restore. We tried to increase the embedded ETCD values for `MaxResourceBytes` and `MaxTxnOps` and the client value for `MaxCallSendMsgSize`. However, this was leading to just another error:

``` 
etcd-main-0 backup-restore {"level":"warn","ts":"2020-09-29T13:42:51.482Z","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"endpoint://client-29dfe419-fed5-46ea-96c5
-5fd69783d014/127.0.0.1:33497","attempt":0,"error":"rpc error: code = InvalidArgument desc = etcdserver: too many operations in txn request"}
``` 

Maybe the [restore loop](https://github.com/gardener/etcd-backup-restore/blob/v0.9.1/pkg/snapshot/restorer/restorer.go#L636-L656) which builds transactions needs to be chunked into max 128 transaction objects (default)?

abdasgupta:
Added CLI-flags (`max-call-send-message-size`, `max-request-bytes` and `max-txn-ops`) to enable restoration for delta snapshots with large amount of data (large number of events or events with large data).

Fixes #263 

**Release note**:

```improvement operator
Added CLI-flags (`max-call-send-message-size`, `max-request-bytes` and `max-txn-ops`) to enable restoration for delta snapshots with large amount of data (large number of events or events with large data).
```

### DIFF
--- a/pkg/snapshot/restorer/restorer.go
+++ b/pkg/snapshot/restorer/restorer.go
@@ -73,7 +73,7 @@ func (r *Restorer) Restore(ro RestoreOptions) error {
 		return nil
 	}
 	r.logger.Infof("Starting embedded etcd server...")
-	e, err := StartEmbeddedEtcd(r.logger, ro.Config.RestoreDataDir, ro.Config.EmbeddedEtcdQuotaBytes)
+	e, err := StartEmbeddedEtcd(r.logger, &ro)
 	if err != nil {
 		return err
 	}
@@ -82,7 +82,8 @@ func (r *Restorer) Restore(ro RestoreOptions) error {
 		e.Close()
 	}()
 
-	client, err := clientv3.NewFromURL(e.Clients[0].Addr().String())
+	cfg := clientv3.Config{MaxCallSendMsgSize: ro.Config.MaxCallSendMsgSize, Endpoints: []string{e.Clients[0].Addr().String()}}
+	client, err := clientv3.New(cfg)
 	if err != nil {
 		return err
 	}
@@ -308,9 +309,9 @@ func makeWALAndSnap(logger *zap.Logger, waldir, snapdir string, cl *membership.R
 }
 
 // StartEmbeddedEtcd starts the embedded etcd server.
-func StartEmbeddedEtcd(logger *logrus.Entry, dataDir string, embeddedEtcdQuotaBytes int64) (*embed.Etcd, error) {
+func StartEmbeddedEtcd(logger *logrus.Entry, ro *RestoreOptions) (*embed.Etcd, error) {
 	cfg := embed.NewConfig()
-	cfg.Dir = filepath.Join(dataDir)
+	cfg.Dir = filepath.Join(ro.Config.RestoreDataDir)
 	DefaultListenPeerURLs := "http://localhost:0"
 	DefaultListenClientURLs := "http://localhost:0"
 	DefaultInitialAdvertisePeerURLs := "http://localhost:0"
@@ -324,7 +325,9 @@ func StartEmbeddedEtcd(logger *logrus.Entry, dataDir string, embeddedEtcdQuotaBy
 	cfg.APUrls = []url.URL{*apurl}
 	cfg.ACUrls = []url.URL{*acurl}
 	cfg.InitialCluster = cfg.InitialClusterFromName(cfg.Name)
-	cfg.QuotaBackendBytes = embeddedEtcdQuotaBytes
+	cfg.QuotaBackendBytes = ro.Config.EmbeddedEtcdQuotaBytes
+	cfg.MaxRequestBytes = ro.Config.MaxRequestBytes
+	cfg.MaxTxnOps = ro.Config.MaxTxnOps
 	cfg.Logger = "zap"
 	e, err := embed.StartEtcd(cfg)
 	if err != nil {
--- a/pkg/snapshot/restorer/types.go
+++ b/pkg/snapshot/restorer/types.go
@@ -33,6 +33,9 @@ const (
 	defaultInitialAdvertisePeerURLs = "http://localhost:2380"
 	defaultInitialClusterToken      = "etcd-cluster"
 	defaultMaxFetchers              = 6
+	defaultMaxCallSendMsgSize       = 10 * 1024 * 1024 //10Mib
+	defaultMaxRequestBytes          = 10 * 1024 * 1024 //10Mib
+	defaultMaxTxnOps                = 10 * 1024
 	defaultEmbeddedEtcdQuotaBytes   = 8 * 1024 * 1024 * 1024 //8Gib
 )
 
@@ -64,6 +67,9 @@ type RestorationConfig struct {
 	Name                     string   `json:"name"`
 	SkipHashCheck            bool     `json:"skipHashCheck,omitempty"`
 	MaxFetchers              uint     `json:"maxFetchers,omitempty"`
+	MaxRequestBytes          uint     `json:"MaxRequestBytes,omitempty"`
+	MaxTxnOps                uint     `json:"MaxTxnOps,omitempty"`
+	MaxCallSendMsgSize       int      `json:"maxCallSendMsgSize,omitempty"`
 	EmbeddedEtcdQuotaBytes   int64    `json:"embeddedEtcdQuotaBytes,omitempty"`
 }
 
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -38,6 +38,11 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	defaultMaxRequestBytes = 10 * 1024 * 1024 //10Mib
+	defaultMaxTxnOps       = 10 * 1024
+)
+
 var (
 	// A map of valid files that can be present in the snap folder.
 	validFiles = map[string]bool{
@@ -286,7 +291,15 @@ func (d *DataValidator) checkFullRevisionConsistency(dataDir string, latestSnaps
 	var latestSyncedEtcdRevision int64
 
 	d.Logger.Info("Starting embedded etcd server...")
-	e, err := restorer.StartEmbeddedEtcd(logrus.NewEntry(d.Logger), dataDir, d.Config.EmbeddedEtcdQuotaBytes)
+	ro := &restorer.RestoreOptions{
+		Config: &restorer.RestorationConfig{
+			RestoreDataDir:         dataDir,
+			EmbeddedEtcdQuotaBytes: d.Config.EmbeddedEtcdQuotaBytes,
+			MaxRequestBytes:        defaultMaxRequestBytes,
+			MaxTxnOps:              defaultMaxTxnOps,
+		},
+	}
+	e, err := restorer.StartEmbeddedEtcd(logrus.NewEntry(d.Logger), ro)
 	if err != nil {
 		d.Logger.Infof("unable to start embedded etcd: %v", err)
 		return DataDirectoryCorrupt, err
--- a/pkg/initializer/validator/validator_suite_test.go
+++ b/pkg/initializer/validator/validator_suite_test.go
@@ -117,6 +117,12 @@ func copyFile(sourceFilePath, destinationFilePath string, filePermission os.File
 		return fmt.Errorf("unable to read source file %s: %v", sourceFilePath, err)
 	}
 
+	destinationDirectoryPath := path.Dir(destinationFilePath)
+	err = os.MkdirAll(destinationDirectoryPath, filePermission)
+	if err != nil {
+		return fmt.Errorf("unable to create destination directory %s: %v", destinationDirectoryPath, err)
+	}
+
 	err = ioutil.WriteFile(destinationFilePath, data, filePermission)
 	if err != nil {
 		return fmt.Errorf("unable to create destination file %s: %v", destinationFilePath, err)
--- a/pkg/snapshot/restorer/init.go
+++ b/pkg/snapshot/restorer/init.go
@@ -32,6 +32,9 @@ func NewRestorationConfig() *RestorationConfig {
 		Name:                     defaultName,
 		SkipHashCheck:            false,
 		MaxFetchers:              defaultMaxFetchers,
+		MaxCallSendMsgSize:       defaultMaxCallSendMsgSize,
+		MaxRequestBytes:          defaultMaxRequestBytes,
+		MaxTxnOps:                defaultMaxTxnOps,
 		EmbeddedEtcdQuotaBytes:   int64(defaultEmbeddedEtcdQuotaBytes),
 	}
 }
@@ -45,6 +48,9 @@ func (c *RestorationConfig) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.Name, "name", c.Name, "human-readable name for this member")
 	fs.BoolVar(&c.SkipHashCheck, "skip-hash-check", c.SkipHashCheck, "ignore snapshot integrity hash value (required if copied from data directory)")
 	fs.UintVar(&c.MaxFetchers, "max-fetchers", c.MaxFetchers, "maximum number of threads that will fetch delta snapshots in parallel")
+	fs.IntVar(&c.MaxCallSendMsgSize, "max-call-send-message-size", c.MaxCallSendMsgSize, "maximum size of message that the client sends")
+	fs.UintVar(&c.MaxRequestBytes, "max-request-bytes", c.MaxRequestBytes, "Maximum client request size in bytes the server will accept")
+	fs.UintVar(&c.MaxTxnOps, "max-txn-ops", c.MaxTxnOps, "Maximum number of operations permitted in a transaction")
 	fs.Int64Var(&c.EmbeddedEtcdQuotaBytes, "embedded-etcd-quota-bytes", c.EmbeddedEtcdQuotaBytes, "maximum backend quota for the embedded etcd used for applying delta snapshots")
 }
 
@@ -55,6 +61,9 @@ func (c *RestorationConfig) Validate() error {
 	}
 	if _, err := types.NewURLs(c.InitialAdvertisePeerURLs); err != nil {
 		return fmt.Errorf("failed parsing peers urls for restore cluster: %v", err)
+	}
+	if c.MaxCallSendMsgSize <= 0 {
+		return fmt.Errorf("max call send message should be greater than zero")
 	}
 	if c.MaxFetchers <= 0 {
 		return fmt.Errorf("max fetchers should be greater than zero")
--- a/pkg/snapshot/restorer/restorer_test.go
+++ b/pkg/snapshot/restorer/restorer_test.go
@@ -53,6 +53,9 @@ var _ = Describe("Running Restorer", func() {
 		restoreCluster         string = "default=http://localhost:2380"
 		skipHashCheck          bool   = false
 		maxFetchers            uint   = 6
+		maxCallSendMsgSize            = 2 * 1024 * 1024 //2Mib
+		maxRequestBytes               = 2 * 1024 * 1024 //2Mib
+		maxTxnOps                     = 2 * 1024
 		embeddedEtcdQuotaBytes int64  = 8 * 1024 * 1024 * 1024
 	)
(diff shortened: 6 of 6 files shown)


amshuman-kr (approved):
LGTM

@abdasgupta I have updated the PR title, description and the releases notes. Please check if it makes sense and see anything needs to be changed. You can follow this kind template in the future PRs.

amshuman-kr:
The integration test is failing. @shreyas-s-rao any idea why this is?
```
An error occurred (BucketAlreadyOwnedByYou) when calling the CreateBucket operation: Your previous request to create the named bucket succeeded and you already own it.
```

shreyas-s-rao:
@abdasgupta Thanks for the well-written PR! Can you please check why the unit tests for restoration are failing?


@amshuman-kr The issue was with two pipeline jobs running in parallel and trying to access the same bucket. I have temporarily deleted one of the PR jobs for running tm tests, since the tm tests run on head update anyway. I'll refactor the pipeline definitions today and re-enable on-demand tm tests for PRs later.

abdasgupta:
@shreyas-s-rao I have updated the test cases now

pkg/snapshot/restorer/types.go
```diff
@@ -33,6 +33,9 @@ const (
 	defaultInitialAdvertisePeerURLs = "http://localhost:2380"
 	defaultInitialClusterToken      = "etcd-cluster"
 	defaultMaxFetchers              = 6
+	defaultMaxCallSendMsgSize       = 10 * 1024 * 1024       //10Mib
+	defaultMaxRequestBytes          = 10 * 1024 * 1024       //10Mib
+	defaultMaxTxnOps                = 10 * 1024              //10Kib
```
Gerrit91, line 38: As this value does not represent a size but simply the amount of operations, I think I would avoid the `Kib` comment.

Question to me would also be if it would make sense to chop the operations in `applyEventsToEtcd` into smaller batches in case it exceeds the `maxTxnOps` value. Wouldn't this prevent this kind of failure in the first place? But this would be a separate issue.

Gerrit91 (commented):
Very useful. Thanks for this PR. :+1: 

Hope it's fine that I added a review comment.

pkg/snapshot/restorer/types.go
abdasgupta: Hi @Gerrit91 Yes, I would remove the comment shortly.. 

Yeah, I am currently working on this.. So far I have seen that `check perf` or any large operations are transacting in a particular order which is okay with the Etcd cluster.. But after we are taking delta snapshot, we are losing that orders probably. I am trying to look for a way to preserve the same order and if not possible, then I will do something as you suggested.. I would like to raise that as another PR.

abdasgupta:
Hi @Gerrit91 As you mentioned, the request time out during restoration of very large delta snapshot was due to the fact that there were very high number of operations getting executed in one transaction. So I added another commit which will basically divide the operations into smaller chunks as per MaxTxnOps. With this commit restoration after `check perf` of all sizes are working fine. Please review it. cc @amshuman-kr @shreyas-s-rao

abdasgupta:
I also have something to discuss on this regard. After this PR, the events for restoration operations are executed in chunks serially. Should we think about spawning multiple clients for each chunk? If we do it, the chunks of operations will be performed in parallel and restoration operation will be quicker as a whole. Is there any downside of it? What do you think? cc @Gerrit91 @amshuman-kr @shreyas-s-rao @majst01

amshuman-kr:
> I also have something to discuss on this regard. After this PR, the events for restoration operations are executed in chunks serially. Should we think about spawning multiple clients for each chunk? If we do it, the chunks of operations will be performed in parallel and restoration operation will be quicker as a whole. Is there any downside of it? What do you think? cc @Gerrit91 @amshuman-kr @shreyas-s-rao @majst01

Unfortunately, this is not possible. Delta snapshot changes have to be applied serially. Otherwise will potentially lose consistency.

pkg/snapshot/restorer/restorer.go
```diff
@@ -659,7 +662,14 @@ func applyEventsToEtcd(client *clientv3.Client, events []event) error {
 			return fmt.Errorf("Unexpected event type")
 		}
 	}
-	_, err := client.Txn(ctx).Then(ops...).Commit()
+
+	// Divide the total operations in smaller chunks as mer MaxTxnOps
+	var err error
+	for len(ops)/int(ro.Config.MaxTxnOps) > 0 {
```
amshuman-kr, line 668:
```suggestion
	for len(ops) > ro.Config.MaxTxnOps {
```

amshuman-kr (changes requested):
@abdasgupta Thanks for the enhancing the PR already!

Couple of small changes.

majst01:
From what i understood from reading the code, downloading of the chunks to perform a restore is already done in parallel, but the actual apply of these must be done in serial, in the right order, as @amshuman-kr already pointed out.

pkg/snapshot/restorer/restorer.go
Gerrit91, line 668: I find this quite hard to read, but this could be a personal thing, I don't know. Probably, to enhance readability, I would first construct batches of operations, and then iterate again over the batches and apply them. But if this you tested this and it works, go for it. :)

Gerrit91 (approved):

pkg/snapshot/restorer/restorer.go
```diff
@@ -659,7 +662,20 @@ func applyEventsToEtcd(client *clientv3.Client, events []event) error {
 			return fmt.Errorf("Unexpected event type")
 		}
 	}
-	_, err := client.Txn(ctx).Then(ops...).Commit()
+
+	// Divide the total operations in smaller chunks as per MaxTxnOps
+	fmt.Println("Length of ops is: ", len(ops))
```
amshuman-kr, line 667: Can you please use log instead of printing here?

pkg/snapshot/restorer/restorer.go
```diff
@@ -659,7 +662,20 @@ func applyEventsToEtcd(client *clientv3.Client, events []event) error {
...
+	var err error
+	/*for _, op := range ops {
+		_, err = client.Do(context.Background(), op)
+	}*/
+	for len(ops) > int(ro.Config.MaxTxnOps) {
+		chunk := ops[ro.Config.MaxTxnOps:]
+		_, err = client.Txn(ctx).Then(chunk...).Commit()
+		ops = ops[:ro.Config.MaxTxnOps]
+		fmt.Println("Length of ops is: ", len(ops))
```
amshuman-kr, line 676: Can you please use log instead of printing here?

pkg/snapshot/restorer/restorer.go
```diff
@@ -659,7 +662,20 @@ func applyEventsToEtcd(client *clientv3.Client, events []event) error {
...
-	_, err := client.Txn(ctx).Then(ops...).Commit()
+
+	// Divide the total operations in smaller chunks as per MaxTxnOps
+	fmt.Println("Length of ops is: ", len(ops))
+	var err error
+	/*for _, op := range ops {
+		_, err = client.Do(context.Background(), op)
+	}*/
+	for len(ops) > int(ro.Config.MaxTxnOps) {
+		chunk := ops[ro.Config.MaxTxnOps:]
```
amshuman-kr, line 673: Shouldn't the chunks be applied left (older) to right (newer)?
```suggestion
		chunk := ops[:ro.Config.MaxTxnOps]
```

pkg/snapshot/restorer/restorer.go
```diff
@@ -659,7 +662,20 @@ func applyEventsToEtcd(client *clientv3.Client, events []event) error {
...
+		_, err = client.Txn(ctx).Then(chunk...).Commit()
+		ops = ops[:ro.Config.MaxTxnOps]
```
amshuman-kr, line 675: Shouldn't the chunks be applied left (older) to right (newer)?
```suggestion
		ops = ops[ro.Config.MaxTxnOps:]
```

amshuman-kr (changes requested):
Thanks @abdasgupta for the changes. I see that the chunking is done right to left. Shouldn't the chunking be left to right?

abdasgupta:
Hi 👋, I reverted back this PR to the old commit which only enhance ETCDBR to accept CLI flags. What I noticed that if we divide the operations into smaller chunks of operations, the consistency is lost. The restore operation throws error with revision mismatch. Possible reason is might be that the transactions of different chunks are running as threads in Etcd server. So the operations from different chunks might not be in sync. I didn't found any lightweight method to actually make the transactions from different chunks execute serially in ETCD server. I also tried to use PUT and DELETE operations instead of transaction. But I found revision mismatch error with that as well. Moreover, executing PUT and DELETE operations one by one is very slow. Any thoughts on this?

abdasgupta:
Another thought is, if we are to support large number of transactions during delta snapshot, we should start a full snapshot if we find the number of operations are more than , say 256 for delta snapshot.

amshuman-kr:
Quote from https://etcd.io/docs/v3.4.0/learning/api/#transaction.
> A transaction can atomically process multiple requests in a single request. For modifications to the key-value store, this means the store’s revision is incremented only once for the transaction and all events generated by the transaction will have the same revision.

But I doubt all the events in a single delta snapshot ever belong to a single transaction while the changes are originally made on etcd.

After that there is this 👇 in the same documentation.
>However, modifications to the same key multiple times within a single transaction are forbidden.

I am not sure we are checking this either which is again another problem.

> The restore operation throws error with revision mismatch.

@abdasgupta I suppose you meant that a revision check after restoration fails with this error. But I don't know if we check after restoration if the restored revision matches the backup latest revision? From the above documentation, it looks like they may not match even based on the current implementation (before this PR)!
I double checked with [v3.3.12 documentation](https://etcd.io/docs/v3.3.13/learning/api/#transaction) to make sure the behaviour has not changed in v.3.4.0.

Anyway, if I understood right, we have the following options. Anything else will be technically incorrect.
1. Continue to use transactions during delta snapshot restoration restoration but keep them faithful (one-to-one) with original source transactions. Maybe there is a way to group events using their `ModRevision`. This can be complicated because some events of the same source transaction might possibly be split across delta snapshots. But with this approach we can retain the one-to-one mapping of original and restored revisions.
1. Give up on one-to-one mapping of original and restored revisions (i.e. no checks on revisions after restoration) but still continue to use transaction during delta snapshot restoration to _potentially_ speed up delta snapshot restoration.
1. Give up on one-to-one mapping of original and restored revisions (i.e. no checks on revisions after restoration), give up on optimizing speed of delta snapshot restoration and just apply each event in its own transaction.

Or I have misunderstood completely, which is certainly possible 😉 

cc @shreyas-s-rao

shreyas-s-rao:
@amshuman-kr I don't particularly feel comfortable giving up on revision checks during restoration, because right now that's our only safety net against restoration inconsistencies, if at all the snapshot was not uploaded correctly, or something went wrong in the bucket or even during the download of the object.

> However, modifications to the same key multiple times within a single transaction are forbidden

I too remember seeing this in the documentation, but there never seemed to an issue regarding this when we applied the transactions during restoration. Surprising indeed.

Since delta snapshots record events (which are either PUT or DELETE), I think we should pursue restoring delta snapshots in the same way, serially applying PUT and DELETE events instead of batching them into transactions. @abdasgupta @amshuman-kr WDYT?

abdasgupta:
I have tried to do only PUT and DELETE operations, instead of using transactions. First problem is, each operation would commit as I understood. So, for large Delta events, it should take larger time. Second problem, that also gave error of revision mismatch. 

The changes I Made:
```
func applyEventsToEtcd(client *clientv3.Client, events []event) error {
	var (
		lastRev int64
		ops     = []clientv3.Op{}
		ctx     = context.TODO()
	)

	for _, e := range events {
		ev := e.EtcdEvent
		nextRev := ev.Kv.ModRevision
		if lastRev != 0 && nextRev > lastRev {
			if _, err := client.Txn(ctx).Then(ops...).Commit(); err != nil {
				return err
			}
			ops = []clientv3.Op{}
		}
		lastRev = nextRev
		switch ev.Type {
		case mvccpb.PUT:
			//ops = append(ops, clientv3.OpPut(string(ev.Kv.Key), string(ev.Kv.Value))) //, clientv3.WithLease(clientv3.LeaseID(ev.Kv.Lease))))
			_, err := client.Put(ctx, string(ev.Kv.Key), string(ev.Kv.Value))
			if err != nil {
				return err
			}

		case mvccpb.DELETE:
			//ops = append(ops, clientv3.OpDelete(string(ev.Kv.Key)))
			_, err := client.Delete(ctx, string(ev.Kv.Key))
			if err != nil {
				return err
			}
		default:
			return fmt.Errorf("Unexpected event type")
		}
	}
	//_, err := client.Txn(ctx).Then(ops...).Commit()
	return nil
}
```

The error:
```
INFO[0148] Cleanup complete                              actor=restorer
ERRO[0148] Restoration failed.                           actor=restorer
{"level":"info","ts":"2020-11-09T22:49:56.595+0530","caller":"etcdserver/server.go:1464","msg":"skipped leadership transfer for single voting member cluster","local-member-id":"8e9e05c52164694d","current-leader-member-id":"8e9e05c52164694d"}
{"level":"info","ts":"2020-11-09T22:49:56.614+0530","caller":"embed/etcd.go:360","msg":"closing etcd server","name":"default","data-dir":"default.etcd","advertise-peer-urls":["http://localhost:0"],"advertise-client-urls":["http://localhost:0"]}
{"level":"info","ts":"2020-11-09T22:49:56.614+0530","caller":"etcdserver/server.go:1464","msg":"skipped leadership transfer for single voting member cluster","local-member-id":"8e9e05c52164694d","current-leader-member-id":"8e9e05c52164694d"}
{"level":"info","ts":"2020-11-09T22:49:56.615+0530","caller":"embed/etcd.go:555","msg":"stopping serving peer traffic","address":"127.0.0.1:51885"}
{"level":"info","ts":"2020-11-09T22:49:56.615+0530","caller":"embed/etcd.go:562","msg":"stopped serving peer traffic","address":"127.0.0.1:51885"}
{"level":"info","ts":"2020-11-09T22:49:56.615+0530","caller":"embed/etcd.go:364","msg":"closed etcd server","name":"default","data-dir":"default.etcd","advertise-peer-urls":["http://localhost:0"],"advertise-client-urls":["http://localhost:0"]}
FATA[0148] Failed to restore snapshot: snapshot revision verification failed for delta snapshot Incr-00008597-00009002-1604909996 : mismatched event revision while applying delta snapshot, expected 9002 but applied 18001
```

Hope this is what you were suggesting? @shreyas-s-rao 

As a reason for this error, I think that the delta snapshot operations are queued to the ETCD server in a order which is not matching with the order the changes were made actually .

shreyas-s-rao:
@abdasgupta Yes, this is what I was thinking. If I'm right, most of the delta snapshots events are getting applied correctly, but that applying that last event is throwing the error right?

amshuman-kr:
> I don't particularly feel comfortable giving up on revision checks during restoration, because right now that's our only safety net against restoration inconsistencies, if at all the snapshot was not uploaded correctly, or something went wrong in the bucket or even during the download of the object.

 I agree.

> Since delta snapshots record events (which are either PUT or DELETE), I think we should pursue restoring delta snapshots in the same way, serially applying PUT and DELETE events instead of batching them into transactions.

I am afraid this is not an option. As @abdasgupta showed [above](https://github.com/gardener/etcd-backup-restore/pull/282#issuecomment-724156277), if the original events were done using transaction then not using transaction while restoring will still lead to mismatch in revisions after restoration.

If we do not want to give up on revision check after restoration (and I agree that we should not), then the correct thing is to group the events into transactions only if they share the same `ModRevision`. It seems to me that generally using transactions during restoration _optimization_ like we have done so far is incorrect. Generally _not using_ transactions during restoration seems also equally incorrect.

Furthermore, the implementation idea for [compaction](https://github.com/gardener/etcd-druid/issues/88) of compacting the delta snapshots asynchronously by periodically restoring and replacing incremental snapshots with full snapshots would work only if revisions are consistent across restorations. So, we don't really have a choice here. As far as I can see, we have to honour original transaction during restoration. The only tricky part for me is how to ensure events from the same original transaction do not get split into multiple delta snapshot files. If that can be ensured, then the rest is manageable, I think.

amshuman-kr:
> The only tricky part for me is how to ensure events from the same original transaction do not get split into multiple delta snapshot files.

I think this can be achieved if etcd watch returns all the events from the same transaction in one watch response. I am reasonably confident that it will. We need to check if this is guaranteed to be true.

amshuman-kr:
TL;DR 😉 

Transactions are an ETCD functionality designed for us by clients of ETCD however they see fit. It clearly has semantics associated with it and not just an optimization hack. We should not use it for any other purpose. Equally, we should not avoid using it for any other purpose either.

amshuman-kr:
BTW I am ok merging this PR as is with the additional CLI flags without the chunking changes during restoration because the issue with transactions is clearly a pre-existing condition. We can create a separate issue to track it (and address it in a separate PR).

abdasgupta:
There is a conflict appearing in this PR due to https://github.com/gardener/etcd-backup-restore/pull/275

shreyas-s-rao:
@abdasgupta Yes, can you please rebase the PR?

abdasgupta:
@amshuman-kr @shreyas-s-rao Took the liberty of passing `RestoreOption` as parameter to startEmbeddedEtcd for both the validator and restorer. Please review if I missed something in validator. `make test` is passing on my system and I checked the restoration operation is also passing.

shreyas-s-rao (approved):
@abdasgupta Thanks for the PR. LGTM.

amshuman-kr (approved):
LGTM

Issue #285, opened 2020-11-23
[Feature] Use Unix Domain Socket (UDS) for sidecar from the etcd RPCs


mvladev:
**Feature (What you would like to be added):**

Instead of running a TLS server for the `initialization/status` and `initialization/start`, use unix domain socket (UDS) to connect from etcd -> bacup sidecar. The socked can be created in a `emptyVolume` that is shared between containers.

I would very **recommend** to replace the [bootstrap bash logic](https://github.com/gardener/etcd-backup-restore/blob/a48eac88fc15b298f9709c4767bdc839af7d23e2/chart/etcd-backup-restore/templates/etcd-bootstrap-configmap.yaml#L13-L87) and replace with golang code.

**Motivation (Why is this needed?):**

This removes the need for TLS and unneeded overhead.

**Approach/Hint to the implement solution (optional):**

1. Create a new uds listener and add status and start endpoints to it.
1. Keep the TLS old code and HTTP handler for backwards compatibility.

amshuman-kr:
> Instead of running a TLS server for the initialization/status and initialization/start, use unix domain socket (UDS) to connect from etcd -> bacup sidecar. The socked can be created in a emptyVolume that is shared between containers.

Considering that there are other _API_ which anyway are required (e.g. request on-demand full/delta snapshot) to be REST over TLS, I am not sure how much moving only this part to UDS would help.

> I would very recommend to replace the bootstrap bash logic and replace with golang code.

Agreed. Earlier when we were using the upstream ETCD image, we were handicapped but now with [gardener/etcd-custom-image](https://github.com/gardener/etcd-custom-image) we can pick this.

PR #286, opened 2020-11-23
Add UDS listener for start and status

mvladev:

Add UDS listener for start and status. This would allow for the etcd to talk to the backup sidecar without TCP via Unix Domain Socket.

The old behavior still exists and it's not changed.

Partially #285 

**Release note**:

```noteworthy operator
The server now listens on Unix Domain Socket (UDS) and serves `initialization/status` and `initialization/start` resources. This allows for the etcd bootstrap to use that socket directly without TLS/TCP. 
The socket path can be configured with `--uds-path` flag. It defaults to `/var/run/etcdbrctl/etcdbrctl.socket`.
```

### DIFF
--- a/chart/etcd-backup-restore/templates/etcd-bootstrap-configmap.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-bootstrap-configmap.yaml
@@ -10,18 +10,8 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
 data:
   bootstrap.sh: |-
-    #!/bin/sh
+    #!/bin/bash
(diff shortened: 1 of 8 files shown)


mvladev:
ping

timuthy:
@gardener/etcd-druid-maintainers  is this PR still relevant? Having a TLS endpoint will still be necessary in a multi-node etcd setup as certain requests are being relayed to the leader.
So, we cannot entirely rely on UDS for communication and only using it for some scenarios while we still use TLS for other might not be too beneficial, right?

timuthy:
/closed
See https://github.com/gardener/etcd-backup-restore/pull/286#issuecomment-1031496509

PR #293, opened 2020-12-07
Compress the snapshots(full as well as Incremental) before uploading to object store and decompress the snapshot before restoration.

### LINKED ISSUE #255: [Feature] Compress/decompress snapshots before/after uploading/downloading to/from object store
At present, (full and incremental) snapshots are stored in the object store uncompressed. 

Can we optionally support storing compressed (full and incremental) snapshots in object store? I.e. Compress the snapshots before uploading to object store and decompress backups after downloading from object store for restoration.

This could be controlled by a configuration/cli flag, e.g. `compressSnapshots: true` or `--compress-snapshots=true`.

While storing uncompressed snapshots in the object store may lead to some simplicity in implementation and optimization in time performance, it also leads to higher network and storage usage.

- Almost none of the object store providers seem to support on-the-go HTTP `Transfer-Encoding`.
- Almost all the object store providers seem to support `Content-Encoding: gzip` header to mark the content in the object store as compressed while supporting the `Content-Type` header for the actual uncompressed content.
  - However, there is no uniformity in supporting (see links below) automatic compression while uploading or automatic decompression while downlloading/serving.

So, it makes sense for `etcd-backup-restore` to
- Compress the snapshots explicitly before uploading.
  - In the case of a multipart object, the whole object is to be compressed before splitting into chunks.
- Mark the content in the `Content-Encoding: gzip` in the object store to indicate that the content is compressed while maintaining the original `Content-Type`.
- Decompress the snapshots after downloading for restoration.
- Support backward compatibility for restoring older non-compressed snapshots for the same etcd.
- Streaming or on-the-go compression/decompression would be preferable so that the footprint on the disk doesn't increase.

1. https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html
1. https://cloud.google.com/storage/docs/transcoding
1. https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob
1. https://docs.microsoft.com/en-us/azure/cdn/cdn-improve-performance
1. https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#CompressedS3

ishan16696:
This PR added the compression/decompression feature. So, it compress/decompress the snapshots(full as well as Incremental) before/after uploading/downloading to/from object store. 
It also supports the backward compatibility for restoring from non-compressed snapshots as well as when different compressionPolicy were used to take the snapshot.

Fixes #255 

**Release note**:
 
```noteworthy user
Add support for snapshot compression/decompression. Compression and compression policy can be configured through flags:  `--compress-snapshots` and `--compression-policy` respectively. Supported compression policies currently are `gzip` (default), `lzw` and `zlib`. Snapshot compression is disabled by default.
```

### DIFF
--- a/example/00-backup-restore-server-config.yaml
+++ b/example/00-backup-restore-server-config.yaml
@@ -43,3 +43,7 @@ restorationConfig:
   embeddedEtcdQuotaBytes: 8589934592
 
 defragmentationSchedule: "0 0 */3 * *"
+
+compressionConfig:
+   enabled: true
+   policy: "gzip"
--- a/pkg/compressor/compressor.go
+++ b/pkg/compressor/compressor.go
@@ -1,0 +1,152 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compressor
+
+import (
+	"compress/gzip"
+	"compress/lzw"
+	"compress/zlib"
+	"fmt"
+	"io"
+
+	"github.com/sirupsen/logrus"
+)
+
+// CompressSnapshot takes uncompressed data as input and compress the data according to Compression Policy
+// and write the compressed data into one end of pipe.
+func CompressSnapshot(data io.ReadCloser, compressionPolicy string) (io.ReadCloser, error) {
+	pReader, pWriter := io.Pipe()
+
+	var gWriter io.WriteCloser
+	logger := logrus.New().WithField("actor", "compressor")
+	logger.Infof("start compressing the snapshot using %v Compression Policy", compressionPolicy)
+
+	switch compressionPolicy {
+	case GzipCompressionPolicy:
+		gWriter = gzip.NewWriter(pWriter)
+
+	case LzwCompressionPolicy:
+		gWriter = lzw.NewWriter(pWriter, lzw.LSB, LzwLiteralWidth)
+
+	case ZlibCompressionPolicy:
+		gWriter = zlib.NewWriter(pWriter)
+
+	// It is actually unreachable but just to be on safe side:
+	// for unsupported CompressionPolicy return the error
+	default:
+		return nil, fmt.Errorf("Unsupported Compression Policy")
+
+	}
+
+	go func() {
+		var err error
+		defer pWriter.CloseWithError(err)
+		defer gWriter.Close()
+		defer data.Close()
+		_, err = io.Copy(gWriter, data)
+		if err != nil {
+			logger.Infof("compression failed: %v", err)
+			return
+		}
+	}()
+
+	return pReader, nil
+}
+
+// DecompressSnapshot take compressed data and compressionPolicy as input and
+// it decompresses the data according to compression Policy and return uncompressed data.
+func DecompressSnapshot(data io.ReadCloser, compressionPolicy string) (io.ReadCloser, error) {
+	var deCompressedData io.ReadCloser
+	var err error
+
+	logger := logrus.New().WithField("actor", "de-compressor")
+	logger.Infof("start decompressing the snapshot with %v compressionPolicy", compressionPolicy)
+
+	switch compressionPolicy {
+	case ZlibCompressionPolicy:
+		deCompressedData, err = zlib.NewReader(data)
+		if err != nil {
+			logger.Infof("Unable to decompress: %v", err)
+			return data, err
+		}
+
+	case GzipCompressionPolicy:
+		deCompressedData, err = gzip.NewReader(data)
+		if err != nil {
+			logger.Infof("Unable to decompress: %v", err)
+			return data, err
+		}
+
+	case LzwCompressionPolicy:
+		deCompressedData = lzw.NewReader(data, lzw.LSB, LzwLiteralWidth)
+
+	// It is actually unreachable but just to be on safe side:
+	// for unsupported CompressionPolicy return the same data with error
+	default:
+		return data, fmt.Errorf("Unsupported Compression Policy")
+	}
+
+	return deCompressedData, nil
+}
+
+// GetCompressionSuffix returns the suffix for snapshot w.r.t Compression Policy
+// if compression is not enabled, it will simply return UnCompressSnapshotExtension(empty string).
+func GetCompressionSuffix(compressionEnabled bool, compressionPolicy string) (string, error) {
+
+	if !compressionEnabled {
+		return UnCompressSnapshotExtension, nil
+	}
+
+	switch compressionPolicy {
+	case ZlibCompressionPolicy:
+		return ZlibCompressionExtension, nil
+
+	case LzwCompressionPolicy:
+		return LzwCompressionExtension, nil
+
+	case GzipCompressionPolicy:
+		return GzipCompressionExtension, nil
+
+	// unreachable but just to be on safe side:
+	// for unsupported CompressionPolicy return the error
+	default:
+		return "", fmt.Errorf("Unsupported Compression Policy")
+
+	}
+}
+
+// IsSnapshotCompressed is helpful in determining whether the snapshot is compressed or not.
+// it will return boolean, compressionPolicy corresponding to compressionSuffix and error.
+func IsSnapshotCompressed(compressionSuffix string) (bool, string, error) {
+
+	switch compressionSuffix {
+	case ZlibCompressionExtension:
+		return true, ZlibCompressionPolicy, nil
+
+	case GzipCompressionExtension:
+		return true, GzipCompressionPolicy, nil
+
+	case LzwCompressionExtension:
+		return true, LzwCompressionPolicy, nil
+
+	case UnCompressSnapshotExtension:
+		return false, "", nil
+
+	// actually unreachable but just to be on safe side:
+	// for unsupported CompressionPolicy return the error
+	default:
+		return false, "", fmt.Errorf("Unsupported Compression Policy")
+	}
+}
--- a/pkg/compressor/init.go
+++ b/pkg/compressor/init.go
@@ -1,0 +1,53 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compressor
+
+import (
+	"fmt"
+
+	flag "github.com/spf13/pflag"
+)
+
+// NewCompressorConfig returns the compression config.
+func NewCompressorConfig() *CompressionConfig {
+	return &CompressionConfig{
+		Enabled:           DefaultCompression,
+		CompressionPolicy: DefaultCompressionPolicy,
+	}
+}
+
+// AddFlags adds the flags to flagset.
+func (c *CompressionConfig) AddFlags(fs *flag.FlagSet) {
+
+	fs.BoolVar(&c.Enabled, "compress-snapshots", c.Enabled, "whether to compress the snapshots or not")
+	fs.StringVar(&c.CompressionPolicy, "compression-policy", c.CompressionPolicy, "Policy for compressing the snapshots")
+}
+
+// Validate validates the compression Config.
+func (c *CompressionConfig) Validate() error {
+
+	// if compression is not enabled then to check CompressionPolicy becomes irrelevant
+	if c.Enabled == false {
+		return nil
+	}
+
+	for _, policy := range []string{GzipCompressionPolicy, ZlibCompressionPolicy, LzwCompressionPolicy} {
+		if c.CompressionPolicy == policy {
+			return nil
+		}
+	}
+	return fmt.Errorf("%s: Compression Policy is not supported", c.CompressionPolicy)
+
+}
--- a/pkg/compressor/types.go
+++ b/pkg/compressor/types.go
@@ -1,0 +1,49 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compressor
+
+const (
+
+	// GzipCompressionPolicy is constant for gzip compression algorithm.
+	GzipCompressionPolicy = "gzip"
+	// LzwCompressionPolicy is constant for lzw compression algorithm.
+	LzwCompressionPolicy = "lzw"
+	// ZlibCompressionPolicy is constant for zlib compression algorithm.
+	ZlibCompressionPolicy = "zlib"
+
+	// DefaultCompression is constant used for whether to compress the snapshots or not.
+	DefaultCompression = false
+	// DefaultCompressionPolicy is constant for default compression algorithm(only if compression is enabled).
+	DefaultCompressionPolicy = "gzip"
+
+	// UnCompressSnapshotExtension is used for snapshot suffix when compression is not enabled.
+	UnCompressSnapshotExtension = ""
+	// GzipCompressionExtension is used for snapshot suffix when compressionPolicy is gzip.
+	GzipCompressionExtension = ".gz"
+	// LzwCompressionExtension is used for snapshot suffix when compressionPolicy is lzw.
+	LzwCompressionExtension = ".Z"
+	// ZlibCompressionExtension is used for snapshot suffix when compressionPolicy is zlib.
+	ZlibCompressionExtension = ".zlib"
+	// Reference: https://en.wikipedia.org/wiki/List_of_archive_formats
+
+	// LzwLiteralWidth is constant used as literal Width in lzw compressionPolicy.
+	LzwLiteralWidth = 8 //[2,8]
+)
+
+// CompressionConfig holds the compression configuration.
+type CompressionConfig struct {
+	Enabled           bool   `json:"enabled"`
+	CompressionPolicy string `json:"policy,omitempty"`
+}
--- a/pkg/server/types.go
+++ b/pkg/server/types.go
@@ -15,6 +15,7 @@
 package server
 
 import (
+	"github.com/gardener/etcd-backup-restore/pkg/compressor"
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
 	"github.com/gardener/etcd-backup-restore/pkg/snapshot/restorer"
 	"github.com/gardener/etcd-backup-restore/pkg/snapshot/snapshotter"
@@ -33,6 +34,7 @@ type BackupRestoreComponentConfig struct {
 	SnapshotterConfig       *snapshotter.Config            `json:"snapshotterConfig,omitempty"`
 	SnapstoreConfig         *snapstore.Config              `json:"snapstoreConfig,omitempty"`
 	RestorationConfig       *restorer.RestorationConfig    `json:"restorationConfig,omitempty"`
+	CompressionConfig       *compressor.CompressionConfig  `json:"compressionConfig,omitempty"`
 	DefragmentationSchedule string                         `json:"defragmentationSchedule"`
 }
 
--- a/pkg/snapshot/restorer/restorer.go
+++ b/pkg/snapshot/restorer/restorer.go
@@ -31,6 +31,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gardener/etcd-backup-restore/pkg/compressor"
 	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/clientv3"
@@ -123,17 +124,30 @@ func (r *Restorer) restoreFromBaseSnapshot(ro RestoreOptions) error {
 
 	walDir := filepath.Join(memberDir, "wal")
 	snapdir := filepath.Join(memberDir, "snap")
-	if err = makeDB(r.zapLogger, snapdir, ro.BaseSnapshot, len(cl.Members()), r.store, false); err != nil {
+	if err = r.makeDB(snapdir, ro.BaseSnapshot, len(cl.Members()), false); err != nil {
 		return err
 	}
 	return makeWALAndSnap(r.zapLogger, walDir, snapdir, cl, ro.Config.Name)
 }
 
 // makeDB copies the database snapshot to the snapshot directory.
-func makeDB(logger *zap.Logger, snapdir string, snap snapstore.Snapshot, commit int, ss snapstore.SnapStore, skipHashCheck bool) error {
-	rc, err := ss.Fetch(snap)
+func (r *Restorer) makeDB(snapdir string, snap snapstore.Snapshot, commit int, skipHashCheck bool) error {
+	rc, err := r.store.Fetch(snap)
 	if err != nil {
 		return err
+	}
+
+	startTime := time.Now()
+	isCompressed, compressionPolicy, err := compressor.IsSnapshotCompressed(snap.CompressionSuffix)
+	if err != nil {
+		return err
+	}
+	if isCompressed {
+		// decompress the snapshot
+		rc, err = compressor.DecompressSnapshot(rc, compressionPolicy)
+		if err != nil {
+			return fmt.Errorf("unable to decompress the snapshot: %v", err)
+		}
 	}
 	defer rc.Close()
 
@@ -150,6 +164,13 @@ func makeDB(logger *zap.Logger, snapdir string, snap snapstore.Snapshot, commit 
 		return err
 	}
 	db.Sync()
+	totalTime := time.Now().Sub(startTime).Seconds()
+
+	if isCompressed {
+		r.logger.Infof("successfully fetched data of base snapshot in %v seconds [CompressionPolicy:%v]", totalTime, compressionPolicy)
+	} else {
+		r.logger.Infof("successfully fetched data of base snapshot in %v seconds", totalTime)
+	}
 
 	off, err := db.Seek(0, io.SeekEnd)
 	if err != nil {
@@ -199,9 +220,9 @@ func makeDB(logger *zap.Logger, snapdir string, snap snapstore.Snapshot, commit 
 	// having a new raft instance
 	be := backend.NewDefaultBackend(dbPath)
 	// a lessor that never times out leases
-	lessor := lease.NewLessor(logger, be, lease.LessorConfig{MinLeaseTTL: math.MaxInt64})
-	s := mvcc.NewStore(logger, be, lessor, (*initIndex)(&commit), mvcc.StoreConfig{})
-	trace := traceutil.New("write", logger)
+	lessor := lease.NewLessor(r.zapLogger, be, lease.LessorConfig{MinLeaseTTL: math.MaxInt64})
+	s := mvcc.NewStore(r.zapLogger, be, lessor, (*initIndex)(&commit), mvcc.StoreConfig{})
+	trace := traceutil.New("write", r.zapLogger)
 	txn := s.Write(trace)
 	btx := be.BatchTx()
 	del := func(k, v []byte) error {
@@ -433,7 +454,7 @@ func (r *Restorer) fetchSnaps(fetcherIndex int, fetcherInfoCh <-chan fetcherInfo
 		default:
 			r.logger.Infof("Fetcher #%d fetching delta snapshot %s", fetcherIndex+1, path.Join(fetcherInfo.Snapshot.SnapDir, fetcherInfo.Snapshot.SnapName))
 
-			eventsData, err := getEventsDataFromDeltaSnapshot(r.store, fetcherInfo.Snapshot)
+			eventsData, err := r.getEventsDataFromDeltaSnapshot(fetcherInfo.Snapshot)
 			if err != nil {
 				errCh <- fmt.Errorf("failed to read events data from delta snapshot %s : %v", fetcherInfo.Snapshot.SnapName, err)
 				applierInfoCh <- applierInfo{SnapIndex: -1} // cannot use close(ch) as concurrent fetchSnaps routines might try to send on channel, causing a panic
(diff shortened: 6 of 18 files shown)


pkg/compressor/init.go
```diff
@@ -0,0 +1,45 @@
...
+	fs.StringVar(&c.CompressionPolicy, "compression-policy", c.CompressionPolicy, "Policy for compress the snapshots")
+}
+
+// Validate validates the compression Config.
+func (c *CompressionConfig) Validate() error {
+
+	if c.CompressionPolicy != GzipCompressionPolicy && c.CompressionPolicy != ZlibCompressionPolicy && c.CompressionPolicy != LzwCompressionPolicy && c.CompressionPolicy != FlateCompressionPolicy {
+		return fmt.Errorf("%s: Compression Policy is not supported", c.CompressionPolicy)
+	}
+	return nil
```
amshuman-kr, line 44:
```suggestion
	for policy := range []string{ GzipCompressionPolicy, ZlibCompressionPolicy, LzwCompressionPolicy, FlateCompressionPolicy} {
		if c.CompressionPolicy == policy {
			return nil
		}
	}
	return fmt.Errorf("%s: Compression Policy is not supported", c.CompressionPolicy)
```

example/00-backup-restore-server-config.yaml
```diff
@@ -42,4 +42,8 @@ restorationConfig:
   maxFetchers: 6
   embeddedEtcdQuotaBytes: 8589934592
 
-defragmentationSchedule: "0 0 */3 * *"
\ No newline at end of file
+defragmentationSchedule: "0 0 */3 * *"
+
+compressorConfig:
+   compressionEnabled: true
+   compressionPolicy: "flate"
```
amshuman-kr, line 49:
```suggestion
compressionConfig:
   enabled: true
   policy: "flate"
```

pkg/compressor/types.go
```diff
@@ -0,0 +1,47 @@
...
+	// LzwCompressionExtension is used for snapshot suffix when compressionPolicy is lzw.
+	LzwCompressionExtension = ".lzw"
+	// ZlibCompressionExtension is used for snapshot suffix when compressionPolicy is zlib.
+	ZlibCompressionExtension = ".zlib"
+)
+
+// CompressionConfig holds the compression configuration.
+type CompressionConfig struct {
+	Enabled           bool   `json:"compressionEnabled"`
+	CompressionPolicy string `json:"compressionPolicy,omitempty"`
```
amshuman-kr, line 46:
```suggestion
	Enabled           bool   `json:"enabled"`
	CompressionPolicy string `json:"policy,omitempty"`
```

pkg/server/types.go
```diff
@@ -33,6 +34,7 @@ type BackupRestoreComponentConfig struct {
 	SnapshotterConfig       *snapshotter.Config            `json:"snapshotterConfig,omitempty"`
 	SnapstoreConfig         *snapstore.Config              `json:"snapstoreConfig,omitempty"`
 	RestorationConfig       *restorer.RestorationConfig    `json:"restorationConfig,omitempty"`
+	CompressorConfig        *compressor.CompressionConfig  `json:"compressorConfig,omitempty"`
```
amshuman-kr, line 37:
```suggestion
	CompressionConfig        *compressor.CompressionConfig  `json:"compressionConfig,omitempty"`
```

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -241,8 +243,22 @@ func (ssr *Snapshotter) takeFullSnapshot() (*snapstore.Snapshot, error) {
 				Message: fmt.Sprintf("failed to create etcd snapshot: %v", err),
 			}
 		}
+		defer rc.Close()
```
amshuman-kr, line 246: Shouldn't this move below after the compression check?

pkg/snapshot/snapshotter/types.go
```diff
@@ -61,6 +62,7 @@ type Snapshotter struct {
 	etcdConnectionConfig *etcdutil.EtcdConnectionConfig
 	store                snapstore.SnapStore
 	config               *Config
+	compressorConfig     *compressor.CompressionConfig
```
amshuman-kr, line 65:
```suggestion
	compressionConfig     *compressor.CompressionConfig
```

amshuman-kr (changes requested):
Thanks for the PR @ishan16696! It is looking good. Some small suggestions.

Also, could you please add some tests to cover the compression cases. Especially, for restoration in the mixed case (mixing of uncompressed and compressed backups with different algo)?

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,134 @@
...
+	case LzwCompressionExtension:
+		unCompressData = lzw.NewReader(data, 0, 8) // 0 - LSB and 1- for MSB
+
+	case FlateCompressionExtension:
+		unCompressData = flate.NewReader(data)
+
+	}
+
+	totalDeCompressTime := time.Now().Sub(startDeCompressTime).Seconds()
+	logger.Infof("decompress the snapshot in %v seconds", totalDeCompressTime)
```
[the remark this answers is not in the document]
ishan16696: I agree with your compression comment but here isn't this works fine because in decompression everything is sequential as compared to compression where we run also separate thread.

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,123 @@
...
+
+// DecompressSnapshots take compressed data and snapshotSuffix as input and
+// it decompresses the data according to compression Policy and return uncompressed data.
+func DecompressSnapshots(data io.ReadCloser, snapshotSuffix string) io.ReadCloser {
+	var unCompressData io.ReadCloser
+	var err error
+
+	log := logrus.NewEntry(logrus.New())
+	logger := log.WithField("actor", "de-compressor")
+	logger.Infof("decompressing the snapshot with %v Policy", snapshotSuffix)
```
amshuman-kr, line 75: Technically logging suffix as the policy is incorrect.

pkg/snapshot/restorer/restorer.go
```diff
@@ -123,19 +124,29 @@ func (r *Restorer) restoreFromBaseSnapshot(ro RestoreOptions) error {
...
 	}
-	defer rc.Close()
+
+	isCompressed := isSnapshotCompressed(snap)
+	if isCompressed {
+		// decompress the snapshots
+		startDeCompressTime := time.Now()
+		reader = compressor.DecompressSnapshots(reader, snap.CompressionSuffix)
+		totalDeCompressTime := time.Now().Sub(startDeCompressTime).Seconds()
+		r.logger.Infof("Decompress the snapshot in %v seconds", totalDeCompressTime)
```
amshuman-kr, line 147: This logging should happen after the `io.Copy()` call because the stream reading happens there. -> https://github.com/gardener/etcd-backup-restore/pull/293/files?file-filters%5B%5D=.go#diff-9547d8cc0c155be97c8e3a057b9351b6bf83108543f957fb8bac1adf111ccfaaR160

pkg/snapshot/restorer/restorer.go
```diff
@@ -583,15 +594,24 @@ func getEventsFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapsh
...
 	}
-	defer rc.Close()
+
+	isCompressed := isSnapshotCompressed(snap)
+	if isCompressed {
+		// decompress the snapshots
+		startDeCompressTime := time.Now()
+		reader = compressor.DecompressSnapshots(reader, snap.CompressionSuffix)
+		totalDeCompressTime := time.Now().Sub(startDeCompressTime).Seconds()
+		r.logger.Infof("decompress the snapshot in %v seconds", totalDeCompressTime)
```
amshuman-kr, line 609: This logging should happen after `buf.ReadFrom()` call where the stream reading happens -> https://github.com/gardener/etcd-backup-restore/pull/293/files?file-filters%5B%5D=.go#diff-9547d8cc0c155be97c8e3a057b9351b6bf83108543f957fb8bac1adf111ccfaaR614

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -241,8 +243,24 @@ func (ssr *Snapshotter) takeFullSnapshot() (*snapstore.Snapshot, error) {
...
+		s := snapstore.NewSnapshot(snapstore.SnapshotKindFull, 0, lastRevision, compressionSuffix)
+
+		// if compression is enabled
+		//    then compress the snapshot.
+		if ssr.compressionConfig.Enabled {
+			ssr.logger.Info("starting the Compression of full snapshot")
+			startCompressTime := time.Now()
+			rc = compressor.CompressSnapshots(rc, ssr.compressionConfig.CompressionPolicy)
+			totalCompressTime := time.Now().Sub(startCompressTime).Seconds()
+			ssr.logger.Infof("Successfully compresses the full snapshot in %v seconds", totalCompressTime)
```
amshuman-kr, line 260: This should be logged after the `ssr.store.Save()` call where the stream is read -> https://github.com/gardener/etcd-backup-restore/pull/293/files?file-filters%5B%5D=.go#diff-5c59e1a2063c6043cf7cad27aee199b833950c80ec75202ac69699216210a2f1R265

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -333,8 +354,22 @@ func (ssr *Snapshotter) TakeDeltaSnapshot() (*snapstore.Snapshot, error) {
...
+	rc := ioutil.NopCloser(bytes.NewReader(ssr.events))
+
+	// if compression is enabled
+	//    then compress the snapshot.
+	if ssr.compressionConfig.Enabled {
+		ssr.logger.Info("starting the Compression of delta snapshot")
+		startCompressTime := time.Now()
+		rc = compressor.CompressSnapshots(rc, ssr.compressionConfig.CompressionPolicy)
+		totalCompressTime := time.Now().Sub(startCompressTime).Seconds()
+		ssr.logger.Infof("Successfully compresses the delta snapshot in %v seconds", totalCompressTime)
```
amshuman-kr, line 367: This logging should be done after the `ssr.store.Save()` call where the stream is read -> https://github.com/gardener/etcd-backup-restore/pull/293/files?file-filters%5B%5D=.go#diff-5c59e1a2063c6043cf7cad27aee199b833950c80ec75202ac69699216210a2f1R372

amshuman-kr (changes requested):
A small comment about the logging time. I think we can simply rely on existing tim logging for download or upload which would be enough. No additional logging time for compression is required. We can still log that we are compressing/uncompressing.

pkg/compressor/compressor.go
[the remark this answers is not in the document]
amshuman-kr: We need to log only the restoration time for each individual file. That way we can compare the times with or without compression. There is no good way to split the time for compression with the time for download or restoration for a single run.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: 1. I think there is already a separate timer ... which actually measures the time of upload. and with help of this we can measure both time... compression time as well as upload to object store time separately .
2. And When Compression is not enabled ... in that case your log still get printed but it should not.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: same Counter-argument as above

pkg/compressor/compressor.go
ishan16696: yes, I changed it to `snapshotSuffix[1:]`

pkg/snapshot/snapshotter/snapshotter.go
amshuman-kr: > I think there is already a separate timer which actually measures the time of upload and with help of this we can measure the time for both (compression time as well as upload to object store time) separately.

Yes. This is enough.

> Both timer will be helpful from future prospective like we can debug which one is taking more time either compression or upload to obj store.

No. The compression timer is not required because it won't measure only the time to construct a pipe and a compressing writer which will practically be instantaneous and not particularly meaningful.

pkg/snapshot/snapshotter/snapshotter.go
amshuman-kr: Please see https://github.com/gardener/etcd-backup-restore/pull/293#discussion_r538283599

example/00-backup-restore-server-config.yaml
```diff
@@ -46,4 +46,4 @@ defragmentationSchedule: "0 0 */3 * *"
 
 compressionConfig:
    enabled: true
-   policy: "flate"
+   policy: "lzw"
```
amshuman-kr, line 49: I would suggest the default to be gzip.

pkg/snapshot/restorer/restorer.go
```diff
@@ -139,10 +139,10 @@ func (r *Restorer) makeDB(logger *zap.Logger, snapdir string, snap snapstore.Sna
 	}
 
 	startTime := time.Now()
-	isCompressed := isSnapshotCompressed(snap)
+	isCompressed, compressionPolicy := isSnapshotCompressed(snap)
```
amshuman-kr, line 142: Nice! I liked the way you solved it :-)

amshuman-kr (commented):
Thanks for the changes. Once you have added the tests for the mixed case for restoration, this PR is good to be merged.

example/00-backup-restore-server-config.yaml
ishan16696: deafult is still gzip if we comment this configuration . see here [https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-edf8a5224f35b87bf6e0367f2b78e678d442e8329b0143d0b9653b42be7b752bR29](url)

pkg/snapshot/restorer/restorer.go
```diff
@@ -583,18 +599,32 @@ func getEventsFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapsh
 }
 
 // getEventsDataFromDeltaSnapshot fetches the events data from delta snapshot from snap store.
-func getEventsDataFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapshot) ([]byte, error) {
-	rc, err := store.Fetch(snap)
+func (r *Restorer) getEventsDataFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapshot) ([]byte, error) {
```
shreyas-s-rao, line 602: Any particular reason to add this function as a member under `Restorer` object? This has led to subsequent changes in other function signatures as well, like `getEventsFromDeltaSnapshot`, `makeWALAndSnap` and `makeDB`

pkg/snapshot/restorer/restorer.go
```diff
@@ -123,19 +124,27 @@ func (r *Restorer) restoreFromBaseSnapshot(ro RestoreOptions) error {
...
 	}
-	return makeWALAndSnap(r.zapLogger, walDir, snapdir, cl, ro.Config.Name)
+	return r.makeWALAndSnap(r.zapLogger, walDir, snapdir, cl, ro.Config.Name)
 }
 
 // makeDB copies the database snapshot to the snapshot directory.
-func makeDB(logger *zap.Logger, snapdir string, snap snapstore.Snapshot, commit int, ss snapstore.SnapStore, skipHashCheck bool) error {
-	rc, err := ss.Fetch(snap)
+func (r *Restorer) makeDB(logger *zap.Logger, snapdir string, snap snapstore.Snapshot, commit int, ss snapstore.SnapStore, skipHashCheck bool) error {
+	reader, err := ss.Fetch(snap)
```
shreyas-s-rao, line 135: Any particular reason to change the variable name from `rc` to `reader` here?

pkg/snapshot/restorer/restorer.go
```diff
@@ -583,18 +599,32 @@ func getEventsFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapsh
 }
 
 // getEventsDataFromDeltaSnapshot fetches the events data from delta snapshot from snap store.
-func getEventsDataFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapshot) ([]byte, error) {
-	rc, err := store.Fetch(snap)
+func (r *Restorer) getEventsDataFromDeltaSnapshot(store snapstore.SnapStore, snap snapstore.Snapshot) ([]byte, error) {
+	reader, err := store.Fetch(snap)
```
shreyas-s-rao, line 603: Similar to previous comment

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -436,6 +468,82 @@ var _ = Describe("Running Restorer", func() {
 				// behavior of etcd restart.
 			})
 		})
+
+		Context("when snapshots are compressed using different compressionPolicy", func() {
+			It("Should able to restore when different types of compressed and uncompressed snapshots are available", func() {
+				logger.Infoln("Starting restoration check when snapshots are availabe of different SnapshotSuffix")
```
shreyas-s-rao, line 474:
```suggestion
				logger.Infoln("Starting restoration check when snapshots are available of different SnapshotSuffix")
```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -436,6 +468,82 @@ var _ = Describe("Running Restorer", func() {
 				// behavior of etcd restart.
 			})
 		})
+
+		Context("when snapshots are compressed using different compressionPolicy", func() {
```
shreyas-s-rao, line 472: I like the concept of adding snapshots with different compression algorithms into the same snapstore and then restoring successfully from that and testing that this works. Nice! 🙂

shreyas-s-rao (changes requested):
@ishan16696 Thanks a lot for the well-written PR. Overall, code looks fine. I will test it out in the morning. Other than that, I have some minor code-related concerns/suggestions. Please address them. Thanks.
Also, please document your findings about compression and decompression (time taken, size reduction ratio, etc) for the different algorithms, so that it can serve as a reference point for someone new to compression and wanting to choose a suitable algorithm for running etcdbr.

pkg/compressor/types.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// GzipCompressionPolicy is constant for gzip compression algorithm.
+	GzipCompressionPolicy = "gzip"
+	// LzwCompressionPolicy is constant for lzw compression algorithm.
+	LzwCompressionPolicy = "lzw"
+	// ZlibCompressionPolicy is constant for zlib compression algorithm.
+	ZlibCompressionPolicy = "zlib"
+
+	// DefaultCompression is constant used for whether to compress the snapshots or not.
+	DefaultCompression = true
```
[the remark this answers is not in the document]
ishan16696: I think it will not break the current behaviour as it also supports the backward compatibility, we still can restore from the uncompressed snapshots,  but still I can make `DefaultCompression` to `false`, no issues.
CC @amshuman-kr

pkg/snapshot/restorer/restorer.go
ishan16696: no big reason ....I will change it back to `rc `

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,114 @@
...
+// DecompressSnapshots take compressed data and compressionPolicy as input and
+// it decompresses the data according to compression Policy and return uncompressed data.
+func DecompressSnapshots(data io.ReadCloser, compressionPolicy string) io.ReadCloser {
+	var unCompressData io.ReadCloser
+	var err error
+
+	logger := logrus.New().WithField("actor", "de-compressor")
+	logger.Infof("start decompressing the snapshot with %v compressionPolicy", compressionPolicy)
+
+	switch compressionPolicy {
```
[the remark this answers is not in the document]
ishan16696: I think there is no need to add default case because if compressionPolicy is not supported  then [this](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-7adcf3334be4957b8c200cd4ad81c23a84d4be71d629640291a51ba3b20a198eR46) already gonna handle  that  unsupported compression policy.

pkg/snapshot/restorer/restorer.go
ishan16696: yes , reason is I need to print the log for decompression time . you can see [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-9547d8cc0c155be97c8e3a057b9351b6bf83108543f957fb8bac1adf111ccfaaR165)

pkg/snapshot/restorer/restorer.go
ishan16696: no big reason .... changed it back to `rc`

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		logger.Infof("compressed data size: %v KB", n/1024)
+	}()
+
+	return pReader
+}
+
+// DecompressSnapshots take compressed data and compressionPolicy as input and
+// it decompresses the data according to compression Policy and return uncompressed data.
+func DecompressSnapshots(data io.ReadCloser, compressionPolicy string) io.ReadCloser {
```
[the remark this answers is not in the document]
ishan16696: I find to return the error difficult because [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-d5eff48d93149f39295c1872939affa38fd58bf430c468ce42dea93760e49253R87) lzw decompression is not returning the error

pkg/compressor/compressor.go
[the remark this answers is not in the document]
amshuman-kr: Additional safety check never hurts :-)

pkg/compressor/compressor.go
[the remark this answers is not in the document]
amshuman-kr: Additional safety check never hurts :-)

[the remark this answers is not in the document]
ishan16696: yeah ... I already handled the error for rest of 2 compression Policies and return the same data. you can  check [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-d5eff48d93149f39295c1872939affa38fd58bf430c468ce42dea93760e49253R75) and [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-d5eff48d93149f39295c1872939affa38fd58bf430c468ce42dea93760e49253R82)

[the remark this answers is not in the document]
amshuman-kr: No. I meant returning and bubbling up the error and effectively failing the snapshot upload or restoration. I would not advice silently skipping compression/decompression. It is much better to fail early.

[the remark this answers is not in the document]
ishan16696: I handled the error in `DecompressSnapshot()` but in case of `CompressSnapshot() `… error handling will be difficult because NewWriter() in all compression algos is not returning any error ….  we have to depend on error of  `n,err:= io.Copy()`  , so I  tried this but encountered some race condition when I run `make verify`, may be because `CompressSnapshot()` also have  thread running.

pkg/compressor/types.go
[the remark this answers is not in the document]
amshuman-kr: Generally, keeping things backward compatible by default is a good principle. In this case, the default behaviour for the compression as `false` makes sense in that context. We will have other levels to _default_ this to `true` (say, in the gardener integration PR).

example/00-backup-restore-server-config.yaml
```diff
@@ -42,4 +42,8 @@ restorationConfig:
   maxFetchers: 6
   embeddedEtcdQuotaBytes: 8589934592
 
-defragmentationSchedule: "0 0 */3 * *"
\ No newline at end of file
+defragmentationSchedule: "0 0 */3 * *"
+
+compressionConfig:
+   enabled: true
+   policy: "zlib"
```
amshuman-kr, line 49:
```suggestion
   policy: "gzip"
```
Let't keep `gzip` as the default.

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,138 @@
...
+
+	case ZlibCompressionPolicy:
+		gWriter = zlib.NewWriter(pWriter)
+
+	}
+
+	go func() {
+		var err error
+		defer pWriter.CloseWithError(err)
+		defer gWriter.Close()
```
amshuman-kr, line 51:
```suggestion
		defer pWriter.CloseWithError(err)
```
Only one of `Close() or `CloseWithError(err)` should be called on the `pWriter`.

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,138 @@
...
+// CompressSnapshot takes uncompressed data as input and compress the data according to Compression Policy
+// and write the compressed data into one end of pipe.
+func CompressSnapshot(data io.ReadCloser, compressionPolicy string) io.ReadCloser {
+	pReader, pWriter := io.Pipe()
+
+	var gWriter io.WriteCloser
+	logger := logrus.New().WithField("actor", "compressor")
+	logger.Infof("start compressing the snapshot using %v Compression Policy", compressionPolicy)
+
+	switch compressionPolicy {
```
amshuman-kr, line 36: Can you please add the default case and return error for unknown compression policy (you will have to change the function signature)?

pkg/compressor/compressor.go
```diff
@@ -0,0 +1,138 @@
...
+	case GzipCompressionPolicy:
+		return GzipCompressionExtension
+	}
+
+	return emptyString
+}
+
+// IsSnapshotCompressed is helpful in determining whether the snapshot is compressed or not.
+// it will return boolean and compressionPolicy corresponding to compressionSuffix.
+func IsSnapshotCompressed(compressionSuffix string) (bool, string) {
```
amshuman-kr, line 123: Should we also return error for the default case (unknown suffix/policy)?

pkg/snapshot/restorer/restorer.go
```diff
@@ -150,6 +161,13 @@ func makeDB(logger *zap.Logger, snapdir string, snap snapstore.Snapshot, commit
 		return err
 	}
 	db.Sync()
+	totalTime := time.Now().Sub(startTime).Seconds()
+
+	if isCompressed {
+		r.logger.Infof("successfully fetches data of base snapshot in %v seconds [CompressionPolicy:%v]", totalTime, compressionPolicy)
+	} else {
+		r.logger.Infof("successfully fetches data of base snapshot in %v seconds", totalTime)
+	}
```
amshuman-kr, line 170:
```suggestion
	if isCompressed {
		r.logger.Infof("successfully fetched data of base snapshot in %v seconds [CompressionPolicy:%v]", totalTime, compressionPolicy)
	} else {
		r.logger.Infof("successfully fetched data of base snapshot in %v seconds", totalTime)
	}
```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -187,6 +188,32 @@ var _ = Describe("Running Restorer", func() {
 		})
 	})
 
+	Describe("NEGATIVE:some Negative Compression Scenarios", func() {
```
amshuman-kr, line 191:
```suggestion
	Describe("NEGATIVE: Negative Compression Scenarios", func() {
```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -187,6 +188,32 @@ var _ = Describe("Running Restorer", func() {
...
+		})
+		Context("with invalid compressionPolicy", func() {
+			It("should return error", func() {
+				compressionConfig.CompressionPolicy = "someRandomAlgo"
+				err = compressionConfig.Validate()
+				Expect(err).Should(HaveOccurred())
+			})
+		})
+
+		Context("with Compression is not Enabled and invalid compressionPolicy ", func() {
```
amshuman-kr, line 206:
```suggestion
		Context("with compression is not enabled and invalid compressionPolicy ", func() {
```

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -436,6 +468,82 @@ var _ = Describe("Running Restorer", func() {
...
+				err = runSnapshotter(logger, deltaSnapshotPeriod, endpoints, ctx.Done(), true, compressionConfig)
+				Expect(err).ShouldNot(HaveOccurred())
+				cancel()
+
+				// populate the etcd with some more data
+				resp := &utils.EtcdDataPopulationResponse{}
+				utils.PopulateEtcd(testCtx, logger, endpoints, 0, keyTo, resp)
+				Expect(resp.Err).ShouldNot(HaveOccurred())
+
+				// start the Snapshotter with compressionPolicy = "lzw"
```
amshuman-kr, line 489: Can we be sure that the gzip snapshot has been taken by this time?

pkg/snapshot/restorer/restorer_test.go
amshuman-kr, line 472: Can you please add one more case where the base full snapshot itself is compressed?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -241,9 +243,23 @@ func (ssr *Snapshotter) takeFullSnapshot() (*snapstore.Snapshot, error) {
...
-		s := snapstore.NewSnapshot(snapstore.SnapshotKindFull, 0, lastRevision)
+		s := snapstore.NewSnapshot(snapstore.SnapshotKindFull, 0, lastRevision, compressionSuffix)
+
 		startTime := time.Now()
+
+		// if compression is enabled
+		//    then compress the snapshot.
+		if ssr.compressionConfig.Enabled {
+			ssr.logger.Info("start the Compression of full snapshot")
+			rc = compressor.CompressSnapshot(rc, ssr.compressionConfig.CompressionPolicy)
```
amshuman-kr, line 259: If `CompressSnapshot` returns error (if you add the default case) then it will have to be handled here.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -333,8 +352,19 @@ func (ssr *Snapshotter) TakeDeltaSnapshot() (*snapstore.Snapshot, error) {
...
+
 	startTime := time.Now()
-	if err := ssr.store.Save(*snap, ioutil.NopCloser(bytes.NewReader(ssr.events))); err != nil {
+	rc := ioutil.NopCloser(bytes.NewReader(ssr.events))
+
+	// if compression is enabled
+	//    then compress the snapshot.
+	if ssr.compressionConfig.Enabled {
+		ssr.logger.Info("start the Compression of delta snapshot")
+		rc = compressor.CompressSnapshot(rc, ssr.compressionConfig.CompressionPolicy)
```
amshuman-kr, line 363: If `CompressSnapshot` returns error (if you add the default case) then it will have to be handled here.

amshuman-kr (changes requested):
@ishan16696 Thanks a lot for the changes and also the tests. The PR is now looking good. Just some final comments below.

example/00-backup-restore-server-config.yaml
ishan16696: actually default is still gzip [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-edf8a5224f35b87bf6e0367f2b78e678d442e8329b0143d0b9653b42be7b752bR29) .... but I can change compressionPolicy in config file as well.

pkg/compressor/compressor.go
ishan16696: actually unknown compression Policy get handled [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-7adcf3334be4957b8c200cd4ad81c23a84d4be71d629640291a51ba3b20a198eR38-R51) .... but still I can default case for safety purpose (it will be unreachable code.)

pkg/compressor/compressor.go
ishan16696: actually unknown compression Policy get handled [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-7adcf3334be4957b8c200cd4ad81c23a84d4be71d629640291a51ba3b20a198eR38-R51) .... and for safety purpose: we are returning `false and emptyString` .... so I think for unknown compression Policy it will treat them as uncompressed and will return `false and emptyString`

pkg/compressor/compressor.go
amshuman-kr: Yet, there is a default case in the decompress case. Additional checks don't hurt.

example/00-backup-restore-server-config.yaml
amshuman-kr: Better to be consistent in the examples too, no?

pkg/compressor/compressor.go
ishan16696: but `Close() `is called on gWritter not on pWriter and which I think it is necessary to close gWriter as well.

amshuman-kr: Sorry. My bad. You are right.

pkg/snapshot/restorer/restorer_test.go
ishan16696: added  the test case.... please have a look

ishan16696:
/reviewed/ok-to-test

pkg/compressor/compressor.go
amshuman-kr: Thanks for addressing the default case above. Can you also address it here? The same reason applies.

amshuman-kr:
Thanks @ishan16696 for the test case and the changes. I have only a few remaining questions.

1. In the restoration tests, it is not clear to me how the first `runSnapshotter` call reliably takes a full snapshot and also how we can be sure that the snapshot are completed by the time the function call returns.
1. The default vase for `IsSnapshotCompressed` for completeness.

ishan16696:
> 1. In the restoration tests, it is not clear to me how the first `runSnapshotter` call reliably takes a full snapshot 

there is boolean variable [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-230e8e1897f9e0bcbbb2062bb32fd5ee94d1fdc8fa444efe6dd120f7281103b2R578)  by which we configure either take fullSnapshot or deltaSnapshot.
> how we can be sure that the snapshot are completed by the time the function call returns.

I am also not so sure .... but `runSnapshotter` is returning an error so by this error we can be sure that `runSnapshotter` is completed otherwise [here](https://github.com/gardener/etcd-backup-restore/pull/293/files#diff-230e8e1897f9e0bcbbb2062bb32fd5ee94d1fdc8fa444efe6dd120f7281103b2R565) test get fails.

>  The default vase for `IsSnapshotCompressed` for completeness.

done

amshuman-kr (approved):
Thanks for the changes @ishan16696! LGTM

shreyas-s-rao (approved):
@ishan16696 Thanks for making all the changes. LGTM 👍🏼

PR #294, opened 2020-12-07
Added support for running tests suites from user specified packages.

abdasgupta:
Currently, tests  run for `cmd` and `pkg` . But there is no provision to run tests just on any package of etcd-backup-retstore. This PR adds the support. It makes easier for developer to write/modify packages and test on just that package iteratively.

### DIFF
--- a/.ci/unit_test
+++ b/.ci/unit_test
@@ -54,7 +54,12 @@ function test_with_coverage() {
 
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


abdasgupta:
cc @shreyas-s-rao

amshuman-kr (approved):
Thanks for the added flexibility! LGTM :-)

shreyas-s-rao (approved):

PR #315, opened 2021-03-16
Added Auto-Compaction for embedded Etcd and Update helm-charts to run auto-compaction on Etcd.

ishan16696:
Introduce the auto-compaction-mode and auto-compaction-retention CLI flags for embedded ETCD to use auto-compaction during restoration.

Updated the helm-charts to run auto-compaction on Etcd as well as on embedded-Etcd of etcd-backup-restore sidecar.

Fixes  # [136](https://github.com/gardener/etcd-druid/issues/136) (Partially)

**Release note**:
```improvement user
Added CLI flags `--auto-compaction-mode` and `--auto-compaction-retention` to configure auto-compaction for embedded etcd. Default values: auto-compaction-mode="periodic" and auto-compaction-retention="30m"
```

### DIFF
--- a/chart/etcd-backup-restore/templates/etcd-bootstrap-configmap.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-bootstrap-configmap.yaml
@@ -113,6 +113,18 @@ data:
     # Initial cluster state ('new' or 'existing').
     initial-cluster-state: 'new'
 
+    {{- if .Values.autoCompaction }}
+    # auto-compaction-mode ("periodic" or "revision").
+    {{- if .Values.autoCompaction.mode }}
+    auto-compaction-mode: {{ .Values.autoCompaction.mode }}
+    {{- end }}
+
+    # auto-compaction-retention defines Auto compaction retention length for etcd.
+    {{- if .Values.autoCompaction.retentionLength }}
+    auto-compaction-retention: {{ .Values.autoCompaction.retentionLength }}
+    {{- end }}
+    {{- end }}
+
 {{- if .Values.etcdTLS }}
     client-transport-security:
       # Path to the client server TLS cert file.
--- a/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
@@ -138,6 +138,14 @@ spec:
         - --compression-policy={{ .Values.backup.compression.policy }}
         {{- end }}
 {{- end }}
+{{- if .Values.autoCompaction }}
+        {{- if .Values.autoCompaction.mode }}
+        - --auto-compaction-mode={{ .Values.autoCompaction.mode }}
+        {{- end }}
+        {{- if .Values.autoCompaction.retentionLength }}
+        - --auto-compaction-retention={{ .Values.autoCompaction.retentionLength }}
+        {{- end }}
+{{- end }}
         image: {{ .Values.images.etcdBackupRestore.repository }}:{{ .Values.images.etcdBackupRestore.tag }}
         imagePullPolicy: {{ .Values.images.etcdBackupRestore.pullPolicy }}
         ports:
--- a/chart/etcd-backup-restore/values.yaml
+++ b/chart/etcd-backup-restore/values.yaml
@@ -33,6 +33,12 @@ servicePorts:
 
 storageCapacity: 20Gi
 
+# auto-compaction mode for etcd: 'periodic' mode or 'revision' mode.
+# auto-compaction retention length for etcd.
+autoCompaction:
+  mode: periodic
+  retentionLength: "5m"
+
 backup:
   # schedule is cron standard schedule to take full snapshots.
   schedule: "0 */1 * * *"
--- a/chart/etcd-backup-restore/values.yaml
+++ b/chart/etcd-backup-restore/values.yaml
@@ -7,7 +7,7 @@ images:
   # etcd-backup-restore image to use
   etcdBackupRestore:
     repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
-    tag: v0.12.0-dev-e6df91b7463cf784f651bd6cc6ac7fc4d2e5a70b
+    tag: v0.12.0-dev-34f854c06838d14d6e7b08099d6cfa6fca2ad50d
     pullPolicy: IfNotPresent
 
 resources:
@@ -33,11 +33,12 @@ servicePorts:
 
 storageCapacity: 20Gi
 
-# auto-compaction mode for etcd: 'periodic' mode or 'revision' mode.
-# auto-compaction retention length for etcd.
+# autoCompaction defines the specification to be used by Etcd as well as by embedded-Etcd of backup-restore sidecar during restoration.
+# auto-compaction mode for etcd and embedded-Etcd: 'periodic' mode or 'revision' mode.
+# auto-compaction retention length for etcd as well as for  embedded-Etcd of backup-restore sidecar.
 autoCompaction:
   mode: periodic
-  retentionLength: "5m"
+  retentionLength: "30m"
 
 backup:
   # schedule is cron standard schedule to take full snapshots.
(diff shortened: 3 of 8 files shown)


shreyas-s-rao:
Adding label `reviewed/do-not-merge` in order to finish related performance tests first.

amshuman-kr:
/hold until impact of frequent revision auto-compaction has been evaluated.

1. Potential DB size increase if compaction is not also followed by defragmentation -> https://www.compose.com/articles/how-to-keep-your-etcd-lean-and-mean/#letsmanuallycompact 
1. if having `auto-compaction period <= incremental backup period` leads to lost revisions especially when the watch connections of the incremental snapshotter is disrupted for long enough interval. This would be the case if set auto-compaction period to `5m`. This scenario can be manually simulated by disabling incremental snapshots for a suitable duration (by setting the interval to `0`), letting the compaction happen and then enabling the incremental snapshots back again.

ishan16696:
> 1. Potential DB size increase if compaction is not also followed by defragmentation -> https://www.compose.com/articles/how-to-keep-your-etcd-lean-and-mean/#letsmanuallycompact


I have tested the `Etcd` for the following Scenarios .

**Note**: 
> -  No of Revision is cumulative.
> -  I have applied defragmentation only at the end of each testing.

-  #### Large `Etcd` but here I have to test with manual Compaction rather than auto-compaction as the rate of the benchmark tool was very slow.

image

**Observation**:
1. As you can see DBsize is not much affected by compaction 😊
2. After compaction, when we put more data DBsize grows very slowly (here for 300k more revisions DBsize has increased only from 704MB to 776MB.)
3. There is a significant reduction in `DBsize` after applying `defragmentation` (from 845MB reduced to 553MB)


-  #### For `auto-compaction-retention`: 1hr 

image

**Observation**:
1. Here also DBsize is not much affected by compaction 😊
2. Similar to the above `Etcd` test here also confirms the behavior that After compaction when we put more data DBsize grows very slowly.
2. There is not a significant reduction in `DBsize` after applying `defragmentation` (from 153MB reduced to 138MB) which is kind of a little baffling.

ishan16696:
> 2. if having `auto-compaction period <= incremental backup period` leads to lost revisions especially when the watch connections of the incremental snapshotter is disrupted for long enough interval. This would be the case if set auto-compaction period to `5m`. This scenario can be manually simulated by disabling incremental snapshots for a suitable duration (by setting the interval to `0`), letting the compaction happen, and then enabling the incremental snapshots back again.

I have simulated this scenario, following are my observations:
**Observations**:
It will take a full snapshot when it fails to take the delta snapshot because [here](https://github.com/ishan16696/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L207-L234)  if you see `initialDeltaSnapshotTaken`  will remain `false` due to an error occurred since the required revision has been compacted, hence it will endup [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L231) and take full snapshot.

Logs:
```
WARN[0000] Failed to collect events for first delta snapshot(s): etcdserver: mvcc: required revision has been compacted  actor=backup-restore-server
INFO[0000] Taking scheduled snapshot for time: 2021-04-08 11:14:15.740048 +0530 IST  actor=snapshotter
{"level":"warn","ts":"2021-04-08T11:14:15.741+0530","caller":"clientv3/retry_interceptor.go:116","msg":"retry stream intercept"}
INFO[0000] Successfully opened snapshot reader on etcd   actor=snapshotter
INFO[0000] start the Compression of full snapshot        actor=snapshotter
INFO[0000] start compressing the snapshot using gzip Compression Policy  actor=compressor
INFO[0000] Total time to save snapshot: 0.028607 seconds.
INFO[0000] Successfully saved full snapshot at: Backup-1617860655/Full-00000000-00000991-1617860655.gz  actor=snapshotter
INFO[0000] Applied watch on etcd from revision: 992
```
**Note**:  In this Scenario, FullSnapshot will only be taken if the sidecar container got restarted but if this issue of the `apply watch mechanism on compacted revision` happens [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/snapshotter/snapshotter.go#L507-L557), then it will result in an error.

amshuman-kr:
> There is not a significant reduction in DBsize after applying defragmentation (from 153MB reduced to 138MB) which is kind of a little baffling.

When you did manual compaction, to which revision did you compact? Latest revision as of them?

ishan16696:
> When you did manual compaction, to which revision did you compact? Latest revision as of then?

Which case you are referring to?  I have clearly mentioned in the image of each test case.

amshuman-kr:
> Which case you are referring to? I have clearly mentioned in the image of each test case.

I see the number of revisions but I meant what is the revision you massed to the `compaction` command -> https://etcd.io/docs/v3.4/dev-guide/interacting_v3/#compacted-revisions

ishan16696:
> I see the number of revisions but I meant what is the revision you passed to the `compaction` command -> https://etcd.io/docs/v3.4/dev-guide/interacting_v3/#compacted-revisions

Got your point @amshuman-kr ,
For **Test1** (Large Etcd): I did manual compaction at No_of_Revisions/2 eg: No_of_revisions= 500000 then `etcdctl compact 250000`.

For Test2: auto-compaction-retention was set to 1hr, I don't remember at which revision it did the compaction as it was automated

ishan16696:
And it will not compact the latest revision, Just to avoid confusion , see this eg:
`--auto-compaction-retention=5m`  means we are telling etcd to keep the history of last 5mins and rest history it can compact.

amshuman-kr:
> For **Test1** (Large Etcd): I did manual compaction at No_of_Revisions/2 eg: No_of_revisions= 500000 then `etcdctl compact 250000`.
> 
> For Test2: auto-compaction-retention was set to 1hr, I don't remember at which revision it did the compaction as it was automated

Do you have any idea about how many keys are generated in the database by the benchmark tool? If it generates unique keys then compaction will have little impact (because compaction reclaims only _superceded_ value, i.e. new values for the same keys).

ishan16696:
> Do you have any idea about how many keys are generated in the database by the benchmark tool? 

Yes, I generated the keys with this command 
```
benchmark  --conns=1 --clients=1 \
    put --key-size=16 --sequential-keys --total=100000 --val-size=512
```
But not sure about How many unique keys have generated by Benchmark tool.

> If it generates unique keys then compaction will have little impact (because compaction reclaims only superceded value, i.e. new values for the same keys).

Yes, you are correct.

ishan16696:
I have re-tested the Etcd for the following Scenarios, Please have a look.

Note:
1. No of Revision is cumulative.
2. `keySpaceSize` can be interpreted as no. of unique keys (please refer benchmark code [here](https://github.com/etcd-io/etcd/blob/master/tools/benchmark/cmd/put.go#L86) and [here](https://github.com/etcd-io/etcd/blob/master/tools/benchmark/cmd/put.go#L109-L115) ).
3. `keySize`=32B and `valueSize`=1024B have taken for all test Scenarios .
4. Applied defragmentation on Etcd only at the end of each testing. 


#### Large Etcd but here I have to test with manual Compaction at 3*(No_of_Revisions)/4
image

#### Small Etcd with Small keySpaceSize and with auto-compaction-retention set to 15mins
image


#### For KeySpaceSize=500 for Medium size Etcd.
image

#### Observations:

1. Physical DBsize is not much affected by compaction in all test Scenarios.
2. There is a significant reduction in Logical DBSize after compaction is applied in all test Scenarios.
3. After compaction,  defragmentation is applied then there is a significant reduction in Physical DBsize.
            -  For large Etcd, DBsize decreases from 729MB to 346MB 
            -  Similar behaviour can be observed for the other 2 tests Scenarios too.

ishan16696:
Keeping in mind of [this](https://github.com/gardener/etcd-backup-restore/pull/315#issuecomment-815108726) Scenario, I have changed the `defaultAutoCompactionRetention` time from 5m to **30m**.

cc @amshuman-kr

shreyas-s-rao (approved):
@ishan16696 thanks for the well-written PR. LGTM

PR #319, opened 2021-04-09
depend on publish-trait rather than injected step(s)

ccwienk:
prepares switch to kaniko-builder


ccwienk:
@amshuman-kr : could you estimate when you would be able to review this PR? it will not change anything w.r.t. the pipeline's behaviour (but in the current state, it blocks kaniko-builder from being set as default in our pipeline-template)

shreyas-s-rao (approved):

PR #320, opened 2021-04-09
Refactor pipeline def rel v0.11.0

ccwienk:

prepare for switch to kaniko-builder


ccwienk:
I think you should switch to python3 + the integrationtest-errs are not related to the change introduced by this PR, I think

shreyas-s-rao:
Unrelated error. Merging this PR. @ccwienk Thanks for the PR and the suggestion to switch to python3. Deprecated python2 is becoming a mess to maintain.

PR #344, opened 2021-06-09
Fixed Test Machinary integration tests.

abdasgupta:

Fixes the error in Test Machinary tests that are arising due to Backup version change for compaction support.

### DIFF
--- a/pkg/snapshot/restorer/restorer_test.go
+++ b/pkg/snapshot/restorer/restorer_test.go
@@ -38,9 +38,16 @@ import (
 )
 
 const (
-	allSnapsInV1 = "allSnapsInV1"
-	fullSnapInV1 = "fullSnapInV1"
-	fullSnapInV2 = "fullSnapInV2"
+	v1 = iota
+	v2
+	mixed
+)
+
+const (
+	allSnapsInV1 = iota
+	fullSnapInV1
+	fullSnapInV2
+	allSnapsInV2
 )
 
 var _ = Describe("Running Restorer", func() {
@@ -65,7 +72,7 @@ var _ = Describe("Running Restorer", func() {
 		maxTxnOps                      = 2 * 1024
 		embeddedEtcdQuotaBytes  int64  = 8 * 1024 * 1024 * 1024
 		autoCompactionMode      string = "periodic"
-		autoCompactionRetention string = "5m"
+		autoCompactionRetention string = "0"
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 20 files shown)


pkg/snapshot/restorer/restorer_test.go
```diff
@@ -665,290 +665,338 @@ var _ = Describe("Running Restorer", func() {
...
+			err = utils.CheckDataConsistency(testCtx, restoreOpts.Config.RestoreDataDir, stop, logger)
+			Expect(err).ShouldNot(HaveOccurred())
 		})
+	})
 
-		Context("With first few snapshots in v1 dir and some more incr snapshots are in v2 dir", func() {
-			It("should restore from v1 dir and v2 dir", func() {
-				memberPath := path.Join(etcdDir, "member")
+	Context("With first few snapshots in v1 dir and some more incr snapshots are in v2 dir", func() {
+		It("should restore from v1 dir and v2 dir", func() {
```
shreyas-s-rao, line 932: Shouldn't you also be taking some v2 snapshots in this spec? I see that only v1 snapstore is being populated before checking for restoration.

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -665,290 +665,338 @@ var _ = Describe("Running Restorer", func() {
...
+			err = utils.CheckDataConsistency(testCtx, restoreOpts.Config.RestoreDataDir, resp.KeyTo, logger)
+			Expect(err).ShouldNot(HaveOccurred())
 		})
+	})
 
-		Context("With first few snapshots in v2 dir and some more incr snapshots are in v1 dir", func() {
-			It("should restore from v1 dir and v2 dir", func() {
-				memberPath := path.Join(etcdDir, "member")
+	Context("With first few snapshots in v2 dir and some more incr snapshots are in v1 dir", func() {
+		It("should restore from v1 dir and v2 dir", func() {
```
shreyas-s-rao, line 967: Shouldn't you also be taking some v2 snapshots first in this spec before taking v1 snapshots? I see that only v1 snapstore is being populated before checking for restoration.

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -961,62 +1009,63 @@ func corruptEtcdDir() error {
 
 //takeValidV1Snaps saves valid snaps in the v1 prefix dir of snapstore so that restorer could restore from them
 //TODO: Consider removing when backward compatibility no longer needed
-func takeValidV1Snaps(logger *logrus.Entry, container string, deltaSnapshotPeriod time.Duration, endpoints []string, stopCh <-chan struct{}, compressionConfig *compressor.CompressionConfig, mode string) error {
+func takeValidV1Snaps(logger *logrus.Entry, container string, resp *utils.EtcdDataPopulationResponse, deltaSnapshotPeriod time.Duration, endpoints []string, stopCh <-chan struct{}, compressionConfig *compressor.CompressionConfig, mode string) error {
 	//Here we run the snapshotter to take snapshots. The snapshotter by default stores the snaps in the v2 directory.
```
shreyas-s-rao, line 1013: Can you encapsulate this functionality into a separate function like `takeV2Snaps` and then use it within `takeValidV1Snaps` and `takeInvalidV1Snaps` functions? So that you can also reuse the `takeV2Snaps` function in all the test cases under `Running Restorer when both v1 and v2 directory structures are present` too.

test/e2e/integrationcluster/backup_test.go
```diff
@@ -202,9 +203,32 @@ var _ = Describe("Backup", func() {
...
 			Expect(err).ShouldNot(HaveOccurred())
 
-			Expect(latestSnapshots.FullSnapshot).Should(Or(Equal(fullSnap), Equal(newFullSnap)))
-			for i, snap := range latestSnapshots.DeltaSnapshots {
-				Expect(snap).Should(Or(Equal(deltaSnapList[i]), Equal(newDeltaSnapList[i])))
+			if latestSnapshots.FullSnapshot == nil {
+				Expect(latestSnapshots.FullSnapshot).Should(Or(Equal(fullSnap), Equal(newFullSnap)))
+			} else {
+				// prefix is not determined during http call So prefix can't be tested here.
+				Expect(*latestSnapshots.FullSnapshot).To(MatchFields(IgnoreExtras, Fields{
```
shreyas-s-rao, line 210: I don't think we need to match each field here, since `.Equals()` already does that for us right?

test/utils/utils.go
```diff
@@ -66,7 +66,7 @@ func StartEmbeddedEtcd(ctx context.Context, etcdDir string, logger *logrus.Entry
 	cfg.InitialCluster = cfg.InitialClusterFromName(cfg.Name)
 	cfg.Logger = "zap"
 	cfg.AutoCompactionMode = "periodic"
-	cfg.AutoCompactionRetention = "5m"
+	cfg.AutoCompactionRetention = "0"
```
shreyas-s-rao, line 69: Any particular reason for disabling auto-compaction? Gardener etcds usually have auto-compaction enabled, which is why keeping it enabled here mimics the behavior of production etcds to the closest extent possible.

shreyas-s-rao (changes requested):
@abdasgupta Thanks for the fixes to the tests. Great to know that all the tests have now passed! I had a few comments though. PTAL.

amshuman-kr (commented):
Thanks for the fixes @abdasgupta! A couple of small queries below.

test/utils/utils.go
abdasgupta: There was an flaky error coming due to this. I dedicated this PR only for fixing the build. I think, further improvement of the test case should be done in another PR. It would be good for future as well.

test/e2e/integrationcluster/backup_test.go
abdasgupta: `.Equals()` won't work here as I mentioned in the commented line.

pkg/snapshot/restorer/restorer_test.go
abdasgupta: Hmm.. I need to retouch these test cases again.

test/e2e/integrationcluster/backup_test.go
amshuman-kr: Please ignore my comment. The [old code](https://github.com/abdasgupta/etcd-backup-restore/blame/13ddcaa7ec25be4b23f204124ac3f718fe9920c9/test/e2e/integrationcluster/backup_test.go#L205-L208) was also checking both old and new snapshots. It was using `Equals` but because of prefixes in the snapshot paths, use of `MatchFields` was required) 👍 

@shreyas-s-rao Any idea why the old code was checking both old and new snapshots?

amshuman-kr (approved):
/lgtm @shreyas-s-rao Can you please check if your comments are also addressed?

pkg/snapshot/restorer/restorer_test.go
abdasgupta: Done

shreyas-s-rao (approved):
@abdasgupta Thanks for making the changes. LGTM

PR #355, opened 2021-07-05
Updated compact subcommend to throw warning instead of error if revisions are already compacted

aaronfern:
Currently the compact subcommand raises a fatal warning if compaction stops for any reason. Updated this behaviour to throw a warning for a special case where compaction does not happen if revisions are already compacted.

Fixes #

**Release note**:

```improvement user
Compaction job will now throw warning instead of error if revisions are already compacted
```

### DIFF
--- a/cmd/compact.go
+++ b/cmd/compact.go
@@ -16,10 +16,12 @@ package cmd
 
 import (
 	"context"
+	"strings"
 
 	"github.com/gardener/etcd-backup-restore/pkg/compactor"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"go.etcd.io/etcd/mvcc"
 )
 
 // NewCompactCommand compacts the ETCD instance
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-2:
Thank you @aaronfern for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

amshuman-kr (commented):
Thanks for the fix @aaronfern! Are we sure if `ErrCompacted` is the only error that happens when there is nothing to compact?

How about the case where there are no delta snapshots at all?

aaronfern:
> Are we sure if `ErrCompacted` is the only error that happens when there is nothing to compact?

Well, It's the only error I've encountered being thrown up when there is nothing to compact.
Looking at the defined errors, it seems likely to be the only one, but I can't say with certainty that it's the only one that happens.


> How about the case where there are no delta snapshots at all?

I did try out this case and it does not throw up an error, so nothing to explicitly handle here

amshuman-kr (approved):
Thanks for the clarification @aaronfern! /lgtm

shreyas-s-rao (approved):
@aaronfern Nice catch with the error as well as the log message. LGTM

abdasgupta (approved):
LGTM

abdasgupta:
Is it rebased and pushed again after merge of test cases fix?

shreyas-s-rao:
@abdasgupta Ah yes that's the issue. @aaronfern won't be able to rebase this for the next 10 days. I will raise another PR and cherry-pick his change in that.

abdasgupta:
> @abdasgupta Ah yes that's the issue. @aaronfern won't be able to rebase this for the next 10 days. I will raise another PR and cherry-pick his change in that.

Or Maybe, we can ask him over the phone if he has access to a laptop with git installed.

shreyas-s-rao:
> Or Maybe, we can ask him over the phone if he has access to a laptop with git installed.

@abdasgupta I would suggest against doing that, since he's on vacation :)

shreyas-s-rao:
Closing in favour of https://github.com/gardener/etcd-backup-restore/pull/358
/cc @aaronfern

PR #367, opened 2021-08-04
Propagated data directory from user flags to compaction.

abdasgupta:
So far, users are provided with `--data-dir` flag for compaction subcommand but that was not really propagated to compaction as it was using it was setting it's own data-dir [https://github.com/gardener/etcd-backup-restore/blob/1638bef219bef7c5367875d78f8cd6a83ca5c539/pkg/compactor/compactor.go#L70](https://github.com/gardener/etcd-backup-restore/blob/1638bef219bef7c5367875d78f8cd6a83ca5c539/pkg/compactor/compactor.go#L70)

This PR propagates the options provided through `--data-dir` flag and propagate to compaction.

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -59,7 +59,11 @@ func (cp *Compactor) Compact(ro *brtypes.RestoreOptions, needDefragmentation boo
 	}
 
 	// Set a temporary etcd data directory for embedded etcd
-	cmpctDir, err := ioutil.TempDir("/tmp", "compactor-")
+	prefix := cmpctOptions.Config.RestoreDataDir
+	if prefix == "" {
+		prefix = "/tmp"
+	}
+	cmpctDir, err := ioutil.TempDir(prefix, "compactor-")
 	if err != nil {
 		cp.logger.Errorf("Unable to create temporary etcd directory for compaction: %s", err.Error())
 		return nil, err
--- a/pkg/compactor/compactor_test.go
+++ b/pkg/compactor/compactor_test.go
@@ -63,15 +63,16 @@ var _ = Describe("Running Compactor", func() {
 			cptr = compactor.NewCompactor(store, logger)
 			restoreOpts = &brtypes.RestoreOptions{
 				Config: &brtypes.RestorationConfig{
-					InitialClusterToken:      restoreClusterToken,
 					InitialCluster:           restoreCluster,
-					Name:                     restoreName,
+					InitialClusterToken:      restoreClusterToken,
+					RestoreDataDir:           "/tmp",
 					InitialAdvertisePeerURLs: restorePeerURLs,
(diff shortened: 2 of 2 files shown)


amshuman-kr (approved):
Thanks for the fix @abdasgupta! LGTM

How about the folder where the folder where the snapshots are downloaded? Is it required to propagate the configuration for that too?

pkg/compactor/compactor.go
```diff
@@ -59,7 +59,11 @@ func (cp *Compactor) Compact(ro *brtypes.RestoreOptions, needDefragmentation boo
 	}
 
 	// Set a temporary etcd data directory for embedded etcd
-	cmpctDir, err := ioutil.TempDir("/tmp", "compactor-")
+	prefix := cmpctOptions.Config.RestoreDataDir
+	if prefix == "" {
+		prefix = "/tmp"
+	}
+	cmpctDir, err := ioutil.TempDir(prefix, "compactor-")
```
aaronfern, line 66: We should check whether the `prefix` path exists. `ioutils.TempDir` could potentially throw up an error if it's not a valid dir

abdasgupta:
> Thanks for the fix @abdasgupta! LGTM
> 
> How about the folder where the folder where the snapshots are downloaded? Is it required to propagate the configuration for that too (probably configured using the flag `--snapstore-temp-directory`)?

Yes, we can do that. We can pass a tmpfs path to there also..

pkg/compactor/compactor.go
abdasgupta: That's fine. Even if the path doesn't exist, compaction still should throw error

amshuman-kr:
> Yes, we can do that. We can pass a tmpfs path to there also..

@abdasgupta Is the `--snapstore-temp-directory` configuration already passed and honoured in `compact` subcommand?

abdasgupta:
> > Yes, we can do that. We can pass a tmpfs path to there also..
> 
> @abdasgupta Is the `--snapstore-temp-directory` configuration already passed and honoured in `compact` subcommand?

I checked that already.. So here is what happens, compact subcommand doesn't intervene `--snapstore-temp-directory` like the way it was interrupting restoredir. compaction sub command provide option for `----snapstore-temp-directory` and that is passed as snapstore config to storage functions like save or get.

amshuman-kr:
> I checked that already.. So here is what happens, compact subcommand doesn't intervene --snapstore-temp-directory like the way it was interrupting restoredir. compaction sub command provide option for ----snapstore-temp-directory and that is passed as snapstore config to storage functions like save or get.

So, `compact` sub-command supports custom values for `--snapstore-temp-directory` then, right? If so, do we need to configure it properly in the cronjob template to make sure the tem files go into the right mounted volume?

PR #369, opened 2021-08-09
Adding copy operation and additional authentication parameters forsource and target

### LINKED ISSUE #356: [Feature] Add ability to copy backups between backup buckets
As part of the [Control Plane Migration] topic(https://github.com/gardener/gardener/blob/c00ec60c8423d5b95598e7a764d67f366f8c06b6/docs/proposals/07-shoot-control-plane-migration.md) (note that this is the revised GEP) we would like to add the functionality to copy shoot ETCD backups between seed backup buckets. 

Currently, when a shoot cluster is migrated to a new seed, its backup is kept in the old seed's backup bucket. This means that seeds can have access to backup buckets which do not belong to them and also means that seeds can have BackupEntries which point to BackupBuckets of other seeds.

There is already a ready POC which was outlined in this issue: https://github.com/gardener/gardener/issues/3875 and largely developed in this repo https://github.com/stoyanr/etcd-backup-restore/tree/cpm-poc/master

The plan is to first develop the backup copying as part of the Control Plane Migration "good case scenario" and only later on add functionality for "owner detection" which will be necessary for the "bad case scenario".

kris94:
1. Additional command `copy` is added with the following syntax and parameters. This command can copy data between two buckets.
`$ etcdbrctl copy [flags]`
>    --max-parallel-chunk-uploads uint   maximum number of parallel chunk uploads allowed  (default 5)
      --snapstore-temp-directory string   temporary directory for processing (default "/tmp")
      --source-storage-provider string    source snapshot storage provider
      --source-store-container string     source container which will be used as snapstore
      --source-store-prefix string        source prefix or directory inside container under which snapstore is created
      --storage-provider string           snapshot storage provider
      --store-container string            container which will be used as snapstore
      --store-prefix string               prefix or directory inside container under which snapstore is created

2.  The `ABS`, `S3`, `GCS` `SWIFT` and `OSS` snap stores are enhanced  to accept `SOURCE_` environment variables in order to be able to authenticate to the `source` and `target` snapstores that will be used for the copy operation.

More information about this feature and why it is needed [Issue 356](https://github.com/gardener/etcd-backup-restore/issues/356)
Fixes #356 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/cmd/copy.go
+++ b/cmd/copy.go
@@ -1,0 +1,55 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+
+	"github.com/gardener/etcd-backup-restore/pkg/snapshot/copier"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// NewCopyCommand creates a cobra command for copy.
+func NewCopyCommand(ctx context.Context) *cobra.Command {
+	opts := newCopierOptions()
+	var command = &cobra.Command{
+		Use:   "copy",
+		Short: "copy data between buckets",
+		Long:  `Copy data between buckets`,
+		Run: func(cmd *cobra.Command, args []string) {
+			printVersionInfo()
+			logger := logrus.NewEntry(logrus.New())
+			if err := opts.validate(); err != nil {
+				logger.Fatalf("failed to validate the options: %v", err)
+			}
+			opts.complete()
+
+			sourceStorage, destStorage, err := copier.GetSourceAndDestinationStores(opts.sourceSnapStoreConfig, opts.snapstoreConfig)
+			if err != nil {
+				logger.Fatalf("Could not get source and destination snapstores: %v", err)
+			}
+
+			copier := copier.NewCopier(sourceStorage, destStorage, logger, opts.maxBackups, opts.maxBackupAge)
+			if err := copier.Run(); err != nil {
+				logger.Fatalf("Copy operation failed: %v", err)
+			}
+
+			logger.Info("Shutting down...")
+		},
+	}
+	opts.addFlags(command.Flags())
+	return command
+}
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -16,16 +16,16 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"io/ioutil"
 
 	"github.com/gardener/etcd-backup-restore/pkg/compressor"
 	"github.com/gardener/etcd-backup-restore/pkg/snapshot/snapshotter"
+	"github.com/gardener/etcd-backup-restore/pkg/types"
 
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
-
-	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
-
 	"github.com/gardener/etcd-backup-restore/pkg/initializer/validator"
+	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
 
 	"github.com/gardener/etcd-backup-restore/pkg/server"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
@@ -254,11 +254,51 @@ func (c *snapshotterOptions) validate() error {
 	if err := c.compressionConfig.Validate(); err != nil {
 		return err
 	}
-
 	return c.etcdConnectionConfig.Validate()
 }
 
 // complete completes the config.
 func (c *snapshotterOptions) complete() {
 	c.snapstoreConfig.Complete()
+}
+
+type copierOptions struct {
+	sourceSnapStoreConfig *types.SnapstoreConfig
+	snapstoreConfig       *types.SnapstoreConfig
+	maxBackups            int
+	maxBackupAge          int
+}
+
+func newCopierOptions() *copierOptions {
+	sourceSnapStoreConfig := snapstore.NewSnapstoreConfig()
+	sourceSnapStoreConfig.IsSource = true
+	return &copierOptions{
+		sourceSnapStoreConfig: sourceSnapStoreConfig,
+		snapstoreConfig:       snapstore.NewSnapstoreConfig(),
+	}
+}
+
+func (c *copierOptions) addFlags(fs *flag.FlagSet) {
+	fs.IntVar(&c.maxBackups, "max-backups-to-copy", -1, "copy the specified number of backups sorted by date from newest to oldest")
+	fs.IntVar(&c.maxBackupAge, "max-backup-age", -1, "copy only the backups not older than the specified number of days")
+	c.sourceSnapStoreConfig.AddSourceFlags(fs)
+	c.snapstoreConfig.AddFlags(fs)
+}
+
+func (c *copierOptions) validate() error {
+	if c.maxBackups < -1 {
+		return errors.New("parameter max-backups-to-copy must not be less than -1")
+	}
+	if c.maxBackupAge < -1 {
+		return errors.New("parameter max-backup-age must not be less than -1")
+	}
+	if err := c.snapstoreConfig.Validate(); err != nil {
+		return err
+	}
+	return c.sourceSnapStoreConfig.Validate()
+}
+
+func (c *copierOptions) complete() {
+	c.snapstoreConfig.Complete()
+	c.sourceSnapStoreConfig.MergeWith(c.snapstoreConfig)
 }
--- a/doc/usage/getting_started.md
+++ b/doc/usage/getting_started.md
@@ -144,3 +144,35 @@ INFO[0008] Successfully restored the etcd data directory.
 ### Etcdbrctl server
 
 With sub-command `server` you can start a http server which exposes an endpoint to initialize etcd over REST interface. The server also keeps the backup schedule thread running to keep taking periodic backups. This is mainly made available to manage an etcd instance running in a Kubernetes cluster. You can deploy the example [helm chart](../../chart/etcd-backup-restore) on a Kubernetes cluster to have a fault-resilient, self-healing etcd cluster.
+
+## Etcdbrctl copy
+
+With sub-command `copy` you can copy all snapshots (Full and Delta) fom one snapstore to another. Using the two filter parameters `max-backups-to-copy` and `max-backup-age` you can also limit the number of snapshots that will be copied or target only the newest snapshots.
+
+```console
+$ ./bin/etcdbrctl copy \
+--storage-provider="GCS" \
+--snapstore-temp-directory="/temp" \
+--store-prefix="target-prefix" \
+--store-container="target-container" \
+--source-store-prefix="prefix" \
+--source-store-container="container" \
+--source-storage-provider="GCS" \
+--max-backup-age=15 \
+INFO[0000] etcd-backup-restore Version: v0.14.0-dev     
+INFO[0000] Git SHA: b821ee55                            
+INFO[0000] Go Version: go1.16.5                         
+INFO[0000] Go OS/Arch: darwin/amd64                     
+INFO[0000] Getting source backups...                     actor=copier
+...
+INFO[0026] Copying Incr snapshot Incr-ID.gz...  actor=copier
+INFO[0027] Uploading snapshot of size: 123456, chunkSize: 123456, noOfChunks: 1
+INFO[0027] Triggered chunk upload for all chunks, total: 1
+INFO[0027] No of Chunks:= 1
+INFO[0027] Uploading chunk with offset : 0, attempt: 0
+INFO[0027] Received chunk result for id: 1, offset: 0
+INFO[0027] Received successful chunk result for all chunks. Stopping workers.
+INFO[0027] All chunk uploaded successfully. Uploading composite object.
+INFO[0027] Composite object uploaded successfully.
+INFO[0027] Shutting down...
+```
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -61,6 +61,61 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
 	return fullSnapshot, deltaSnapList, nil
 }
 
+type backup struct {
+	FullSnapshot      *brtypes.Snapshot
+	DeltaSnapshotList brtypes.SnapList
+}
+
+// GetFilteredBackups returns sorted by date (new -> old) SnapList. It will also filter the snapshots that should be included or not using the filter function.
+// If the filter is nil it will return all snapshots. Also, maxBackups can be used to target only the last N snapshots (-1 = all).
+func GetFilteredBackups(store brtypes.SnapStore, maxBackups int, filter func(snaps brtypes.Snapshot) bool) (brtypes.SnapList, error) {
+	snapList, err := store.List()
+	if err != nil {
+		return nil, err
+	}
+	backups := getStructuredBackupList(snapList)
+	sort.Slice(backups, func(i, j int) bool {
+		return backups[i].FullSnapshot.CreatedOn.After(backups[j].FullSnapshot.CreatedOn)
+	})
+
+	list := brtypes.SnapList{}
+	count := 0
+	for _, backup := range backups {
+		if maxBackups >= 0 && count == maxBackups {
+			break
+		}
+		if filter != nil && !filter(*backup.FullSnapshot) {
+			continue
+		}
+		list = append(list, backup.FullSnapshot)
+		list = append(list, backup.DeltaSnapshotList...)
+		count++
+	}
+
+	return list, nil
+}
+
+func getStructuredBackupList(snapList brtypes.SnapList) []backup {
+	var (
+		backups    []backup
+		tempBackup = backup{}
+	)
+
+	for i := len(snapList) - 1; i >= 0; i-- {
+		if snapList[i].IsChunk {
+			continue
+		}
+		if snapList[i].Kind == brtypes.SnapshotKindFull {
+			tempBackup.FullSnapshot = snapList[i]
+			backups = append(backups, tempBackup)
+			tempBackup = backup{}
+			continue
+		}
+		tempBackup.DeltaSnapshotList = append(tempBackup.DeltaSnapshotList, snapList[i])
+	}
+	return backups
+}
+
 // StartEmbeddedEtcd starts the embedded etcd server.
 func StartEmbeddedEtcd(logger *logrus.Entry, ro *brtypes.RestoreOptions) (*embed.Etcd, error) {
 	cfg := embed.NewConfig()
--- a/pkg/miscellaneous/miscellaneous_test.go
+++ b/pkg/miscellaneous/miscellaneous_test.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package miscellaneous
+
+import (
+	"fmt"
+	"io"
+	"reflect"
+	"time"
+
+	"github.com/gardener/etcd-backup-restore/pkg/types"
+	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var (
+	snapList types.SnapList
+	ds       DummyStore
+)
+
+const (
+	generatedSnaps = 20
+)
+
+var _ = Describe("Filtering snapshots", func() {
+	BeforeEach(func() {
+		snapList = generateSnapshotList(generatedSnaps)
+		ds = NewDummyStore(snapList)
+
+	})
+	It("should return the whole snaplist", func() {
+		snaps, err := GetFilteredBackups(&ds, -1, nil)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(snaps)).To(Equal(len(snapList)))
+	})
+	It("should get the last 3 snapshots and its deltas", func() {
+		n := 3
+		snaps, err := GetFilteredBackups(&ds, n, nil)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(snaps)).To(Equal(n * 2))
+		expectedSnapID := 0
+		for i := 0; i < n; i++ {
+			if reflect.DeepEqual(snaps[i].Kind, types.SnapshotKindFull) {
+				Expect(snaps[i].SnapName).To(Equal(fmt.Sprintf("%s-%d", types.SnapshotKindFull, expectedSnapID)))
+				Expect(snaps[i+1].SnapName).To(Equal(fmt.Sprintf("%s-%d", types.SnapshotKindDelta, expectedSnapID)))
+				expectedSnapID++
+			}
+		}
+	})
+	It("should get the last backups created in the past 5 days", func() {
+		n := 5
+		backups, err := GetFilteredBackups(&ds, n, func(snap types.Snapshot) bool {
+			return snap.CreatedOn.After(time.Now().UTC().AddDate(0, 0, -n))
+		})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(backups)).To(Equal(n * 2))
+		for i := 0; i < n; i++ {
+			if reflect.DeepEqual(backups[i].Kind, types.SnapshotKindFull) {
+				backups[i].CreatedOn.After(time.Now().UTC().AddDate(0, 0, -n))
+			}
+		}
+	})
+})
+
+func generateSnapshotList(n int) types.SnapList {
+	snapList := types.SnapList{}
+	for i := 0; i < n; i++ {
+		fullSnap := &types.Snapshot{
+			SnapName:  fmt.Sprintf("%s-%d", types.SnapshotKindFull, i),
+			Kind:      types.SnapshotKindFull,
+			CreatedOn: time.Now().UTC().AddDate(0, 0, -i),
+		}
+		deltaSnap := &types.Snapshot{
+			SnapName:  fmt.Sprintf("%s-%d", types.SnapshotKindDelta, i),
+			Kind:      types.SnapshotKindDelta,
+			CreatedOn: time.Now().UTC().AddDate(0, 0, -i),
+		}
+		snapList = append(snapList, fullSnap, deltaSnap)
+	}
+	return snapList
+}
+
+type DummyStore struct {
+	SnapList types.SnapList
+}
+
+func NewDummyStore(snapList types.SnapList) DummyStore {
+	return DummyStore{SnapList: snapList}
+}
+
+func (ds *DummyStore) List() (types.SnapList, error) {
+	return ds.SnapList, nil
+}
+
+func (ds *DummyStore) Delete(s brtypes.Snapshot) error {
+	return nil
+}
+
+func (ds *DummyStore) Save(snap brtypes.Snapshot, rc io.ReadCloser) error {
+	return nil
+}
+
+func (ds *DummyStore) Fetch(snap brtypes.Snapshot) (io.ReadCloser, error) {
+	return nil, nil
+}
--- a/pkg/snapshot/copier/copier.go
+++ b/pkg/snapshot/copier/copier.go
@@ -1,0 +1,128 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package copier
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/gardener/etcd-backup-restore/pkg/miscellaneous"
+	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
+	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
+
+	"github.com/sirupsen/logrus"
+)
+
+// GetSourceAndDestinationStores returns the source and destination stores for the given source and destination and store configs.
+func GetSourceAndDestinationStores(sourceSnapStoreConfig *brtypes.SnapstoreConfig, destSnapStoreConfig *brtypes.SnapstoreConfig) (brtypes.SnapStore, brtypes.SnapStore, error) {
+	sourceSnapStore, err := snapstore.GetSnapstore(sourceSnapStoreConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not get source snapstore: %v", err)
+	}
+
+	destSnapStore, err := snapstore.GetSnapstore(destSnapStoreConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not get destination snapstore: %v", err)
+	}
+
+	return sourceSnapStore, destSnapStore, nil
+}
+
+// Copier can be used to copy backups from a source to a destination store.
+type Copier struct {
+	logger          *logrus.Entry
+	sourceSnapStore brtypes.SnapStore
+	destSnapStore   brtypes.SnapStore
+	maxBackups      int
+	maxBackupAge    int
+}
+
+// NewCopier creates a new copier.
+func NewCopier(sourceSnapStore brtypes.SnapStore, destSnapStore brtypes.SnapStore, logger *logrus.Entry, maxBackups int, maxBackupAge int) *Copier {
+	return &Copier{
+		logger:          logger.WithField("actor", "copier"),
+		sourceSnapStore: sourceSnapStore,
+		destSnapStore:   destSnapStore,
+		maxBackups:      maxBackups,
+		maxBackupAge:    maxBackupAge,
+	}
+}
+
+// Run executes the copy command.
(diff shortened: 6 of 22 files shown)


pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,42 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
...
+			backups = append(backups, backup)
+			backup = Backup{}
+			continue
+		}
+		backup.DeltaSnapshotList = append(backup.DeltaSnapshotList, snapList[i])
+	}
+
+	for _, backup := range backups {
+		sort.Sort(backup.DeltaSnapshotList)
+	}
```
amshuman-kr, line 95: The `List` call is expected to return the snapshots in sorted order. So, ideally, this is not required.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,42 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
 	return fullSnapshot, deltaSnapList, nil
 }
 
+// Backup contains a full snapshot and all of its delta snapshots.
+type Backup struct {
+	FullSnapshot      *brtypes.Snapshot
+	DeltaSnapshotList brtypes.SnapList
```
amshuman-kr, line 67: Is the distinction between full and delta snapshots important while copying? Could we not simply have a list of snapshots? We could still group them together in the `Backup` structure based on a full snapshot and its following delta snapshots. But by grouping them together in one list might simplify the rest of the copier logic. WDYT?

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,135 @@
...
+		for _, deltaSnapshot := range backup.DeltaSnapshotList {
+			if _, ok := snapshotsMap[deltaSnapshot.SnapName]; !ok {
+				c.logger.Infof("Copying delta snapshot %s...", deltaSnapshot.SnapName)
+				if err := c.copySnapshot(deltaSnapshot); err != nil {
+					return err
+				}
+			} else {
+				c.logger.Infof("Skipping delta snapshot %s as it already exists", deltaSnapshot.SnapName)
+			}
+		}
```
amshuman-kr, line 119: Please see the comment [above](https://github.com/gardener/etcd-backup-restore/pull/369/files#diff-d67a01c5aa3ff7bb12c7e9945be604ac31bd8a766f5962760de4d8223e639aedR66-R67).

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,61 @@
...
+		var err error
+		ds, ss, err = GetSourceAndDestinationStores(copierConfig, snapstoreConfig)
+		Expect(err).ToNot(HaveOccurred())
+		copier = NewCopier(ss, ds, logger)
+	})
+	It("should run the copy command succesfully", func() {
+		Expect(copier.Run()).ToNot(HaveOccurred())
+		_, err = os.Stat(targetSnapstoreDir)
+		Expect(err).ToNot(HaveOccurred())
+		os.Remove(targetSnapstoreDir)
```
amshuman-kr, line 59: Could you please add a check that `miscellaneous. GetLatestFullSnapshotAndDeltaSnapList()` returns the expected snapshots from the destination store after copy?

pkg/snapstore/abs_snapstore.go
```diff
@@ -131,7 +142,9 @@ func (a *ABSSnapStore) List() (brtypes.SnapList, error) {
 		// Process the blobs returned in this result segment
 		for _, blob := range listBlob.Segment.BlobItems {
 			if strings.Contains(blob.Name, backupVersionV1) || strings.Contains(blob.Name, backupVersionV2) {
-				s, err := ParseSnapshot(path.Join(prefix, blob.Name))
+				//the blob may contain the full path in its name including the prefix
+				blobName := strings.TrimPrefix(blob.Name, prefix)
+				s, err := ParseSnapshot(path.Join(prefix, blobName))
```
amshuman-kr, line 147: @abdasgupta Can you please check if this change is required for the changes for folder structure change also?

pkg/snapstore/gcs_snapstore.go
```diff
@@ -74,7 +87,7 @@ func NewGCSSnapStoreFromClient(bucket, prefix, tempDir string, maxParallelChunkU
 
 // Fetch should open reader for the snapshot file from store.
 func (s *GCSSnapStore) Fetch(snap brtypes.Snapshot) (io.ReadCloser, error) {
-	objectName := path.Join(snap.Prefix, snap.SnapDir, snap.SnapName)
+	objectName := path.Join(s.getPrefix(snap), snap.SnapDir, snap.SnapName)
```
amshuman-kr, line 90: @abdasgupta Can you please check if this change is required for the changes for folder structure change also?

amshuman-kr (commented):
Thanks @kris94 for the feature! Looks good overall. Just some minor comments below.

pkg/snapstore/abs_snapstore.go
abdasgupta: This is not required for folder structure change

pkg/snapstore/gcs_snapstore.go
abdasgupta: @kris94 Would you please explain that why do we need it here? Like under what situation we would have empty prefix in snapshot but a path in GCS Store?  Elsewhere,`List` function is actually making sure that every snapshot should have prefix field set.

pkg/snapstore/abs_snapstore.go
abdasgupta: @kris94 Why are we trimming of prefix and then joining again? Is it to get rid of any extra `/` ?

kris94: During the my tests with `v2` I have observed  that the `blobName` comes with the prefix concatenated to it. So if we have `--store-prefix=prefix` then the blobName would be `prefix/Name` and with that `Join` `prefix/prefix/Name`, which points to a snapshot that does not exists. I assumed that this was not the case with `v1` or at leas in some point in the past, so I just put a `TrimPrefix` for backward compatibility.

pkg/snapstore/gcs_snapstore.go
amshuman-kr: The rationale behind the question is, if there is a bug we would like to include a fix in the release for backup compaction that we are planning to cut this week.

pkg/miscellaneous/miscellaneous.go
kris94: I did a bit of refactoring based on your feedback and indeed it looks better that way. Functionally I don't find find differences.

amshuman-kr: Yes, I didn't mean to have any change in functionality. Just to simplify the logic. That is all. Thanks for taking up the changes.

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,67 @@
...
+		var err error
+		ds, ss, err = GetSourceAndDestinationStores(copierConfig, snapstoreConfig)
+		Expect(err).ToNot(HaveOccurred())
+		copier = NewCopier(ss, ds, logger)
+	})
+	It("should run the copy command succesfully", func() {
+		fullSourceStoreSnapshot, deltaSrourceStoreSnapList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ss)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(copier.Run()).ToNot(HaveOccurred())
+		fullTargetStoreSnapshot, deltaTargetStoreSnapList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ss)
```
amshuman-kr, line 60:
```suggestion
		fullTargetStoreSnapshot, deltaTargetStoreSnapList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ds)
```

Should the listing here be from the destination snapstore?

amshuman-kr (dismissed):
Thanks for taking up the changes @kris94 ! One small change and it looks good.

pkg/snapshot/copier/copier_test.go
kris94: Ah yes, just a copy paste mistake :/ 
Also this won't work, because the `Prefix` is different for `source` and `target`  so I will implement something else.

pkg/snapstore/utils.go
```diff
@@ -38,7 +40,15 @@ func GetSnapstore(config *brtypes.SnapstoreConfig) (brtypes.SnapStore, error) {
...
 	if config.Container == "" {
-		config.Container = os.Getenv(envStorageContainer)
+		if config.IsSource {
+			config.Container = os.Getenv(sourceEnvStorageContainer)
+		} else {
+			config.Container = os.Getenv(envStorageContainer)
+		}
+	}
+
+	if config.Container == "" && (config.Provider != "" && config.Provider != brtypes.SnapstoreProviderLocal) {
```
plkokanov, line 50:
```suggestion
	if config.Container == "" && config.Provider != "" && config.Provider != brtypes.SnapstoreProviderLocal {
```

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,99 @@
...
+			Provider:                "Local",
+			Container:               snapstoreDir,
+		}
+
+		var err error
+		ds, ss, err = GetSourceAndDestinationStores(copierConfig, snapstoreConfig)
+		Expect(err).ToNot(HaveOccurred())
+		copier = NewCopier(ss, ds, logger)
+	})
+	It("should run the copy command succesfully", func() {
```
plkokanov, line 57: Perhaps we should also add a test that checks whether all snapshots are copied, as well as a test that checks if a snapshot already exists in the destination snapstore it is not copied.

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,99 @@
...
+		fullSourceStoreSnapshot, deltaSrourceStoreSnapList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ss)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(copier.Run()).ToNot(HaveOccurred())
+		fullTargetStoreSnapshot, deltaTargetStoreSnapList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ds)
+		Expect(snapsAreTheSame(fullSourceStoreSnapshot, fullTargetStoreSnapshot)).To(BeTrue())
+		for i, ssnap := range deltaSrourceStoreSnapList {
+			Expect(len(deltaSrourceStoreSnapList)).To(Equal(len(deltaTargetStoreSnapList)))
```
plkokanov, line 64: This line should be before the for loop.

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,99 @@
...
+			Expect(snapsAreTheSame(ssnap, deltaTargetStoreSnapList[i])).To(BeTrue())
+		}
+		_, err = os.Stat(targetSnapstoreDir)
+		Expect(err).ToNot(HaveOccurred())
+		os.Remove(targetSnapstoreDir)
```
plkokanov, line 69: Perhaps I am missing something, but should't these lines be part of an AfterEach call. Also if the file does not exist, the tests will fail.

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+})
+
+func snapsAreTheSame(s1 *brtypes.Snapshot, s2 *brtypes.Snapshot) bool {
+	//SnapName and Prefix could be different after the copy operation.
+	logger.Logger.Infof("%v %v", s1, s2)
```
plkokanov, line 75: Add more information to the log message.

pkg/snapshot/copier/copier_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if !reflect.DeepEqual(s1.Kind, s2.Kind) {
```
plkokanov, line 77: The problem with calling DeepEqual like this and just returning false is that we will not know what element caused the tests to fail. Perhaps you should change this method to instead call
`Expect(s1.Kind).To(Equal(s2.Kind))` and so on.

plkokanov (changes requested):

pkg/snapshot/copier/copier_test.go
amshuman-kr: Maybe [gomega gstruct](https://onsi.github.io/gomega/#gstruct-testing-complex-data-types)'s [MatchFields](https://pkg.go.dev/github.com/onsi/gomega/gstruct#MatchFields) is better suited in place of the `snapsAreTheSame` func.

pkg/snapshot/copier/copier_test.go
kris94: Its a leftover it is already called [here](https://github.com/gardener/etcd-backup-restore/blob/512a4c986f68c0f42b54e390ff50a728c8abc260/pkg/snapshot/copier/copier_suite_test.go#L100)

amshuman-kr:
Now that the [v0.13.0](https://github.com/gardener/etcd-backup-restore/releases/tag/v0.13.0) release has been cut, can we merge this?

cc @ishan16696 @shreyas-s-rao @aaronfern @abdasgupta

cmd/copy.go
```diff
@@ -0,0 +1,58 @@
...
+				return
+			}
+			opts.complete()
+
+			destStorage, sourceStorage, err := copier.GetSourceAndDestinationStores(opts.sourceSnapStoreConfig, opts.snapstoreConfig)
+			if err != nil {
+				logger.Fatalf("Could not get source and destination snapstores: %v", err)
+			}
+
+			copier := copier.NewCopier(destStorage, sourceStorage, logger)
```
plkokanov, line 46:
```suggestion
			copier := copier.NewCopier(sourceStorage, destStorage, logger)
```

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,114 @@
...
+// Run executes the copy command.
+func (c *Copier) Run() error {
+	return c.CopyBackups()
+}
+
+// CopyBackups copies all backups from the source store to the destination store.
+func (c *Copier) CopyBackups() error {
+	// Get source backups
+	c.logger.Info("Getting source backups...")
+	sourceBackups, err := miscellaneous.GetAllBackups(c.sourceSnapStore)
```
plkokanov, line 67: Small nit, but we should probably unify the names between `sourceBackups` and the destination `snapshots` on line 80

pkg/types/snapstore.go
```diff
@@ -169,3 +180,17 @@ func (c *SnapstoreConfig) Validate() error {
...
+	if c.Provider == "" {
+		c.Provider = other.Provider
+	}
+	if c.Prefix == "" {
+		c.Prefix = other.Prefix
+	} else {
+		c.Prefix = path.Join(c.Prefix, backupFormatVersion)
+	}
+	c.MaxParallelChunkUploads = other.MaxParallelChunkUploads
+	c.TempDir = other.TempDir
```
plkokanov, line 195: Now that the source store also has `MaxParallelChunkUploads` and `TempDir` these should also be checked if they exist and only then defaulted to `other.MaxParallelChunkUploads` and `other.TempDir`

plkokanov (changes requested):
Just a couple of nits.

pkg/snapstore/abs_snapstore.go
```diff
@@ -48,12 +48,12 @@ type ABSSnapStore struct {
...
 // NewABSSnapStore create new ABSSnapStore from shared configuration with specified bucket
-func NewABSSnapStore(container, prefix, tempDir string, maxParallelChunkUploads uint) (*ABSSnapStore, error) {
-	storageAccount, err := GetEnvVarOrError(absStorageAccount)
-	if err != nil {
-		return nil, err
+func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
+	prefixString := ""
+	if config.IsSource {
+		prefixString = sourcePrefixString
 	}
```
stoyanr, line 55: This calculation of the prefix string is repeated several times. I would extract it in an utility function (in `utils.go`), e.g. `getPrefixString(config.IsSource)` and use it directly instead of assigning it to a variable, e.g. `getCredentials(getPrefixString(config.IsSource))`

pkg/snapstore/s3_snapstore.go
```diff
@@ -71,6 +75,22 @@ func newS3FromSessionOpt(bucket, prefix, tempDir string, maxParallelChunkUploads
 	return NewS3FromClient(bucket, prefix, tempDir, maxParallelChunkUploads, cli), nil
 }
 
+func getSessionOptions(prefixString string) session.Options {
+	if _, isSet := os.LookupEnv(prefixString + awsAcessKeyID); isSet {
+		return session.Options{
+			Config: aws.Config{
+				Credentials: credentials.NewStaticCredentials(os.Getenv(prefixString+awsAcessKeyID), os.Getenv(prefixString+awsSecretAccessKey), ""),
+				Region:      pointer.StringPtr(os.Getenv("AWS_REGION")),
```
stoyanr, line 83: I find it strange that other `AWS_` env variable names are now extracted as constants, but `AWS_REGION` is not, so let's please extract it as well.

pkg/snapstore/swift_snapstore.go
```diff
@@ -52,8 +56,14 @@ const (
...
 // NewSwiftSnapStore create new SwiftSnapStore from shared configuration with specified bucket
-func NewSwiftSnapStore(bucket, prefix, tempDir string, maxParallelChunkUploads uint) (*SwiftSnapStore, error) {
-	authOpts, err := clientconfig.AuthOptions(nil)
+func NewSwiftSnapStore(config *brtypes.SnapstoreConfig) (*SwiftSnapStore, error) {
+	var clientOpts *clientconfig.ClientOpts
+	if config.IsSource {
+		clientOpts = &clientconfig.ClientOpts{
+			EnvPrefix: envPrefixSource,
+		}
+	}
```
stoyanr, line 65: Also here I would extract this in a helper function, even though it's used only once, e.g. `getClientOpts(config.IsSource)`

pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,23 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
...
+	snapshots := brtypes.SnapList{}
+	for i := len(snapList) - 1; i >= 0; i-- {
+		if !snapList[i].IsChunk {
+			snapshots = append(snapshots, snapList[i])
+		}
+	}
+
+	return snapshots, nil
+}
+
```
stoyanr, line 80: I would propose to implement this function as it's currently implemented in the "bad case" PoC, see https://github.com/stoyanr/etcd-backup-restore/commit/0037b37363176cee0840a5a06d99ea5022970608. The new implementation guarantees that the snapshots are grouped and sorted correctly. With it, it would be easier to filter only snapshots (full + incremental) that are within a certain time range, etc.

stoyanr (changes requested):
Besides the inline comments, I would propose to add additional configuration parameters to the copier so that it could:
* copy only the latest full snapshot and its incremental snapshots (or the latest n full snapshots and their incremental snapshots)
* copy only the full snapshots and their incremental snapshots that are newer than a certain configured number of days (e.g. 14 days)

pkg/miscellaneous/miscellaneous.go
kris94: This was initially implemented that way, but then we decided to refactor it (you can check one of the comments above). But you might be right that it could be easier to filter that way, it was just not the case before you suggested that change.

cmd/options.go
```diff
@@ -254,11 +254,54 @@ func (c *snapshotterOptions) validate() error {
...
+	sourceSnapStoreConfig.IsSource = true
+	return &copierOptions{
+		sourceSnapStoreConfig: sourceSnapStoreConfig,
+		snapstoreConfig:       snapstore.NewSnapstoreConfig(),
+	}
+}
+
+func (c *copierOptions) addFlags(fs *flag.FlagSet) {
+	fs.IntVar(&c.maxSnapshots, "last-n-snapshots", -1, "copy the last N snapshots form the store sorted by date from newest to oldest")
+	fs.IntVar(&c.maxSnapshotAge, "max-age-days", -1, "copy the last snapshots form the store not older than N days")
```
stoyanr, line 283: Let's think about a better and more consistent naming of these 2 parameters. My first proposal would be to name them after the variables, I like the naming of variables much better here, so `max-snapshots`, `max-snapshot-age`. @plkokanov any other suggestions? In any case, the naming of flags and their variable should be the same, the variables should match the flags as closely as possible.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,100 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
...
+// GetFilteredBackups returns backups (full snapshots and all their delta snapshots) based on the filter function.
+func GetFilteredBackups(store brtypes.SnapStore, filter func(snaps brtypes.SnapList) (brtypes.SnapList, error)) (brtypes.SnapList, error) {
```
stoyanr, line 82: I think there are too many functions that perform the same task with slight variations. I would combine all functions starting with `GetAllBackups` and ending with `FilterLastSnapshotsAndItsDeltasSinceNDays` with a single function `GetFilteredSnapshots` that accepts a store, a max number of backups (0 for all), and a "predicate" function to filter a single full snapshot (nil for no effect), and that performs all fetching, sorting, grouping, and filtering.
It should return `(brtypes.SnapList, error)` and therefore the notion of `backup` would be internal to it and not leak into any external names.

pkg/snapstore/utils.go
```diff
@@ -163,3 +154,18 @@ func collectChunkUploadError(chunkUploadCh chan<- chunk, resCh <-chan chunkUploa
...
 }
+
+func getEnvPrefixString(isSource bool) string {
+	if isSource {
+		return sourcePrefixString
+	}
+	return ""
+}
+
+func resolveBackupPrefix(snap *brtypes.Snapshot, snapstorePrefix string) string {
```
stoyanr, line 165: Nit: A name such as `getPrefix` or `adaptPrefix` is more appropriate here IMO.

stoyanr (changes requested):
Thanks for addressing my proposal, I still have a few things that I think could be improved.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,59 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
...
 }
 
+type backup struct {
+	FullSnapshot      *brtypes.Snapshot
+	DeltaSnapshotList brtypes.SnapList
+}
+
+// GetFilteredSnapshots returns sorted by date (new -> old) SnapList. It will also filter the snapshots that should be included or not useing the filter funcion.
+// If the filter is nil it will return all snapshots. Also, maxNumberOfSnapshots can be used to target only the last N snapshots (0 = all).
+func GetFilteredSnapshots(store brtypes.SnapStore, maxNumberOfSnapshots int, filter func(snaps brtypes.Snapshot) bool) (brtypes.SnapList, error) {
```
stoyanr, line 71: I am thinking that perhaps we should talk about "backups" and not "snapshots" here, see also my other comments.
```suggestion
func GetFilteredBackups(store brtypes.SnapStore, maxBackups int, filter func(snaps brtypes.Snapshot) bool) (brtypes.SnapList, error) {
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,59 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
 	return fullSnapshot, deltaSnapList, nil
 }
 
+type backup struct {
+	FullSnapshot      *brtypes.Snapshot
+	DeltaSnapshotList brtypes.SnapList
+}
+
+// GetFilteredSnapshots returns sorted by date (new -> old) SnapList. It will also filter the snapshots that should be included or not useing the filter funcion.
+// If the filter is nil it will return all snapshots. Also, maxNumberOfSnapshots can be used to target only the last N snapshots (0 = all).
```
stoyanr, line 70:
```suggestion
// GetFilteredBackups returns a SnapList containing at most maxBackups for which the filter function returns true as a SnapList, sorted by date (new -> old).
```

cmd/options.go
```diff
@@ -254,11 +253,51 @@ func (c *snapshotterOptions) validate() error {
...
+	sourceSnapStoreConfig.IsSource = true
+	return &copierOptions{
+		sourceSnapStoreConfig: sourceSnapStoreConfig,
+		snapstoreConfig:       snapstore.NewSnapstoreConfig(),
+	}
+}
+
+func (c *copierOptions) addFlags(fs *flag.FlagSet) {
+	fs.IntVar(&c.maxSnapshots, "max-snapshots", -1, "copy the last N snapshots form the store sorted by date from newest to oldest")
+	fs.IntVar(&c.maxSnapshotAge, "max-snapshot-age", -1, "copy the last snapshots form the store not older than N days")
```
stoyanr, line 282: Even though I proposed (or approved) this naming in my previous review, now I believe it would be more correct to talk about "backups" and not snapshots here, a backup being the collection of a single full snapshot and all incremental snapshots based on it.

```suggestion
	fs.IntVar(&c.maxBackups, "max-backups", -1, "copy the specified number of backups sorted by date from newest to oldest")
	fs.IntVar(&c.maxBackupAge, "max-backup-age", -1, "copy only the backups not older than the specified number of days")
```

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+				return err
+			}
+		} else {
+			c.logger.Infof("Skipping %s snapshot %s as it already exists", backup.Kind, backup.SnapName)
+		}
+	}
+	return nil
+}
+
+func (c *Copier) getBackups() (brtypes.SnapList, error) {
```
stoyanr, line 108:
```suggestion
func (c *Copier) getSnapshots() (brtypes.SnapList, error) {
```

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+	return sourceSnapStore, destSnapStore, nil
+}
+
+// Copier can be used to copy backups from a source to a destination store.
+type Copier struct {
+	logger           *logrus.Entry
+	sourceSnapStore  brtypes.SnapStore
+	destSnapStore    brtypes.SnapStore
+	maxSnapshots     int
+	lastSnapshotDays int
```
stoyanr, line 49:
```suggestion
	maxBackups     int
	maxBackupAge int
```

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+// Run executes the copy command.
+func (c *Copier) Run() error {
+	return c.CopyBackups()
+}
+
+// CopyBackups copies all backups from the source store to the destination store.
+func (c *Copier) CopyBackups() error {
+	// Get source backups
+	c.logger.Info("Getting source backups...")
+	sourceBackups, err := c.getBackups()
```
stoyanr, line 72:
```suggestion
	sourceSnapshot, err := c.getSnapshots()
```

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		} else {
+			c.logger.Infof("Skipping %s snapshot %s as it already exists", backup.Kind, backup.SnapName)
+		}
+	}
+	return nil
+}
+
+func (c *Copier) getBackups() (brtypes.SnapList, error) {
+	if c.lastSnapshotDays > 0 {
```
stoyanr, line 109: The default (meaning all) is -1, not 0.
```suggestion
	if c.maxBackupAge >= 0 {
```

stoyanr (changes requested):
I think we should change the naming to avoid the confusion between snapshots and backups.

cmd/options.go
```diff
@@ -254,11 +253,51 @@ func (c *snapshotterOptions) validate() error {
...
+	return c.sourceSnapStoreConfig.Validate()
+}
+
+func (c *copierOptions) complete() {
+	if c.maxSnapshots <= 0 {
+		c.maxSnapshots = -1
+	}
+	if c.maxSnapshotAge <= 0 {
+		c.maxSnapshotAge = -1
+	}
```
stoyanr, line 300: Instead of this, we should validate `maxSnapshots` and `maxSnapshotAge` for being >= -1 in `validate`.

doc/usage/getting_started.md
```diff
@@ -144,3 +144,35 @@ INFO[0008] Successfully restored the etcd data directory.
...
+```console
+$ ./bin/etcdbrctl copy \
+--storage-provider="GCS" \
+--snapstore-temp-directory="/temp" \
+--store-prefix="target-prefix" \
+--store-container="target-container" \
+--source-store-prefix="prefix" \
+--source-store-container="container" \
+--source-storage-provider="GCS" \
+--max-age-days=15 \
```
stoyanr, line 161: Please adapt this as well (to whatever it's eventually called)
```suggestion
--max-snapshot-age=15 \
```

stoyanr (changes requested):
A few more comments.

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+	destSnapshots, err := c.destSnapStore.List()
+	if err != nil {
+		return fmt.Errorf("could not get destination snapshots: %v", err)
+	}
+	destSnapshotsMap := make(map[string]*brtypes.Snapshot)
+	for _, snapshot := range destSnapshots {
+		destSnapshotsMap[snapshot.SnapName] = snapshot
+	}
+
+	for _, smapshot := range sourceSnapshot {
```
plkokanov, line 94:
```suggestion
	for _, snapshot := range sourceSnapshot {
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -61,6 +61,61 @@ func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Sn
 	return fullSnapshot, deltaSnapList, nil
 }
 
+type backup struct {
+	FullSnapshot      *brtypes.Snapshot
+	DeltaSnapshotList brtypes.SnapList
+}
+
+// GetFilteredBackups returns sorted by date (new -> old) SnapList. It will also filter the snapshots that should be included or not useing the filter funcion.
+// If the filter is nil it will return all snapshots. Also, maxNumberOfSnapshots can be used to target only the last N snapshots (0 = all).
```
plkokanov, line 70:
```suggestion
// If the filter is nil it will return all snapshots. Also, maxBackups can be used to target only the last N snapshots (0 = all).
```

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+type Copier struct {
+	logger          *logrus.Entry
+	sourceSnapStore brtypes.SnapStore
+	destSnapStore   brtypes.SnapStore
+	maxBackups      int
+	maxBackupAge    int
+}
+
+// NewCopier creates a new copier.
+func NewCopier(sourceSnapStore brtypes.SnapStore, destSnapStore brtypes.SnapStore, logger *logrus.Entry, maxSnapshots int, lastSnapshotDays int) *Copier {
```
plkokanov, line 53:
```suggestion
func NewCopier(sourceSnapStore brtypes.SnapStore, destSnapStore brtypes.SnapStore, logger *logrus.Entry, maxBackups int, maxBackupAge int) *Copier {
```

doc/usage/getting_started.md
```diff
@@ -144,3 +144,35 @@ INFO[0008] Successfully restored the etcd data directory.
 ### Etcdbrctl server
 
 With sub-command `server` you can start a http server which exposes an endpoint to initialize etcd over REST interface. The server also keeps the backup schedule thread running to keep taking periodic backups. This is mainly made available to manage an etcd instance running in a Kubernetes cluster. You can deploy the example [helm chart](../../chart/etcd-backup-restore) on a Kubernetes cluster to have a fault-resilient, self-healing etcd cluster.
+
+## Etcdbrctl copy
+
+With sub-command `copy` you can copy all snapshots (Full and Delta) fom one snapstore to another. Using the two filter parameters `max-snapshots` and `max-snapshot-age` you can also limit the number of snapshots that will be copied or target only the newest snapshots.
```
plkokanov, line 150:
```suggestion
With sub-command `copy` you can copy all snapshots (Full and Delta) fom one snapstore to another. Using the two filter parameters `max-backups` and `max-backup-age` you can also limit the number of snapshots that will be copied or target only the newest snapshots.
```

pkg/miscellaneous/miscellaneous_test.go
```diff
@@ -0,0 +1,118 @@
...
+		expectedSnapID := 0
+		for i := 0; i < n; i++ {
+			if reflect.DeepEqual(snaps[i].Kind, types.SnapshotKindFull) {
+				Expect(snaps[i].SnapName).To(Equal(fmt.Sprintf("%s-%d", types.SnapshotKindFull, expectedSnapID)))
+				Expect(snaps[i+1].SnapName).To(Equal(fmt.Sprintf("%s-%d", types.SnapshotKindDelta, expectedSnapID)))
+				expectedSnapID++
+			}
+		}
+	})
+	It("should get the last snapshots created in the past 5 dys", func() {
```
plkokanov, line 63:
```suggestion
	It("should get the last backups created in the past 5 days", func() {
```

tbh not 100% if this should be renamed

plkokanov (changes requested):
Found a few more backup/snapshot naming missmatches

stoyanr (approved):
/lgtm

pkg/snapshot/copier/copier.go
```diff
@@ -0,0 +1,128 @@
...
+	destSnapshotsMap := make(map[string]*brtypes.Snapshot)
+	for _, snapshot := range destSnapshots {
+		destSnapshotsMap[snapshot.SnapName] = snapshot
+	}
+
+	for _, snapshot := range sourceSnapshot {
+		// Copy all the snapshots (if not already copied)
+		if _, ok := destSnapshotsMap[snapshot.SnapName]; !ok {
+			c.logger.Infof("Copying %s snapshot %s...", snapshot.Kind, snapshot.SnapName)
+			if err := c.copySnapshot(snapshot); err != nil {
```
timuthy, line 98: We should think about parallelization in a future PR.

pkg/types/snapstore.go
```diff
@@ -169,3 +180,18 @@ func (c *SnapstoreConfig) Validate() error {
 func (c *SnapstoreConfig) Complete() {
 	c.Prefix = path.Join(c.Prefix, backupFormatVersion)
 }
+
+// CompleteWithOther completes the config based on other config
+func (c *SnapstoreConfig) CompleteWithOther(other *SnapstoreConfig) {
```
timuthy, line 185:
```suggestion
func (c *SnapstoreConfig) MergeWith(config *SnapstoreConfig) {
```

timuthy (commented):
Just a few minor comments from my side, the rest looks good. Thank you.

cmd/copy.go
```diff
@@ -0,0 +1,58 @@
...
+			logger := logrus.NewEntry(logrus.New())
+			if err := opts.validate(); err != nil {
+				logger.Fatalf("failed to validate the options: %v", err)
+				return
+			}
+			opts.complete()
+
+			sourceStorage, destStorage, err := copier.GetSourceAndDestinationStores(opts.sourceSnapStoreConfig, opts.snapstoreConfig)
+			if err != nil {
+				logger.Fatalf("Could not get source and destination snapstores: %v", err)
```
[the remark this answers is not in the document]
stoyanr: `logger.Fatalf` exits the process anyway, so rather the other `return` statement after `logger.Fatalf` should be removed.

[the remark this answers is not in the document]
kris94: Fatal will call os.Exit(1). I will remove the other returns since they are not necessary.

pkg/snapshot/copier/copier.go
stoyanr: Do you think this could increase the performance? I think the bottleneck is likely to be the read / write throughput of the backup container, and if this is the case then copying multiple snapshots in parallel is unlikely to be faster. But of course this is just an unverified assumption that should be checked.

timuthy: I'm not sure either but maybe worth a try :)

timuthy (approved):
/lgtm

PR #301, opened 2021-01-05
Added support for backup compaction

abdasgupta:
**What this PR does / why we need it**:
So far, compaction was done manually on the ETCD instance running in seed cluster. But with this PR, compaction is introduced as a command to the ETCDBR. At this point, ETCDBR will be able to restore-defragment-compact-snapshot without any manual intervention in between.

**Which issue(s) this PR fixes**:
Fixes https://github.com/gardener/etcd-druid/issues/88

**Release note**:

```noteworthy operator
Added compaction command support to ETCD Backup Restore. Users can now use compaction sub command with `etcdbrctl` tool to compact basesnapshot and all it's subsequent deltasnapshot to one single compacted snapshot.
```
```noteworthy operator
Snapshots are not stored under separate prefix (Backup-XXX) in backup storage. Instead, all of the snapshots are stored under `v2` backup version prefix in a flat structure.
```
```noteworthy operator
Support for `v1` backup version is still there but only for restoration and snapshot garbage collection. So backup storages that had `v1` backup version and where snapshots were stored  under separate prefix (Backup-XXX), can still be used with compaction sub command (as well as for restoration). However, the new snapshots will be stored under `v2` backup version prefix.
```
```noteworthy operator
The support for the `v1` backup structure with separate prefix (Backup-XXX) will be dropped in a subsequent release.
```

PR #371, opened 2021-08-11
Handled properly potential of wrong prefix formation in Azure.

abdasgupta:
This a PR fixes a potential bug that may create wrong prefix in Azure backup storage for ETCD.

With v2 directory structure introduced with Compaction [PR](https://github.com/gardener/etcd-backup-restore/pull/301), it was observed that the blobName comes with the prefix concatenated to it. So if store prefix is set with `--store-prefix=prefix` then the blobName would be prefix/Name and with that Join prefix/prefix/Name, which points to a snapshot that does not exists. This PR resolves this bug.

Fixes #

**Release note**:

```improvement operator

```


amshuman-kr (approved):
Thanks for the fix @abdasgupta @kris94! LGTM

shreyas-s-rao (approved):
@abdasgupta Thanks for the fix! Please rebase the PR on latest master.

PR #373, opened 2021-08-14
Fixed a bug that was creating a different directory for delta snapshots.

abdasgupta:
There is a bug in compaction PR. When you upgrade the garden cluster from older ETCDBR image to new ETCD image that has compaction PR and folder structure change, delta snapshots are still saved in older directories. This PR fixed the bug so that delta snapshots are saved only under `v2` prefix.
Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -388,7 +388,6 @@ func (ssr *Snapshotter) TakeDeltaSnapshot() (*brtypes.Snapshot, error) {
 		return nil, fmt.Errorf("failed to get compressionSuffix: %v", err)
 	}
 	snap := snapstore.NewSnapshot(brtypes.SnapshotKindDelta, ssr.prevSnapshot.LastRevision+1, ssr.lastEventRevision, compressionSuffix)
-	snap.SnapDir = ssr.prevSnapshot.SnapDir
 
 	// compute hash
 	hash := sha256.New()


pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -388,7 +388,7 @@ func (ssr *Snapshotter) TakeDeltaSnapshot() (*brtypes.Snapshot, error) {
 		return nil, fmt.Errorf("failed to get compressionSuffix: %v", err)
 	}
 	snap := snapstore.NewSnapshot(brtypes.SnapshotKindDelta, ssr.prevSnapshot.LastRevision+1, ssr.lastEventRevision, compressionSuffix)
-	snap.SnapDir = ssr.prevSnapshot.SnapDir
+	snap.SnapDir = ""
```
amshuman-kr, line 391:
```suggestion
```
`SnapDir` is already `""` as returned by `NewSnapshot`.

amshuman-kr (changes requested):
Thanks for the fix @abdasgupta. Just a small comment below.

pkg/snapshot/snapshotter/snapshotter.go
abdasgupta: Right, I knew. But I tried to be extra sure! in case, some code overwrite it (Though I can't imagine what can cause it). Are you sure we should remove the line?

amshuman-kr: If such a possibility is concrete then we should have a test case. But if we cannot yet think of a way how it might happen, let's not add addition code to handle it.

amshuman-kr (approved):
Thanks for the change @abdasgupta! LGTM

ishan16696:
I have tested the cluster-based Integration test locally. I have attached the log file.
[clusterBasedIntegrationTest.log](https://github.com/gardener/etcd-backup-restore/files/6997613/clusterBasedIntegrationTest.log)

ishan16696 (approved):
LGTM!

PR #374, opened 2021-08-17
Mock ETCD clientv3 for unit tests.

amshuman-kr:
**What this PR does / why we need it**:

Generated mock client for ETCD clientv3 to be used in unit tests for multi-node scenarions like #321, #322 etc.

Some of the specific client interfaces (like `Cluster`, `Maintenance` etc.) do not support the [`Close` function](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/cluster.go#L35-L53).
However, the [connection needs to be closed](https://github.com/etcd-io/etcd/blob/main/client/v3/client.go#L152) (say, by [closing the `Client` object](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/client.go#L143-L155)).

To simplify creation and closing the specific ETCD client objects without referring to the `Client` type directly, I have introduced new interfaces (like `ClusterCloser`, `MaintenanceCloser` etc.) which supports both the corresponding ETCD client interface as well as `io.Closer`.

And to put it all together, I have introduced the `Factory` interface which allows direct construction of the specific ETCD client interfaces allowing the use of _proper_ ETCD clients based on `EtcdConnectionConfig` as well as mock clients in unit tests.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
To be merged after the [release v0.13.0](https://github.com/gardener/etcd-backup-restore/milestone/34) has been made.

**Release note**:

```improvement operator
NONE
```

PR #380, opened 2021-09-09
Moved client factory implementation to etcdutil to avoid potential cyclic imports later.

amshuman-kr:

[Some](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/etcdutil/etcdutil.go#L91) [functions](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/etcdutil/etcdutil.go#L122) in the `etcdutil` package will have to start using the `etcdutil/client`'s `Factory` interface introduced in #374. This might lead to circular imports because `etcdutil/client` already imports `etcdutil` for the `factoryImpl` struct.

This PR moves the `factoryImpl` implementation to the `etcdutil` package so that `etcdutil/client` no longer needs to import `etcdutil`, thus avoiding the potential circular imports.

Fixes #

Credits: @ishan16696 for pointing out this problem.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/etcdutil/client/types.go
+++ b/pkg/etcdutil/client/types.go
@@ -17,7 +17,6 @@ package client
 import (
 	"io"
 
-	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
 	"go.etcd.io/etcd/clientv3"
 )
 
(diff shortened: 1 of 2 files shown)


ishan16696 (commented):
Do we need to generate mock [func](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/mock/etcdutil/client/doc.go#L14) again as these [mock func](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/mock/etcdutil/client/mocks.go#L56-L98) are still in pkg `mock/etcdutil/client`  or Is this fine as `mock/etcdutil/client` won't cause any cyclic dependency while implementing unit tests?

amshuman-kr:
> Do we need to generate mock [func](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/mock/etcdutil/client/doc.go#L14) again as these [mock func](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/mock/etcdutil/client/mocks.go#L56-L98) are still in pkg `mock/etcdutil/client` or Is this fine as `mock/etcdutil/client` won't cause any cyclic dependency while implementing unit tests?

The actual interfaces (`Facrory`, `ClusterCloser` etc.) are still in `etcdutil/client`. Hence, the generated mocks can be in `mock/etcdutil/client`. I have only moved the `factoryImpl` to `etcdutil` to avoid the cyclic imports.

ishan16696 (approved):
Thanks @amshuman-kr to clarify. LGTM!!

PR #381, opened 2021-09-14
To prepare the backup-restore to use Gomock: replaced the etcd client with etcd Maintenance/KV/Cluster/Watch client.

### LINKED ISSUE #378: [Feature]To prepare the backup-restore to use Gomock
To avoid using the [Client struct](https://github.com/etcd-io/etcd/blob/main/client/v3/client.go#L44-L50) directly and want to start using the [Maintenance](https://github.com/etcd-io/etcd/blob/main/client/v3/maintenance.go#L38-L85), [Watcher](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/watch.go#L46-L82), [KV](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/kv.go#L33-L65), [Cluster](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/cluster.go#L35-L53) Interface object according to functionality needed.  

Till now, In `backup-restore` we are using the [Client](https://github.com/etcd-io/etcd/blob/main/client/v3/client.go#L44-L50) struct directly by getting the [clientv3 obj](https://github.com/gardener/etcd-backup-restore/blob/8de2827fc50132f221aee64bdb10f41d2d375224/pkg/etcdutil/etcdutil.go#L37-L88) but Now we want to migrate our unit test to use [gomock](https://pkg.go.dev/github.com/golang/mock/gomock) as gomock don't mock the struct.

ishan16696:
Replaced the [etcd client struct object](https://github.com/etcd-io/etcd/blob/main/client/v3/client.go#L44-L50) from directly using it and to start using the [Maintenance](https://github.com/etcd-io/etcd/blob/main/client/v3/maintenance.go#L38-L85), [Watcher](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/watch.go#L46-L82), [KV](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/kv.go#L33-L65), [Cluster](https://github.com/etcd-io/etcd/blob/a4a82cc98290139f95ae8d2566bdc09948123bdb/client/v3/cluster.go#L35-L53) Interface object according to the functionality needed.

To prepare the backup-restore to use [gomock](https://pkg.go.dev/github.com/golang/mock/gomock) for unit testing.

Fixes #378 

**Release note**:
```improvement operator
NONE
```


ishan16696:
It also passed  the `Cluster-Based Integration Tests`. I have attached the log file.
[clusterBasedIntegrationTest_PR_client.log](https://github.com/gardener/etcd-backup-restore/files/7159577/clusterBasedIntegrationTest_PR_client.log)

pkg/compactor/compactor.go
```diff
@@ -99,32 +98,42 @@ func (cp *Compactor) Compact(opts *brtypes.CompactOptions) (*brtypes.Snapshot, e
...
+		return nil, fmt.Errorf("failed to build etcd KV client")
 	}
-	defer client.Close()
+	defer clientKV.Close()
+
+	clientMaintenance, err := clientFactory.NewMaintenance()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build etcd maintenance client")
+	}
+	defer clientMaintenance.Close()
```
amshuman-kr, line 116: This is used only for defragmentation. Can you please move it inside the `if opts.NeedDefragmentation {}` block?

pkg/etcdutil/etcdutil.go
```diff
@@ -149,7 +157,15 @@ func DefragmentData(defragCtx context.Context, client *clientv3.Client, endpoint
 }
 
 // TakeAndSaveFullSnapshot takes full snapshot and save it to store
-func TakeAndSaveFullSnapshot(ctx context.Context, client *clientv3.Client, store brtypes.SnapStore, lastRevision int64, cc *compressor.CompressionConfig, suffix string, logger *logrus.Entry) (*brtypes.Snapshot, error) {
+func TakeAndSaveFullSnapshot(ctx context.Context, clientFactory client.Factory, store brtypes.SnapStore, lastRevision int64, cc *compressor.CompressionConfig, suffix string, logger *logrus.Entry) (*brtypes.Snapshot, error) {
```
amshuman-kr, line 160: Why not pass `client.MaintenanceCloser` here instead of `client.Factory` here? That would be more consistent with other places in the PR. WDYT?

amshuman-kr (commented):
Thanks for the change @ishan16696. It looks good. Just a couple of minor comments/suggestions below.

pkg/compactor/compactor.go
ishan16696: yes, make sense I will do it

pkg/etcdutil/etcdutil.go
ishan16696: Hi @amshuman-kr , thanks for review. Hope you are doing well. 
`TakeAndSaveFullSnapshot`  is getting called from [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/compactor/compactor.go#L150)(in compaction) and [here](https://github.com/gardener/etcd-backup-restore/blob/5a7fd60f0bf704672a987dcab83c0c156041ff21/pkg/snapshot/snapshotter/snapshotter.go#L303)(in snapshotter).
In compaction, yes it make sense to pass  `client.MaintenanceCloser` instead of `client.Factory` as we already have  `client.MaintenanceCloser` object.
But In snapshotter,  to pass  `client.MaintenanceCloser` we have to first get the `client.MaintenanceCloser` object in [takeFullSnapshot func call](https://github.com/gardener/etcd-backup-restore/blob/5a7fd60f0bf704672a987dcab83c0c156041ff21/pkg/snapshot/snapshotter/snapshotter.go#L264).
So, I thought it will be better to make  `TakeAndSaveFullSnapshot` more abstract by letting the `TakeAndSaveFullSnapshot` to create client by itself and we have to just pass the `client.Factory`

amshuman-kr: `takeFullSnapshot` is already constructing the client object in [the existing code](https://github.com/gardener/etcd-backup-restore/blob/5a7fd60f0bf704672a987dcab83c0c156041ff21/pkg/snapshot/snapshotter/snapshotter.go#L269-L275). So, why not keep the same flow even the changes in this PR?

ishan16696: yeah, make sense but don't you think it will be overhead for caller to create client instead of just passing the configuration and let the `TakeAndSaveFullSnapshot` do all work

amshuman-kr: > but don't you think it will be overhead for caller to create client instead of just passing the configuration and let the TakeAndSaveFullSnapshot do all work

If it is overhead it is already overhead. Also, in the compaction case, it would lead to unnecessary creation of an additional client object. If there were more places like the `takeFullSnapshot` then it might have tilted the argument in favour of creating a client. But as it stands, creating a client outside and passing it to `TakeAndSaveFullSnapshot` seems better.

ishan16696: https://github.com/gardener/etcd-backup-restore/pull/381#discussion_r712076941, so I'm not moving `clientMaintenance` to inside the  `opts.NeedDefragmentation {}` block because of scope as now I need `clientMaintenance` to pass `TakeAndSaveFullSnapshot`

pkg/compactor/compactor.go
ishan16696: refer https://github.com/gardener/etcd-backup-restore/pull/381#discussion_r712148016

amshuman-kr (approved):
Thanks for the changes @ishan16696. LGTM

ishan16696:
I have missed one thing [here](https://github.com/gardener/etcd-backup-restore/pull/381/files#diff-adc81f6ffba52bc1871b0dac7c88237aa1cdea40e7691e35e834b40129ab723fR101) deep copy instance [cmpctOptions](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/compactor/compactor.go#L53) should be used. Now, I have also generate DeepCopy for NewClientFactory, although we are using `NewClientFactory=nil` but it good to have deepCopy also.

cc @amshuman-kr

ishan16696:
Rebased my PR and updated the cluster based Integration test logs too here https://github.com/gardener/etcd-backup-restore/pull/381#issuecomment-918877613

shreyas-s-rao (approved):
@ishan16696 Thanks for the well-written PR. LGTM!

PR #382, opened 2021-09-20
Added new package to allow sidecar to update member lease and snapsho…

### LINKED ISSUE #333: [Feature] Make etcd-backup-restore Kubernetes-aware keep member and backup status up-to-date

Make etcd-backup-restore Kubernetes-aware keep the corresponding [member status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#members) and [`BackupReady` condition](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#backup-failure) up-to-date in the `Etcd` resource [status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#status).

`etcd-backup-restore` is the closest to the source of this information and making it responsible to update these status sections will be a solution at the lowest viable level and avoid `etcd-druid` becoming a bottleneck.

1. Member status can be obtained by a heart-beat on the corresponding ETCD member's [endpoint status](https://github.com/etcd-io/etcd/blob/f82b5cb7768dacad9fb310232c1383b4e6718378/client/v3/maintenance.go#L53). The same heart-beat can be used to decide backup leader election in #321 
1. The `BackupReady` status could be determined based on whether latest backup upload ([full](https://github.com/gardener/etcd-backup-restore/blob/28e9b008e1d7b55e41b23c90d1c48af6c9fedd56/pkg/snapshot/snapshotter/snapshotter.go#L271) or [incremental](https://github.com/gardener/etcd-backup-restore/blob/28e9b008e1d7b55e41b23c90d1c48af6c9fedd56/pkg/snapshot/snapshotter/snapshotter.go#L381)) successful.

aaronfern:
…t leases

Added a new package `health` to allow etcd-backup sidecar to be kubernetes aware. Backup sidecar can now update/renew member leases to indicate member health in the multi node scenario and can update snapshot revision leases when snapshots are successfully taken

Fixes #333

**Release note**:

```improvement operator
Added `health` package to allow backup-restore to renew member leases to indicate member health and snapshot leases to indicate snapshots being taken successfully
```


amshuman-kr:
A quick comment @aaronfern. I see that you are using `k8s.io/client-go`. Why not use `sigs.k8s.io/controller-runtime/pkg/client`? That interface is simpler to use as well as to test using mock implementation generated by `gomock`. WDYT?

aaronfern:
Hi @amshuman-kr, thanks for taking a look
Yes, using `sigs.k8s.io/controller-runtime/pkg/client` might be a bit easier, however I'm running into a lot of import error during `make revendor`. (case-insensitive import collisions)
I'll try to resolve these and incorporate `sigs.k8s.io`

ishan16696:
I got this warning even though I set `--enable-etcd-lease-renewal` flag as `false`

```go
WARN[0030] 2 Failed to update delta snapshot lease with error: Pod namespace env var not present: missing environment variable POD_NAMESPACE  actor=snapshotter
```

aaronfern:
Hey @ishan16696 
yes, that's expected. That flag was meant only to limit backup sidecars ability to update the member lease.
It would not affect it's ability to try to update the snapshot leases

ishan16696:
Hey @aaronfern ,
Thanks to clarify but I'm not seeing anywhere where this `--enable-etcd-lease-renewal` flag is used.
Is this something related to [this](https://github.com/gardener/etcd-backup-restore/pull/382/files#diff-df6f7abbd9069d0a874a18b5717de5b0fcd68dac32184fdcdc2a26019be2296fR160-R165).

aaronfern:
Yes, it's related to that
It's mainly part of the multi-node aspect that would come up later, so I thought it best to comment it out all together
wdyt?

ishan16696:
thanks to clarify .... I was wondering it is commented.
> It's mainly part of the multi-node aspect that would come up later, so I thought it best to comment it out all together.

It will be better to remove it for now because eventually we have to move this inside `leaderElection`  as only `Leader` is allowed to  cleanup the superfluous entries

aaronfern:
Sure, I'll remove the commented portion there and push again to this branch

ishan16696:
It also passed the Cluster-Based Integration Tests. I have attached the log file
[clusterBasedIntegrationTest_PR382.log](https://github.com/gardener/etcd-backup-restore/files/7249833/clusterBasedIntegrationTest_PR382.log)

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Pod namespace env var not present: %v", err),
+		}
+	}
+
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Failed to create k8s clientset: %v", err),
+		}
+	}
```
ishan16696, line 160: if `err!=nil` then you are already returning the error above(line no 150 to 154) then this is unreachable.

ishan16696 (changes requested):
Thanks @aaronfern for PR!!. some nitpicks

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	restConfig, err := config.GetConfig()
+	var cl client.Client
+	if err == nil {
+		cl, err = client.New(restConfig, client.Options{})
+	}
+	if err != nil {
+		logger.Info("could not load InCluster config")
+		cl = fake.NewClientBuilder().Build()
+	}
```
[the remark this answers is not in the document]
aaronfern: yes, if either of these returns an error, fake client is created

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+
+	return nil
+}
+
+// RenewMemberLeasePeriodically has a timer and will periodically call RenewMemberLeases to renew the member lease until stopped
+func RenewMemberLeasePeriodically(ctx context.Context, hconfig *brtypes.HealthConfig, logger *logrus.Entry, etcdConfig *etcdutil.EtcdConnectionConfig) error {
+
+	hb, err := NewHeartbeat(hconfig, logger, etcdConfig)
+	if err != nil {
+		hb.logger.Warnf("Could not initialize new heartbeat: %v", err)
```
timuthy, line 247: Why do we only log but continue when there was an error when creating the Heartbeat instance?

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+		memberID = strconv.FormatUint(response.Header.MemberId, 16) + ":Member"
+	}
+
+	//Change HolderIdentity and RenewTime of lease
+	renewedMemberLease := memberLease.DeepCopy()
+	actor := memberID
+	renewedMemberLease.Spec.HolderIdentity = &actor
+	renewedTime := time.Now()
+	renewedMemberLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
+	renewedMemberLease.Spec.LeaseDurationSeconds = &hb.leaseDuration
```
timuthy, line 125: Isn't the `LeaseDurationSeconds` rather a value that is filled out by the Druid when it creates the `Lease`?

timuthy (changes requested):
Thanks for the PR @aaronfern. I have left a few suggestions and questions here and there.

pkg/server/backuprestoreserver.go
```diff
@@ -157,6 +157,12 @@ func (b *BackupRestoreServer) runServerWithSnapshotter(ctx context.Context, rest
 
 	go defragmentor.DefragDataPeriodically(ctx, b.config.EtcdConnectionConfig, b.defragmentationSchedule, ssr.TriggerFullSnapshot, b.logger)
 
+	// Lease renewal and lease GC enables only of enable flag set to true. Needed for backward compatability
```
shreyas-s-rao, line 160: Please add this as a `// TODO: @aaronfern ` so that you remember to pick it up later. Just a plain comment might be confusing to read until you've activated the member lease renewal.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -558,6 +574,12 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if err != nil {
 				return err
 			}
+			if err = heartbeat.UpdateFullSnapshotLeases(ssr.logger, ssr.PrevFullSnapshot, ssr.k8sClientset); err != nil {
```
shreyas-s-rao, line 577: Please add a check like `if b.config.HealthConfig.Enabled` so that it doesn't attempt to renew the leases unless lease renewal is enabled.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if err != nil {
 				return err
 			}
+			latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
+			if err == nil {
+				if err = heartbeat.UpdateDeltaSnapshotLeases(ssr.logger, latestFullSnapshot, latestDeltaSnapshotList, ssr.k8sClientset); err != nil {
```
shreyas-s-rao, line 596: Same as previous comment.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+					ssr.logger.Warnf("1 Failed to update delta snapshot lease with error: %v", err)
+				}
+			}
 
 		case <-ssr.fullSnapshotTimer.C:
 			if _, err := ssr.TakeFullSnapshotAndResetTimer(); err != nil {
 				return err
 			}
+			if err := heartbeat.UpdateFullSnapshotLeases(ssr.logger, ssr.PrevFullSnapshot, ssr.k8sClientset); err != nil {
```
shreyas-s-rao, line 605: Same as previous comment.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+			}
 
 		case <-ssr.deltaSnapshotTimer.C:
 			if ssr.config.DeltaSnapshotPeriod.Duration >= time.Second {
 				if _, err := ssr.takeDeltaSnapshotAndResetTimer(); err != nil {
 					return err
 				}
+				latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
+				if err == nil {
+					if err := heartbeat.UpdateDeltaSnapshotLeases(ssr.logger, latestFullSnapshot, latestDeltaSnapshotList, ssr.k8sClientset); err != nil {
```
shreyas-s-rao, line 619: Same as previous comment.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
 			}
+			snapshots := len(ssr.PrevDeltaSnapshots)
 			if err := ssr.handleDeltaWatchEvents(wr); err != nil {
 				return err
 			}
+			//Call UpdateDeltaSnapshotLease only if new delta snapshot taken
+			if snapshots < len(ssr.PrevDeltaSnapshots) {
+				latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
+				if err == nil {
+					if err := heartbeat.UpdateDeltaSnapshotLeases(ssr.logger, latestFullSnapshot, latestDeltaSnapshotList, ssr.k8sClientset); err != nil {
```
shreyas-s-rao, line 637: Same as previous comment.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if err != nil {
 				return err
 			}
+			latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
```
shreyas-s-rao, line 594: Please add error handling for this `err` when it is not `nil`.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+			if err := heartbeat.UpdateDeltaSnapshotLeases(ssr.logger, ssr.PrevFullSnapshot, nil, ssr.k8sClientset); err != nil {
+				ssr.logger.Warnf("1 Failed to update delta snapshot lease with error: %v", err)
+			}
 
 		case <-ssr.deltaSnapshotTimer.C:
 			if ssr.config.DeltaSnapshotPeriod.Duration >= time.Second {
 				if _, err := ssr.takeDeltaSnapshotAndResetTimer(); err != nil {
 					return err
 				}
+				latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
```
shreyas-s-rao, line 617: Same as previous comment.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
 			if !ok {
 				return fmt.Errorf("watch channel closed")
 			}
+			snapshots := len(ssr.PrevDeltaSnapshots)
 			if err := ssr.handleDeltaWatchEvents(wr); err != nil {
 				return err
 			}
+			//Call UpdateDeltaSnapshotLease only if new delta snapshot taken
+			if snapshots < len(ssr.PrevDeltaSnapshots) {
+				latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
```
shreyas-s-rao, line 635: Same as previous comment.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +591,54 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if err != nil {
 				return err
 			}
+			latestFullSnapshot, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(ssr.store)
+			if err == nil {
+				if err = heartbeat.UpdateDeltaSnapshotLeases(ssr.logger, latestFullSnapshot, latestDeltaSnapshotList, ssr.k8sClientset); err != nil {
+					ssr.logger.Warnf("1 Failed to update delta snapshot lease with error: %v", err)
```
shreyas-s-rao, line 597: Is there a particular reason for the `1` at the beginning of the error log? I see this in the subsequent logs as well, like 1, 2 and 3. Can you please explain this? Maybe it gives some extra information for debugging live issues, or it was just used for testing your PR perhaps?

pkg/types/health.go
```diff
@@ -0,0 +1,83 @@
...
+	flag "github.com/spf13/pflag"
+)
+
+const (
+	// DefaultEnableEtcdLeaseRenewal is a flag where a user can choose to turn off this lease renewal feature
+	DefaultEnableEtcdLeaseRenewal = false
+	// FullSnapshotLeaseName is the name for the delta snapshot lease.
+	FullSnapshotLeaseName = "full-snapshot-revisions"
+	// DeltaSnapshotLeaseName is the name for the delta snapshot lease.
+	DeltaSnapshotLeaseName = "delta-snapshot-revisions"
```
shreyas-s-rao, line 31: I think these parameters need to be configurable via CLI. Snapshot lease names should not be hardcoded into backup sidecar, because there is a contract between druid and backup sidecar about which leases are used. Since druid creates the leases, it also needs to be able to communicate these names to backup sidecar. Hence, we need CLI flags for the lease names.
You can still keep these as the default values, but ideally you should be taking these from the user via CLI flags before setting to the default values.

pkg/types/health.go
```diff
@@ -0,0 +1,83 @@
...
+	// DefaultHeartbeatDuration is the default heartbeat duration or lease renewal deletion.
+	DefaultHeartbeatDuration = 30 * time.Second
+	// DefaultMemberGarbagecollectionSchedule is the default member garbage collection period.
+	DefaultMemberGarbagecollectionSchedule = 60 * time.Second
```
shreyas-s-rao, line 35: Post our recent discussion, you will be removing the membergarbagecollection package for now and re-introducing it later right? But in any case, please rename this variable to `DefaultMemberGarbageCollectionPeriod`, since it defines a period, not a cron schedule as such.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
+		}
+	}
+	logger.Info("Renewed full snapshot lease with revision ", actor, " at time ", renewedTime)
+
+	return nil
+}
+
+// UpdateDeltaSnapshotLeases updates the holderIdentity field of the delta snapshot lease with the total number or revisions stored in delta snapshots since the last full snapshot was taken
+func UpdateDeltaSnapshotLeases(logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, PrevDeltaSnapshots brtypes.SnapList, k8sClientset client.Client) error {
```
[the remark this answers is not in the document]
shreyas-s-rao: `PrevDeltaSnapshots` -> `prevDeltaSnapshots` and also change occurences in the function

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Failed to get status of etcd endPoint [ %v ] with error: %v", hb.etcdConfig.Endpoints[0], err),
+		}
+	}
+	memberID := ""
+	if response.Header.MemberId == response.Leader {
+		memberID = strconv.FormatUint(response.Header.MemberId, 16) + ":Leader"
+	} else {
+		memberID = strconv.FormatUint(response.Header.MemberId, 16) + ":Member"
```
shreyas-s-rao, line 116: Since the `strconv.FormatUint(response.Header.MemberId, 16)` is common to both paths in the `if` construct, you can move it out of the `if` block to avoid redundancy in computation.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	//Change HolderIdentity and RenewTime of lease
+	renewedMemberLease := memberLease.DeepCopy()
+	actor := memberID
```
shreyas-s-rao, line 121: Redundant variable `actor`. You can directly set `renewedMemberLease.Spec.HolderIdentity = &memberID`

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,119 @@
...
+		cl = fake.NewClientBuilder().Build()
+	}
+	return &MemberGarbageCollector{
+		logger:    logger.WithField("actor", "membergarbagecollector"),
+		K8sClient: cl,
+	}, nil
+}
+
+// RunMemberGarbageCollector checks if any member leases that are not bound to pods are present in the cluster and deletes them if found
+func (membergc *MemberGarbageCollector) RunMemberGarbageCollector() {
```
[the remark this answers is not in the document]
shreyas-s-rao: You're right @timuthy . I think I had got a little confused when explaining the PR requirements to @aaronfern initially. Just had an out-of-band discussion with him to clarify the responsibilities of druid and sidecar.

pkg/health/heartbeat/heartbeat_suite_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"os"
+	"testing"
+
+	"github.com/gardener/etcd-backup-restore/test/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+	"go.etcd.io/etcd/embed"
+	//"sigs.k8s.io/controller-runtime/pkg/manager"
```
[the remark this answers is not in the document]
aaronfern: Done

pkg/health/heartbeat/heartbeat_suite_test.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+var (
+	logger  = logrus.New().WithField("suite", "snapshotter")
+	etcd    *embed.Etcd
+	err     error
+	testCtx = context.Background()
+	//mgr     manager.Manager
```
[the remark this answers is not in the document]
aaronfern: Done

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Pod namespace env var not present: %v", err),
+		}
+	}
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Failed to create k8s clientset: %v", err),
+		}
+	}
```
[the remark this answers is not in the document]
aaronfern: Thanks for catching this. Removed unreachable code

pkg/health/heartbeat/heartbeat.go
aaronfern: Thanks for catching this. Removed unreachable code

pkg/health/membergarbagecollector/membergarbagecollector.go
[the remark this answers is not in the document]
aaronfern: Updated the PR to not include lease deletion. Apologies, got a little bit confused there.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	restConfig, err := config.GetConfig()
+	var cl client.Client
+	if err == nil {
+		cl, err = client.New(restConfig, client.Options{})
+	}
+	if err != nil {
+		logger.Info("could not load InCluster config")
+		cl = fake.NewClientBuilder().Build()
```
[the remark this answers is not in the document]
aaronfern: Thanks. Updated the code to not create a fake client as part of the flow.
Now using the fake client for testing directly from the test file

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	leaseDuration := int32(hconfig.LeaseDuration.Seconds())
+	if leaseDuration < 1 {
+		return nil, &errors.EtcdError{
+			Message: "Could not initialize new heartbeat: Duration provided is too small",
+		}
+	}
+	restConfig, err := config.GetConfig()
+	var cl client.Client
+	if err == nil {
+		cl, err = client.New(restConfig, client.Options{})
```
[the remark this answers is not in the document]
aaronfern: Updated the code for creating clientset.
ptal and let me know

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+	}
+	namespace, err := GetEnvVarOrError("POD_NAMESPACE")
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("POD_NAMESPACE env var not present: %v", err),
+		}
+	}
+	memberLease := &v1.Lease{}
+	err = hb.K8sClient.Get(context.TODO(), client.ObjectKey{
```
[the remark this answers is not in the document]
aaronfern: Thnaks for catching this. Updated with proper contexts.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	heartbeatTimer *time.Timer
+	etcdConfig     *etcdutil.EtcdConnectionConfig
+	leaseDuration  int32
+	K8sClient      client.Client
+}
+
+// NewHeartbeat returns the heartbeat object.
+func NewHeartbeat(hconfig *brtypes.HealthConfig, logger *logrus.Entry, etcdConfig *etcdutil.EtcdConnectionConfig) (*Heartbeat, error) {
+	leaseDuration := int32(hconfig.LeaseDuration.Seconds())
+	if leaseDuration < 1 {
```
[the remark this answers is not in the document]
aaronfern: Is it already part of `pkg/types/health.go`
I had added it here as an extra check initially and later thought it might provide a secondary safety check
wdyt?

pkg/health/heartbeat/heartbeat.go
aaronfern: Thanks for catching this
Updated to return an error if heartbeat could not be initialized

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
aaronfern: Is it already part of `pkg/types/health.go`
I had added it here as an extra check initially and later thought it might provide a secondary safety check
wdyt?

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
aaronfern: Updated the code to create the clientset
ptal and let me know

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+
+// UpdateFullSnapshotLeases updates the holderIdentity field of the full snapshot lease with the last revision in the latest full snapshot
+func UpdateFullSnapshotLeases(logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client) error {
+	if fullSnapshot == nil {
+		return &errors.EtcdError{
+			Message: "can not update full snapshot lease, nil snapshot passed",
+		}
+	}
+
+	namespace, err := GetEnvVarOrError("POD_NAMESPACE")
```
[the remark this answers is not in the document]
aaronfern: Added constants to the file with variables for `POD_NAMESPACE` and `POD_NAME`

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+		etcdConfig:    etcdConfig,
+		leaseDuration: leaseDuration,
+		K8sClient:     cl,
+	}, nil
+}
+
+// RenewMemberLeases is called periodically and renews the pods member lease
+func (hb *Heartbeat) RenewMemberLeases() error {
+	//Fetch lease associated with member
+	memberName, err := GetEnvVarOrError("POD_NAME")
```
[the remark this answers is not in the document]
aaronfern: Added constants to the file with variables for `POD_NAMESPACE` and `POD_NAME

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	fmt.Println("member name is ", memberName)
```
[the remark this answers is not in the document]
aaronfern: Thanks for catching it. Yes, it was only for debugging
removed

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+	fake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+// Heartbeat is a struct for the heartbeat package
+type Heartbeat struct {
+	logger         *logrus.Entry
+	heartbeatTimer *time.Timer
+	etcdConfig     *etcdutil.EtcdConnectionConfig
+	leaseDuration  int32
+	K8sClient      client.Client
```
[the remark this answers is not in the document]
aaronfern: `K8sClient` now is being set to fake client from the test file for tests [here](https://github.com/gardener/etcd-backup-restore/pull/382/files#diff-8ccda3b0950c83908fa626e64ff6498880cd7a992eeba44901eec578c8962ad4R342)
wdyt? is there any way I could go around this?

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+	}
+
+	hb.logger.Info("Renewed member lease for etcd at ", renewedTime)
+
+	return nil
+}
+
+// UpdateFullSnapshotLeases updates the holderIdentity field of the full snapshot lease with the last revision in the latest full snapshot
+func UpdateFullSnapshotLeases(logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client) error {
```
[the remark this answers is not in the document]
aaronfern: Done

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
aaronfern: Done

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,277 @@
...
+			hb.heartbeatTimer.Reset(hconfig.HeartbeatDuration.Duration)
+		case <-ctx.Done():
+			hb.logger.Info("Stopped member lease renewal timer")
+			return nil
+		}
+	}
+}
+
+// GetEnvVarOrError returns the value of specified environment variable or terminates if it's not defined.
+func GetEnvVarOrError(varName string) (string, error) {
```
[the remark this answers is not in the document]
aaronfern: Updated to use the function from the `snapstore` package

pkg/server/backuprestoreserver.go

pkg/types/health.go
aaronfern: Updated the code to add two flags to the `server` subcommand (`full-snapshot-lease-name` and `delta-snapshot-lease-name`) and one flag to the `compact` subcommand (`full-snapshot-lease-name`)
However these parameters remain as default values

pkg/health/heartbeat/heartbeat.go
aaronfern: Updated this

pkg/health/heartbeat/heartbeat.go
aaronfern: Removed

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -558,6 +576,16 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if err != nil {
 				return err
 			}
+			if ssr.healthConfig.Enabled {
+				ctx, cancel := context.WithTimeout(context.TODO(), ssr.config.DeltaSnapshotPeriod.Duration)
```
timuthy, line 580: Can you please use a proper context?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +597,76 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
```
timuthy, line 601: Can you please use a proper context?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -119,12 +124,24 @@ func NewSnapshotter(logger *logrus.Entry, config *brtypes.SnapshotterConfig, sto
 
 	metrics.LatestSnapshotRevision.With(prometheus.Labels{metrics.LabelKind: prevSnapshot.Kind}).Set(float64(prevSnapshot.LastRevision))
 
+	var cl client.Client
+	restConfig, err := cntrlr_config.GetConfig()
+	if err != nil {
+		logger.Info("Could not get InCluster config")
```
ishan16696, line 130: same as previous comment

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -558,6 +576,16 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if err != nil {
 				return err
 			}
+			if ssr.healthConfig.Enabled {
+				ctx, cancel := context.WithTimeout(context.TODO(), ssr.config.DeltaSnapshotPeriod.Duration)
+				if err = heartbeat.UpdateFullSnapshotLease(ctx, ssr.logger, ssr.PrevFullSnapshot, ssr.k8sClientset, ssr.healthConfig.FullSnapshotLeaseName); err != nil {
+					ssr.logger.Warnf("1 Failed to update full snapshot lease with error: %v", err)
```
timuthy, line 582: What does `1` mean?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -119,12 +124,24 @@ func NewSnapshotter(logger *logrus.Entry, config *brtypes.SnapshotterConfig, sto
...
 	metrics.LatestSnapshotRevision.With(prometheus.Labels{metrics.LabelKind: prevSnapshot.Kind}).Set(float64(prevSnapshot.LastRevision))
 
+	var cl client.Client
+	restConfig, err := cntrlr_config.GetConfig()
+	if err != nil {
+		logger.Info("Could not get InCluster config")
+	} else {
+		cl, err = client.New(restConfig, client.Options{})
+		if err != nil {
+			logger.Info("could not create kubernetes clientset")
```
ishan16696, line 134: same as previous comment

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +597,76 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+				}
+				cancel()
+			}
 
 		case <-ssr.fullSnapshotTimer.C:
 			if _, err := ssr.TakeFullSnapshotAndResetTimer(); err != nil {
 				return err
 			}
+			if ssr.healthConfig.Enabled {
+				ctx, cancel := context.WithTimeout(context.TODO(), ssr.config.DeltaSnapshotPeriod.Duration)
```
timuthy, line 618: Please use a proper context.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +597,76 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+				if err := heartbeat.UpdateFullSnapshotLease(ctx, ssr.logger, ssr.PrevFullSnapshot, ssr.k8sClientset, ssr.healthConfig.FullSnapshotLeaseName); err != nil {
+					ssr.logger.Warnf("2 Failed to update full snapshot lease with error: %v", err)
```
timuthy, line 620: Why `2`?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +597,76 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+				cancel()
+			}
 
 		case <-ssr.deltaSnapshotTimer.C:
 			if ssr.config.DeltaSnapshotPeriod.Duration >= time.Second {
 				if _, err := ssr.takeDeltaSnapshotAndResetTimer(); err != nil {
 					return err
 				}
+				if ssr.healthConfig.Enabled {
+					ctx, cancel := context.WithTimeout(context.TODO(), ssr.config.DeltaSnapshotPeriod.Duration)
```
timuthy, line 634: Please use a proper context.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -569,26 +597,76 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
 				return fmt.Errorf("watch channel closed")
 			}
+			snapshots := len(ssr.PrevDeltaSnapshots)
 			if err := ssr.handleDeltaWatchEvents(wr); err != nil {
 				return err
 			}
+			if ssr.healthConfig.Enabled {
+				//Call UpdateDeltaSnapshotLease only if new delta snapshot taken
+				if snapshots < len(ssr.PrevDeltaSnapshots) {
+					ctx, cancel := context.WithTimeout(context.TODO(), ssr.config.DeltaSnapshotPeriod.Duration)
```
timuthy, line 658: Please use a proper context.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,257 @@
...
+func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, prevDeltaSnapshots brtypes.SnapList, k8sClientset client.Client, deltaSnapshotLeaseName string) error {
+	namespace, err := utils.GetEnvVarOrError(podNamespace)
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Pod namespace env var not present: %v", err),
+		}
+	}
+
+	deltaSnapLease := &v1.Lease{}
+	err = k8sClientset.Get(ctx, client.ObjectKey{
```
ishan16696, line 197: I would suggest better to check the null condition with  `k8sClientset`.

```go
 if k8sClientset!=nil {

} else{
// handle the error
}
```

pkg/types/compactor.go
```diff
@@ -56,6 +58,7 @@ func (c *CompactorConfig) AddFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&c.NeedDefragmentation, "defragment", c.NeedDefragmentation, "defragment after compaction")
 	fs.DurationVar(&c.SnapshotTimeout.Duration, "etcd-snapshot-timeout", c.SnapshotTimeout.Duration, "timeout duration for taking compacted full snapshots")
 	fs.DurationVar(&c.DefragTimeout.Duration, "etcd-defrag-timeout", c.DefragTimeout.Duration, "timeout duration for etcd defrag call during compaction.")
+	fs.StringVar(&c.FullSnapshotLeaseName, "full-snapshot-lease-name", c.FullSnapshotLeaseName, "full snapshot lease name")
```
timuthy, line 61: Why do we need that option for the compactor?

pkg/server/backuprestoreserver.go
```diff
@@ -157,6 +157,8 @@ func (b *BackupRestoreServer) runServerWithSnapshotter(ctx context.Context, rest
 
 	go defragmentor.DefragDataPeriodically(ctx, b.config.EtcdConnectionConfig, b.defragmentationSchedule, ssr.TriggerFullSnapshot, b.logger)
 
+	//TODO @aaronfern: Call function for member lease renewal if enable flag is set. Needed for backward compatability
```
timuthy, line 160: I suggest to introduce the flag in this PR and enable the coding if the flag is set (disabled by default). Otherwise we introduce a lot of coding today which is not executed.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+	namespace, err := utils.GetEnvVarOrError(podNamespace)
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("POD_NAMESPACE env var not present: %v", err),
+		}
+	}
+	memberLease := &v1.Lease{}
+	err = hb.K8sClient.Get(ctx, client.ObjectKey{
```
ishan16696, line 92: I would suggest better to check the null condition with `hb.k8sClientset`.

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
timuthy: Hm, can you elaborate why you think we need that extra safety measure? Is there a valid use-case?

ishan16696 (changes requested):
Thanks @aaronfern  for changes, Please address few concerns mentioned below.

timuthy (changes requested):
Thanks for addressing the PR feedback @aaronfern. I had a second look and left some feedback.

pkg/snapshot/snapshotter/snapshotter.go
aaronfern: It was used for local debugging. Meant to be removed, however missed it last time. 
[here](https://github.com/gardener/etcd-backup-restore/pull/382#discussion_r719575321)

pkg/types/compactor.go
aaronfern: The compaction job would just call the `compact` subcommand of `etcdbrctl`. If druid deploys a full-snapshot-lease with a unique name , then druid would need a way to pass on the lease name so that the correct one is updated after compaction
[here](https://github.com/gardener/etcd-backup-restore/pull/382#discussion_r718566805) and [here](https://github.com/gardener/etcd-backup-restore/pull/382#discussion_r719575390)

timuthy: Understood, thanks for explaining.
I wonder if we need to update the `delta-snapshot-lease` as well after the compaction. At the moment it only happens when the next snapshot is taken, right? This can be too late because Druid might already start another compaction job based on an "old" delta revision number. (/cc @shreyas-s-rao)

pkg/snapshot/snapshotter/snapshotter.go
timuthy: Thanks

pkg/types/compactor.go
aaronfern: Yes, that is right. `delta-snapshot-lease` would only be updated when the next delta snapshot is taken
My understanding here is that druid will only react to the `delta-snapshot-lease` in the event of an update in the spec. So, druid would not start a new compaction job with the old delta revision number.

timuthy: Hm, that won't work if Druid is restarted for some reason (e.g. it is scaled by VPA). Then Druid will need to consider the `create` event for the Lease and doesn't know if it already started a compaction job or not.

pkg/compactor/compactor.go
```diff
@@ -33,15 +37,27 @@ const (
...
+	store        brtypes.SnapStore
+	k8sClientset client.Client
 }
 
 // NewCompactor creates compactor
 func NewCompactor(store brtypes.SnapStore, logger *logrus.Entry) *Compactor {
+	var cl client.Client
+	restConfig, err := config.GetConfig()
+	if err != nil {
+		logger.Info("Could not get InCluster config")
```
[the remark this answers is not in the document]
aaronfern: Updated to throw errors if client creation fails.
Ptal

pkg/compactor/compactor.go
```diff
@@ -151,6 +167,11 @@ func (cp *Compactor) Compact(opts *brtypes.CompactOptions) (*brtypes.Snapshot, e
 	if err != nil {
 		return nil, err
 	}
+	//Update revisions in holder identity of full snapshot lease and delta snapshot lease
```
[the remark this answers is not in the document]
aaronfern: Thank you. updated

pkg/compactor/compactor.go
```diff
@@ -33,15 +37,27 @@ const (
...
+	var cl client.Client
+	restConfig, err := config.GetConfig()
+	if err != nil {
+		logger.Info("Could not get InCluster config")
+	} else {
+		cl, err = client.New(restConfig, client.Options{})
+		if err != nil {
+			logger.Info("could not create kubernetes clientset")
+		}
+	}
```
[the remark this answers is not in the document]
aaronfern: Updated to pass the clientset from outside and handle error from client creation

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,257 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const (
+	podName      = "POD_NAME"
+	podNamespace = "POD_NAMESPACE"
+)
+
+// Heartbeat is a struct for the heartbeat package
```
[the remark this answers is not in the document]
aaronfern: Thank you. Updated

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	return &Heartbeat{
+		logger:        logger.WithField("actor", "heartbeat"),
+		etcdConfig:    etcdConfig,
+		leaseDuration: leaseDuration,
+		K8sClient:     cl,
+	}, nil
+}
+
+// RenewMemberLeases is called periodically and renews the pods member lease
```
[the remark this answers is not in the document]
aaronfern: Thank you. Updated

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,257 @@
...
+	//Change HolderIdentity and RenewTime of lease
+	renewedMemberLease := memberLease.DeepCopy()
+	renewedMemberLease.Spec.HolderIdentity = &memberID
+	renewedTime := time.Now()
+	renewedMemberLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
+	renewedMemberLease.Spec.LeaseDurationSeconds = &hb.leaseDuration
+
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return hb.K8sClient.Patch(ctx, renewedMemberLease, client.MergeFrom(memberLease))
+	})
```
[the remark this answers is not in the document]
aaronfern: Updated

pkg/snapshot/snapshotter/snapshotter.go
aaronfern: Sorry, that was used only for debugging
removed

pkg/health/heartbeat/heartbeat.go
aaronfern: Agreed. Removed this piece of code

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
aaronfern: As a consequence of [this](https://github.com/gardener/etcd-backup-restore/pull/382#discussion_r721313644),
this piece of code is not irrelevant and hence removed

pkg/types/compactor.go
aaronfern: After our out-of-band discussions, added a functionality to update delta snapshot on compaction as well

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
aaronfern: Updated to have the client passed to the constructor
`K8sClient` also updated to be a private variable

pkg/health/heartbeat/heartbeat.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+
+	hb.logger.Info("Renewed member lease for etcd at ", renewedTime)
+
+	return nil
+}
+
+// UpdateFullSnapshotLease updates the holderIdentity field of the full snapshot lease with the last revision in the latest full snapshot
+func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
```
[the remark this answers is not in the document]
aaronfern: These leases would also be used by druid to update the `backupReady` condition in the etcd status. So, in a sense they end up being kind of like heartbeats as well

PR #383, opened 2021-09-30
[GEP-17] Add owner checks and taking of final snapshots

stoyanr:
**What this PR does / why we need it**:
Adds functionality to check if the owner domain name resolves to the specified owner ID and if not, take a final full snapshot and disable the cluster, as described in [GEP-17](https://github.com/gardener/gardener/blob/master/docs/proposals/17-shoot-control-plane-migration-bad-case.md).

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
Based on #369 and so in Draft until that one is merged. I would appreciate early feedback (only the last 2 commits are relevant).

**Release note**:

```improvement operator
The backup-restore sidecar will now check if the owner domain name resolves to the specified owner ID and if not, take a final full snapshot and disable the cluster.
```

PR #388, opened 2021-10-04
Ensure integration tests can be executed locally

stoyanr:
Improves the integration tests in the following way:
* Ensures the integration tests can be executed locally by adding proper support for Darwin and using the existing credentials in the `$HOME/.aws` directory if the directory exists.
* If the tests fail, outputs the `etcdbrctl` logs.

Fixes #

While working on #383, I had some failing integration tests, but it was very difficult to tell the reason for the failures. It took me some time to adapt the `integration_test` script to be able to execute them locally, but even then I had no clue because they just passed. I had to add the `etcdbrctl` logs to the test output in order to be able to understand and fix the issue.

To run the integration tests locally:
* Ensure that AWS credentials exist in `$HOME/.aws` by running `aws configure`
* Run the integration tests via `.ci/integration_test` or `make integration-test`

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/.ci/integration_test
+++ b/.ci/integration_test
@@ -56,6 +56,11 @@ ETCD_DATA_DIR=
 TEST_DIR=
 TEST_RESULT=
 
+set +e
+test -d "${HOME}/.aws"
+USE_EXISTING_AWS_SECRET=$?
+set -e
+
 function setup_test_environment() {
   setup_ginkgo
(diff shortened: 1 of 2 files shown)


shreyas-s-rao (approved):
@stoyanr Thanks a lot for the PR. LGTM

PR #403, opened 2021-11-04
Regularly cleanup superfluous etcd members from the cluster

### LINKED ISSUE #321: [Feature] Add leader-election to the backup-restore sidecar according to the multi-node ETCD proposal.

Add leader-election to the backup-restore sidecar according to the multi-node ETCD proposal. Bring the [snapshotters](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#backup) (both full and incremental) and the [defragmentation](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#defragmentation) control loops/go routines under the control of the leader-election as described [here](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#work-flows-in-etcd-backup-restore).

Also, introduce the logic to regularly cleanup superfluous ETCD members which do not have a corresponding `Pod` in the `StatefulSet` as mentioned in https://github.com/gardener/etcd-druid/pull/189.

Pick individually executable pieces of the [multi-node proposal](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md).

The preferred leader-election mechanism (because of the implementation and operational simplicity) is based on [the leading etcd member](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#leading-etcd-main-containers-sidecar-is-the-backup-leader). This can be done using the same heart-beat on the corresponding ETCD member's [endpoint status](https://github.com/etcd-io/etcd/blob/f82b5cb7768dacad9fb310232c1383b4e6718378/client/v3/maintenance.go#L53) as mentioned in #333

aaronfern:
This PR introduce the logic to regularly cleanup superfluous ETCD members which do not have a corresponding pod.

This is required as mentioned in the multi-node proposal [here](https://github.com/gardener/etcd-druid/blob/a2e349bd4abf79710f050ef302c34c836739c734/docs/proposals/multi-node/README.md?plain=1#L604)

Fixes #321 Part 2

**Release note**:

```action developer
Added new package `membergarbagecollector` to remove superfluous members from the ETCD cluster. Due to this, etcd-backup-restore now needs permissions to list `pods` and `statefulsets`.
```
```improvement operator
Added new package `membergarbagecollector` to remove superfluous members from the ETCD cluster.
```


pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,173 @@
...
+		return
+	}
+	mgc, err := NewMemberGarbageCollector(logger, etcdConfig, clientSet)
+	if err != nil {
+		logger.Errorf("failed to initialize new heartbeat: %v", err)
+		return
+	}
+	mgc.gcTimer = time.NewTimer(hconfig.MemberGCDuration.Duration)
+	defer mgc.gcTimer.Stop()
+	mgc.logger.Info("Started etcd member garbage collection timer")
```
ishan16696, line 101:
```suggestion
	mgc.logger.Info("Starting the etcd member garbage collector")
```

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,173 @@
...
+	etcd := &druidv1alpha1.Etcd{}
+	if err := mgc.k8sClient.Get(ctx, client.ObjectKey{
+		Namespace: mgc.podNamespace,
+		Name:      mgc.etcdName,
+	}, etcd); err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Could not fetch etcd %v with error: %v", mgc.etcdName, err),
+		}
+	}
+	// Return is no members present
```
ishan16696, line 129:
```suggestion
	// Return if no etcd member present
```

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	// Return if memberlist size <= desired cluster size. Cleanup not needed
+	if len(etcd.Status.Members) <= etcd.Spec.Replicas {
+		return nil
+	}
+
+	//Create etcd client to get etcd cluster member list
+	etcdClient, err := etcdutil.GetTLSClientForEtcd(mgc.etcdConfig)
```
ishan16696, line 142: you have to change the `etcdClient` once [PR#381](https://github.com/gardener/etcd-backup-restore/pull/381) get merged

ishan16696 (changes requested):
Hi @aaronfern , thanks for PR.
Few nit picks from my side, please address them.

pkg/server/backuprestoreserver.go
```diff
@@ -177,9 +178,9 @@ func (b *BackupRestoreServer) runServerWithSnapshotter(ctx context.Context, rest
 
 	go defragmentor.DefragDataPeriodically(ctx, b.config.EtcdConnectionConfig, b.defragmentationSchedule, ssr.TriggerFullSnapshot, b.logger)
 
-	//TODO @aaronfern: Add functionality for member garbage collection
 	if b.config.HealthConfig.MemberLeaseRenewalEnabled {
 		go heartbeat.RenewMemberLeasePeriodically(ctx, b.config.HealthConfig, b.logger, b.config.EtcdConnectionConfig)
+		go membergarbagecollector.RunMemberGarbageCollectorPeriodically(ctx, b.config.HealthConfig, b.logger, b.config.EtcdConnectionConfig)
```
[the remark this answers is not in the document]
aaronfern: Are there any scenarios where we would like to have member lease updates but no member garbage collection or vice versa?
If yes then another flag makes sense, but if not then we could just rename the flag to something more appropriate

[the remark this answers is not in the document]
ishan16696: the thing is `membergarbagecollector.RunMemberGarbageCollectorPeriodically()` will only start by `leading-sidecar` and leading-sidecar will also gonna stop it once it is no longer `backup-leader`.
And heartbeat.RenewMemberLeasePeriodically() will start by both `backup-follower` as well as `backup-leader`. 
So both have different non-overlapping usage.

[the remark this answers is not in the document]
amshuman-kr: The reason for switching off periodic garbage collection of members is the same as for switching off member lease renewal, which is the backward-compatibility single-node scenario. I cannot think of any scenario of multi-node scenario where you might want to switch off either of them. It might make sense to keep separate flags as @ishan16696 mentioned but both will be switched on or off together.

Also, as @ishan16696 mentioned, it is better to start and stop the member garbage collector goroutine in the on the [start](https://github.com/gardener/etcd-backup-restore/pull/353/files#diff-df6f7abbd9069d0a874a18b5717de5b0fcd68dac32184fdcdc2a26019be2296fR188-R189) and [stop](https://github.com/gardener/etcd-backup-restore/pull/353/files#diff-df6f7abbd9069d0a874a18b5717de5b0fcd68dac32184fdcdc2a26019be2296fR192-R201) callback of backup leader election because it is supposed to be done only by the [leading backup sidecar](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#work-flows-only-on-the-leading-member).

amshuman-kr (commented):
Thanks for implementing more functionality from the multi-node story @aaronfern!

Just a a couple of comments/suggestions below.

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,173 @@
...
+		}, memberPod)
+		if apierrors.IsNotFound(err) {
+			//Remove member from etcd cluster if corresponding pod cannot be found
+			mgc.logger.Info("Removing superfluous member ", x.ID, " : ", err)
+			if _, err = etcdClient.MemberRemove(ctx, x.ID); err != nil {
+				return &errors.EtcdError{
+					Message: fmt.Sprintf("Error removing member %v from the cluster: %v", x.ID, err),
+				}
+			}
+		}
```
amshuman-kr, line 170: Here the member is removed if the corresponding pod doesn't exist. But `etch-druid` removes the member from the status based more checks (like `lease` etc.). So, it might be better to be fully sure by checking that the lease, `Etcd` `status.members` and `pods` all agree that this member is now gone. To be precise, if you check the `status.members` and the pods then checking the `leases` may not be required because the `status.members` is kept in sync with the `leases`.

pkg/server/backuprestoreserver.go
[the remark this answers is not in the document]
ishan16696: I think it is better to leave it like this for now once @aaronfern  rebased this PR over  [LeaderElection PR](https://github.com/gardener/etcd-backup-restore/pull/353) then we can decide whether it make sense to have separate flag or not.

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Changed to use `NewCluster`

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Done

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,179 @@
...
+	if len(etcd.Status.Members) < 1 {
+		return &errors.EtcdError{
+			Message: "There are no members present in the etcd status. Nothing to clean",
+		}
+	}
+
+	// Return if memberlist size <= desired cluster size. Cleanup not needed
+	if len(etcd.Status.Members) <= etcd.Spec.Replicas {
+		return nil
+	}
```
timuthy, line 147: We can't return here since due to a race the `Members` slice can already be updated by the Druid, i.e. `etcd.Status.Members == etcd.Spec.Replicas` but the members themselves are still in the cluster.

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	logger = logrus.New().WithField("suite", "snapshotter")
```
timuthy, line 13: Do we need log messages in tests? Otherwise we can use `logger.Out = io.Discard` to not pollute the test output.

vendor/modules.txt
```diff
@@ -230,6 +230,9 @@ github.com/exponent-io/jsonpath
 github.com/fatih/color
 # github.com/fsnotify/fsnotify v1.4.9
 github.com/fsnotify/fsnotify
+# github.com/gardener/etcd-druid/api v0.6.0
```
timuthy, line 233: Can you use an up-to-date version?

pkg/server/backuprestoreserver.go
```diff
@@ -185,6 +188,10 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 			}
 			go b.runEtcdProbeLoopWithSnapshotter(leCtx, handler, ssr, ss, ssrStopCh, ackCh)
 			go defragmentor.DefragDataPeriodically(leCtx, b.config.EtcdConnectionConfig, b.defragmentationSchedule, defragCallBack, b.logger)
+			//start etcd member garbage collector
+			if b.config.HealthConfig.MemberGCEnabled {
+				go membergarbagecollector.RunMemberGarbageCollectorPeriodically(ctx, memGCStopCh, b.config.HealthConfig, b.logger, b.config.EtcdConnectionConfig)
```
timuthy, line 193: Can't we pass `leCtx` instead of using both `ctx` and `memGCStopCh`? `leCtx` is cancelled when leader election is lost (see [here](https://github.com/gardener/etcd-backup-restore/blob/322d3c74fcd5ea7be9961ce5e9d7ffed48ecc34e/pkg/leaderelection/leaderelection.go#L135)).

timuthy (changes requested):
Some feedback from my side. Does it make sense to add a small document explaining the garbage collection?

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Thanks for catching this 👍🏻 
Removed unneeded condition

pkg/server/backuprestoreserver.go
aaronfern: Thanks for pointing this out. Updated to use only `leCtx` now

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
aaronfern: Done

vendor/modules.txt
aaronfern: Updated to use `v0.7.3` in [`eda8f83`](https://github.com/gardener/etcd-backup-restore/pull/403/commits/eda8f831bd047009f71581346da75ef3e4d4fbd0)

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,136 @@
...
+		memberPod := &v1.Pod{}
+		err := mgc.k8sClient.Get(ctx, client.ObjectKey{
+			Namespace: mgc.podNamespace,
+			Name:      x.Name,
+		}, memberPod)
+		if apierrors.IsNotFound(err) {
+			//Remove member from etcd cluster if corresponding pod cannot be found
+			mgc.logger.Info("Removing superfluous member ", x.ID, " : ", err)
+			etcdClient.MemberRemove(ctx, x.ID)
+		}
```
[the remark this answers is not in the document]
aaronfern: This message use updated in subsequent commits

pkg/types/health.go
```diff
@@ -61,16 +69,23 @@ func NewHealthConfig() *HealthConfig {
 func (c *HealthConfig) AddFlags(fs *flag.FlagSet) {
 
 	fs.BoolVar(&c.SnapshotLeaseRenewalEnabled, "enable-snapshot-lease-renewal", c.SnapshotLeaseRenewalEnabled, "Allows sidecar to renew the snapshot leases when snapshots are taken")
-	fs.BoolVar(&c.MemberLeaseRenewalEnabled, "enable-member-lease-renewal", c.MemberLeaseRenewalEnabled, "Allows sidecar to periodically renew the member leases when snapshots are taken")
+	fs.BoolVar(&c.MemberLeaseRenewalEnabled, "enable-member-lease-renewal", c.MemberLeaseRenewalEnabled, "Allows sidecar to periodically renew the member leases")
+	fs.BoolVar(&c.MemberGCEnabled, "enable-etcd-member-gc", c.MemberGCEnabled, "Allows leading sidecar to remove any superfluous etcd members from the cluster")
```
ishan16696, line 73:
```suggestion
	fs.BoolVar(&c.EtcdMemberGCEnabled, "enable-etcd-member-gc", c.MemberGCEnabled, "Allows leading sidecar to remove any superfluous etcd members from the cluster")
```
Here like this flag `--enable-etcd-member-gc` it is not ambiguous.

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+	mgc.logger.Info("Started etcd member garbage collector")
+
+	for {
+		//Create etcd client to get etcd cluster member list
+		clientFactory := etcdutil.NewFactory(*etcdConfig)
+		etcdCluster, err := clientFactory.NewCluster()
+		if err != nil {
+			mgc.logger.Errorf("Failed to create etcd cluster client: %v", err)
+		}
+		defer etcdCluster.Close()
```
ishan16696, line 107: I don't think this `defer etcdCluster.Close()` will ever get execute and it is in inside infinite loop `for{ .... }`  so, each time loop execute will create a new client and this may leads to exponentially increase in memory although I haven't tested yet.

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+		select {
+		case <-mgc.gcTimer.C:
+			err := mgc.RemoveSuperfluousMembers(ctx, etcdCluster)
+			if err != nil {
+				mgc.logger.Warn(err)
+			}
+			mgc.gcTimer.Reset(hconfig.MemberGCDuration.Duration)
+		case <-ctx.Done():
+			mgc.logger.Info("Stopping etcd member garbage collector")
+			etcdCluster.Close()
```
ishan16696, line 117: you can remove this line as above you are using `defer`  with etcdCluster.Close()

ishan16696 (changes requested):
I have few concern regarding that I have made few comments.

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	podNamespace = "POD_NAMESPACE"
+	podName      = "POD_NAME"
+)
+
+// MGCChecker is an interface used to unit test functions in this package
```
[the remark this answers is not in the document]
aaronfern: Updated, however I've still left a small note that this is used for unit tests only

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Valid concern
moved the `etcd cluster-client` creation outside the for loop as it's not needed to be created every single iteration

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+// NewMemberGarbageCollector returns the member garbage collect object
+func NewMemberGarbageCollector(logger *logrus.Entry, clientSet client.Client) (*MemberGarbageCollector, error) {
+	if clientSet == nil {
+		return nil, &errors.EtcdError{
+			Message: "nil kubernetes clientset passed, can not create etcd member garbage collector",
+		}
+	}
+	namespace, err := utils.GetEnvVarOrError(podNamespace)
+	if err != nil {
+		logger.Fatalf("POD_NAMESPACE env var not present: %v", err)
```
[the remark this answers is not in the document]
aaronfern: imo, better to fail early rather than continue with incorrect configuration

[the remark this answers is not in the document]
ishan16696: >imo, better to fail early rather than continue with incorrect configuration

I would say it depends on what priorities we set, in other words .... Do `etcd-member-gc` configurations are that important that we are ready to cause outage as backup-restore crash will also fails the readinessProbe?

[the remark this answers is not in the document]
aaronfern: I would imagine since `etcd-member-gc` can anyway be turned on or off using its flag, than it would be fine to crash backup-restore for a configuration issue
I however do not have a hard opinion on this, so if you feel this is a bit too strict, I could update this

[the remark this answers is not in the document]
ishan16696: hmm, so Let's keep it the `Fatalf` as [pkg/health/heartbeat/heartbeat.go](https://github.com/gardener/etcd-backup-restore/blob/322d3c74fcd5ea7be9961ce5e9d7ffed48ecc34e/pkg/health/heartbeat/heartbeat.go#L63-L69) is also throwing the `Fatalf` error

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+		podNamespace: namespace,
+		etcdName:     podName[:strings.LastIndex(podName, "-")], //TODO Evaluate if this will always be sufficient
+	}, nil
+}
+
+// RunMemberGarbageCollectorPeriodically periodically calls the member garbage collector
+func RunMemberGarbageCollectorPeriodically(ctx context.Context, hconfig *brtypes.HealthConfig, logger *logrus.Entry, etcdConfig *brtypes.EtcdConnectionConfig) {
+	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+	if err != nil {
+		logger.Errorf("failed to create kubernetes clientset: %v", err)
```
timuthy, line 88:
```suggestion
		mgc.logger.Errorf("failed to create kubernetes clientset: %v", err)
```

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+	mgc, err := NewMemberGarbageCollector(logger, clientSet)
+	if err != nil {
+		logger.Errorf("failed to initialize new etcd member garbage collector: %v", err)
```
timuthy, line 93:
```suggestion
		mgc.logger.Errorf("failed to initialize new etcd member garbage collector: %v", err)
```

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
```diff
@@ -0,0 +1,39 @@
...
+	err = os.RemoveAll(outputDir)
+	Expect(err).ShouldNot(HaveOccurred())
+
+	Expect(err).ShouldNot(HaveOccurred())
+	var data []byte
+	return data
+}, func(data []byte) {})
+
+var _ = SynchronizedAfterSuite(func() {}, func() {
+})
```
timuthy, line 39: Can we remove this?

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+var (
+	logger = logrus.New().WithField("suite", "snapshotter")
+	err    error
+)
+
+const (
+	outputDir = "../../../test/output"
+	etcdDir   = outputDir + "/default.etcd"
+)
```
timuthy, line 21: Is this needed?

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	logger.Logger.Out = io.Discard
+	err = os.RemoveAll(outputDir)
+	Expect(err).ShouldNot(HaveOccurred())
+
+	Expect(err).ShouldNot(HaveOccurred())
+	var data []byte
+	return data
```
timuthy, line 35: Is this needed?

timuthy (changes requested):
Only some nits left. Can you also update the release notes, esp. regarding the RBAC permissions that are needed by the garbage collector?

ishan16696 (changes requested):
Hi @aaronfern,
I have made couple of comments. Please address them!!

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: This error log is printed before the `mgc` object is created and so `mgc.logger` is not used

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Same as above. This error log is printed if there is an error creating the `mgc` object

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
aaronfern: Done

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
aaronfern: No, not needed. Removed

pkg/health/membergarbagecollector/membergarbagecollector_suite_test.go
aaronfern: Removed

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+	// Return if no etcd member present
+	if len(etcd.Status.Members) < 1 {
+		return &errors.EtcdError{
+			Message: "There are no members present in the etcd status. Nothing to clean",
+		}
+	}
+
+	etcdMemberListResponse, err := etcdCluster.MemberList(ctx)
```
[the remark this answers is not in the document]
aaronfern: Added a new context with timeout

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,171 @@
...
+
+	for _, member := range etcdMemberListResponse.Members {
+		memberPod := &v1.Pod{}
+		err := mgc.k8sClient.Get(ctx, client.ObjectKey{
+			Namespace: mgc.podNamespace,
+			Name:      member.Name,
+		}, memberPod)
+		if apierrors.IsNotFound(err) && !statusMembers[member.Name] {
+			//Remove member from etcd cluster if corresponding pod cannot be found and if member not present in etcd.Status.Members
+			if _, err = etcdCluster.MemberRemove(ctx, member.ID); err != nil {
```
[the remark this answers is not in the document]
aaronfern: Added new context

timuthy (changes requested):
As discussed, please try to derive the member state from the `StatefulSet` instead of using `etcd` which required us to have a cycle dependency between `etcd-druid` and `etcd-backup-restore`.

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,175 @@
...
+	listCtx, listCtxCancel := context.WithTimeout(ctx, mgc.etcdConnectionTimeout.Duration)
+	defer listCtxCancel()
+	etcdMemberListResponse, err := etcdCluster.MemberList(listCtx)
+	if err != nil {
+		return &errors.EtcdError{
+			Message: fmt.Sprintf("Could not list etcd cluster members: %v", err),
+		}
+	}
+
+	if stsSpecReplicas >= stsStatusReplicas && stsSpecReplicas >= int32(len(etcdMemberListResponse.Members)) {
```
timuthy, line 145:
```suggestion
	if sts.generation != sts.status.observedGeneration && stsSpecReplicas != sts.status.updatedReplicas && stsSpecReplicas >= int32(len(etcdMemberListResponse.Members)) {
```

Considering `sts.status.updatedReplicas` gives us a more confidence that no update is happening. WDYT?

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,175 @@
...
+			Name:      member.Name,
+		}, memberPod)
+		memberLease := &coordv1.Lease{}
+		leaseErr := mgc.k8sClient.Get(ctx, client.ObjectKey{
+			Namespace: mgc.podNamespace,
+			Name:      member.Name,
+		}, memberLease)
+		if (apierrors.IsNotFound(podErr) && apierrors.IsNotFound(leaseErr)) || member.Name == "" {
+			//Remove member from etcd cluster if corresponding pod cannot be found and if corresponding member lease is not found
+			//Or remove a learner member
```
timuthy, line 163: Hm why is removing a learner required here?

timuthy (changes requested):
Two last comments. Rest looks very good, thanks for taking care 🙂

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: There is a possibility that the superfluous member is still in the learner state and in that case `member.Name` might also be empty if it's etcd container never started

timuthy: Got it 👍 
Is it a good idea to skip the `apierrors.IsNotFound(podErr)` in such cases? This will be a race issue as there is always a short time gap between adding the learner and starting it.

aaronfern: Yes, I think we can skip the `IsNotFound` error as we will get a different error with an empty `member.Name`
I don't think we will have a race issue here as if we expect the learner to eventually start, we will not get past `stsSpecReplicas >= int32(len(etcdMemberListResponse.Members))` in line 145

timuthy: Fair enough. Thanks 🙂

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Yes, makes sense. Done

timuthy (approved):
/lgtm
Thank you very much!

timuthy:
@aaronfern kindly asking you to rebase so that we can proceed with this PR.

aaronfern:
/do-not-merge
Pending for local testing

timuthy (approved):
/lgtm
@ishan16696 do you also want to take a final look?

ishan16696:
>@ishan16696 do you also want to take a final look?
yes, I will take a look

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,174 @@
...
+		stsName:               podName[:strings.LastIndex(podName, "-")],
+		etcdConnectionTimeout: etcdTimeout,
+	}, nil
+}
+
+// RunMemberGarbageCollectorPeriodically periodically calls the member garbage collector
+func RunMemberGarbageCollectorPeriodically(ctx context.Context, hconfig *brtypes.HealthConfig, logger *logrus.Entry, etcdConfig *brtypes.EtcdConnectionConfig) {
+	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+	if err != nil {
+		logger.Errorf("failed to create kubernetes clientset: %v", err)
```
ishan16696, line 87: As we have seen that sometimes kubeapi-server wasn’t available within the timeout and due to that `clientSet` wasn’t able created successfully.
do you think log the error and return will cause any issue ?
Would it make sense to throw a `Fatal` error here but crashing the backup-restore might not be the good. WDYT?

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,174 @@
...
+	etcdclient "github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	podNamespace = "POD_NAMESPACE"
+	podName      = "POD_NAME"
+)
+
+// MGCChecker garbage collects the members from the etcd cluster if its corresponding pod is not present. (Used for unit tests only)
```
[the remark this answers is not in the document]
aaronfern: Updated
I hope it's fine now

pkg/health/membergarbagecollector/membergarbagecollector.go
aaronfern: Yup, makes sense
updated

pkg/server/backuprestoreserver.go
```diff
@@ -236,14 +241,15 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
-	//TODO @aaronfern: Add functionality for member garbage collection
 	if b.config.HealthConfig.MemberLeaseRenewalEnabled {
 		go func() {
 			if err := heartbeat.RenewMemberLeasePeriodically(ctx, mmStopCh, b.config.HealthConfig, b.logger, b.config.EtcdConnectionConfig); err != nil {
 				b.logger.Fatalf("failed RenewMemberLeases: %v", err)
 			}
 		}()
 	}
+	//start etcd member garbage collector
+	go membergarbagecollector.RunMemberGarbageCollectorPeriodically(ctx, b.config.HealthConfig, b.logger, b.config.EtcdConnectionConfig)
```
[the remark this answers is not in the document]

pkg/health/membergarbagecollector/membergarbagecollector.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	mgc.gcTimer = time.NewTimer(hconfig.MemberGCDuration.Duration)
+	defer mgc.gcTimer.Stop()
+	mgc.logger.Info("Started etcd member garbage collector")
+
+	//Create etcd client to get etcd cluster member list
+	clientFactory := etcdutil.NewFactory(*etcdConfig)
+	etcdCluster, err := clientFactory.NewCluster()
+	if err != nil {
+		mgc.logger.Errorf("Failed to create etcd cluster client: %v", err)
```
ishan16696, line 103:
```suggestion
		mgc.logger.Errorf("failed to create etcd cluster client: %v", err)
```
If we failed to create `etcdCluster` then I think it is better to return. WDYT?

ishan16696 (changes requested):
Overall looks good. few nitpicks

ishan16696 (approved):
LGTM!!

ishan16696:
Integration tests again went into this issue: https://github.com/gardener/etcd-backup-restore/issues/483
I will re-run the tests

aaronfern:
The `license/cla` step is still pending
Unable to resign

PR #406, opened 2021-11-19
Merged snapshot lease update function with snapshotting functions

aaronfern:
As of now taking a snapshot and updating the snapshot lease are performed in separate functions. 
These should ideally be part of a single process and the current approach poses a risk that a snapshot could be taken without explicitly updating the lease.
This PR merges these functions so that when either `takeDeltaSnapshotAndResetTimer` or `TakeFullSnapshotAndResetTimer` is called, the corresponding lease is updated as part of it.

Fixes #

**Release note**:

```improvement developer
NONE
```


pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -261,7 +261,7 @@ func (ssr *Snapshotter) closeEtcdClient() {
 
 // TakeFullSnapshotAndResetTimer takes a full snapshot and resets the full snapshot
 // timer as per the schedule.
```
timuthy, line 263:
```suggestion
// timer as per the schedule. Additionally, it updates the full snapshot lease if enabled.
```

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -364,7 +373,7 @@ func (ssr *Snapshotter) cleanupInMemoryEvents() {
 	ssr.lastEventRevision = -1
 }
 
-func (ssr *Snapshotter) takeDeltaSnapshotAndResetTimer() (*brtypes.Snapshot, error) {
+func (ssr *Snapshotter) takeDeltaSnapshotAndResetTimer(leaseUpdateCtx context.Context) (*brtypes.Snapshot, error) {
```
timuthy, line 376:
```suggestion
func (ssr *Snapshotter) takeDeltaSnapshotAndResetTimer(ctx context.Context) (*brtypes.Snapshot, error) {
```

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -518,13 +539,14 @@ func (ssr *Snapshotter) CollectEventsSincePrevSnapshot(stopCh <-chan struct{}) (
...
 		case <-stopCh:
+			leaseUpdateCancel()
 			ssr.cleanupInMemoryEvents()
 			return true, nil
 		}
 	}
 }
 
-func (ssr *Snapshotter) handleDeltaWatchEvents(wr clientv3.WatchResponse) error {
+func (ssr *Snapshotter) handleDeltaWatchEvents(leaseUpdateCtx context.Context, wr clientv3.WatchResponse) error {
```
timuthy, line 549:
```suggestion
func (ssr *Snapshotter) handleDeltaWatchEvents(ctx context.Context, wr clientv3.WatchResponse) error {
```

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -504,12 +523,14 @@ func (ssr *Snapshotter) CollectEventsSincePrevSnapshot(stopCh <-chan struct{}) (
 	metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull}).Set(1)
 
 	for {
+		leaseUpdateCtx, leaseUpdateCancel := context.WithCancel(context.TODO())
```
timuthy, line 526: I suggest to drop the creation of this new context and instead change the function:
```
func (ssr *Snapshotter) CollectEventsSincePrevSnapshot(stopCh <-chan struct{}) -> func (ssr *Snapshotter) CollectEventsSincePrevSnapshot(ctx context.Context)
```

It should be safe to pass the `leCtx` to this function then.

pkg/server/backuprestoreserver.go
```diff
@@ -220,18 +220,13 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 				// If the previous full snapshot doesn't exist or is not marked as final, take a final full snapshot
 				if ssr.PrevFullSnapshot == nil || !ssr.PrevFullSnapshot.IsFinal {
 					b.logger.Infof("Taking final full snapshot...")
-					var snapshot *brtypes.Snapshot
-					if snapshot, err = ssr.TakeFullSnapshotAndResetTimer(true); err != nil {
+					leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
+					defer cancel()
```
timuthy, line 224: If you create a very restricted context already here, then it is probably already expired when the update happens because the full snapshot can take a while. I think you can drop this context creation since you already create another on in `TakeFullSnapshotAndResetTimer`.

pkg/server/backuprestoreserver.go
```diff
@@ -300,21 +296,16 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 		}
 		if !initialDeltaSnapshotTaken {
 			// need to take a full snapshot here
-			var snapshot *brtypes.Snapshot
 			metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindDelta}).Set(0)
 			metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull}).Set(1)
-			if snapshot, err = ssr.TakeFullSnapshotAndResetTimer(false); err != nil {
+			leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
+			defer cancel()
```
timuthy, line 302: Same here: I think it's safe to remove the context creation.

timuthy (changes requested):
Looks good in general. I left some comments about the general context handling.

pkg/server/backuprestoreserver.go
```diff
@@ -269,7 +261,7 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 		const recentFullSnapshotPeriodInHours = 23.5
 		initialDeltaSnapshotTaken = false
 		if ssr.PrevFullSnapshot != nil && !ssr.PrevFullSnapshot.IsFinal && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= recentFullSnapshotPeriodInHours {
-			ssrStopped, err := ssr.CollectEventsSincePrevSnapshot(ssrStopCh)
+			ssrStopped, err := ssr.CollectEventsSincePrevSnapshot(ctx)
```
ishan16696, line 264: Is there any specific reasons to replace the `ssrStopCh` with `ctx`?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -271,6 +272,15 @@ func (ssr *Snapshotter) TakeFullSnapshotAndResetTimer(isFinal bool) (*brtypes.Sn
 		return nil, err
 	}
 
+	//Update Full snapshot lease
+	if ssr.healthConfig.SnapshotLeaseRenewalEnabled {
+		leaseUpdateCtx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
+		if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdateCtx, ssr.logger, s, ssr.K8sClientset, ssr.healthConfig.FullSnapshotLeaseName, ssr.healthConfig.DeltaSnapshotLeaseName); err != nil {
+			ssr.logger.Warnf("Snapshot lease update failed : %v", err)
+		}
+		cancel()
```
ishan16696, line 281: use `defer cancel()`

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -381,6 +391,16 @@ func (ssr *Snapshotter) takeDeltaSnapshotAndResetTimer() (*brtypes.Snapshot, err
...
 		ssr.deltaSnapshotTimer.Reset(ssr.config.DeltaSnapshotPeriod.Duration)
 	}
+
+	//Update delta snapshot lease
+	if ssr.healthConfig.SnapshotLeaseRenewalEnabled && s != nil {
+		leaseUpdateCtx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
+		if err = heartbeat.DeltaSnapshotCaseLeaseUpdate(leaseUpdateCtx, ssr.logger, ssr.K8sClientset, ssr.healthConfig.DeltaSnapshotLeaseName, ssr.store); err != nil {
+			ssr.logger.Warnf("Snapshot lease update failed : %v", err)
+		}
+		cancel()
```
ishan16696, line 401:
```suggestion
		defer cancel()
```

ishan16696 (changes requested):
I think it make sense to pickup this PR later after leader-election as leader-election brings lot of changes in code flow  after that I will take closer look specifically for of all `ctx` passing.

pkg/server/backuprestoreserver.go
timuthy: See https://github.com/gardener/etcd-backup-restore/pull/406#discussion_r754315829.

aaronfern:
This PR is quite old, and a lot of changes have been made since this PR was initially filed. This PR is also not vital to etcd-br operations
Hence I'm closing this PR. Please feel free to reopen it if anyone thinks otherwise 
/close

PR #409, opened 2021-12-13
Moved leaseUpdateCtx to not create the context multiple times

aaronfern:
This PR fixes a bug that sees the `leaseUpdateCtx` being created multiple times without being cancelled. This causes a memory leak without the contexts being cancelled

Fixes #

**Release note**:

```improvement operator
Took care of a bug where contexts were created multiple times without being cancelled
```


timuthy (approved):
/lgtm

PR #410, opened 2021-12-17
Renew snapshot leases even if snapshot skipped

aaronfern:
Currently snapshot lease updates are optimised in such a way that if revisions are not updated, the lease will not be renewed.
Going forward, the snapshot leases will be used as a heartbeat and so it would be desirable to renew them even if revisions have not been updated.
This PR does just that. It allows that heartbeat package to renew snapshot leases even if revisions are not updated

Fixes #

~~This PR is based on #406, hence lease ignore [`4a9496e`](https://github.com/gardener/etcd-backup-restore/commit/4a9496ee38a066f2714426e0ba17213fb44b03e7) and [`1fe44c5`](https://github.com/gardener/etcd-backup-restore/commit/1fe44c562c0d9d13ac7b3e484b24a82979f59304).~~
Reworked this PR to be standalone and hence can be reviewed and merged 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -192,7 +192,7 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
 		renewedLease := fullSnapLease.DeepCopy()
 		renewedTime := time.Now()
 		renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
-		// Update revisions in deltaSnapLease.Spec.HolderIdentity only when its value is less than latest deltaSnap.LastRevision
+		// Update revisions in fullSnapLease.Spec.HolderIdentity only when its value is less than latest fullSnap.LastRevision
 		if fullSnapLease.Spec.HolderIdentity == nil || rev < fullSnapshot.LastRevision {
 			actor := strconv.FormatInt(fullSnapshot.LastRevision, 10)
 			renewedLease.Spec.HolderIdentity = &actor
@@ -260,7 +260,7 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 			renewedLease.Spec.HolderIdentity = &actor
 			logString += " with revision " + actor
 		}
-	} else if len(prevDeltaSnapshots) == 0 && deltaSnapLease.Spec.HolderIdentity == nil {
+	} else if deltaSnapLease.Spec.HolderIdentity == nil {
 		// Special case: Set revisions in deltaSnapLease.Spec.HolderIdentity to 0 if no delta snaps taken at all
 		actor := strconv.FormatInt(0, 10)
 		renewedLease.Spec.HolderIdentity = &actor
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -260,6 +260,10 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 			renewedLease.Spec.HolderIdentity = &actor
 			logString += " with revision " + actor
 		}
+	} else if len(prevDeltaSnapshots) == 0 && deltaSnapLease.Spec.HolderIdentity == nil {
+		// Special case: Set revision number to 0 if no delta snaps taken at all
+		actor := strconv.FormatInt(0, 10)
+		renewedLease.Spec.HolderIdentity = &actor
 	}
 
 	err = k8sClientset.Patch(ctx, renewedLease, client.MergeFrom(deltaSnapLease))
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -192,7 +192,7 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
 		renewedLease := fullSnapLease.DeepCopy()
 		renewedTime := time.Now()
 		renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
-		//Update revision number only if revisions in existing lease is lower
+		// Update revisions in deltaSnapLease.Spec.HolderIdentity only when its value is less than latest deltaSnap.LastRevision
 		if fullSnapLease.Spec.HolderIdentity == nil || rev < fullSnapshot.LastRevision {
 			actor := strconv.FormatInt(fullSnapshot.LastRevision, 10)
 			renewedLease.Spec.HolderIdentity = &actor
@@ -245,7 +245,7 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 	renewedTime := time.Now()
 	renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
 
-	//Update revision number only if revisions in existing lease is lower
+	// Update revisions in deltaSnapLease.Spec.HolderIdentity only when its value is less than latest deltaSnap.LastRevision
 	if len(prevDeltaSnapshots) > 0 {
 		deltaSnap := prevDeltaSnapshots[len(prevDeltaSnapshots)-1]
 		var deltaRev int64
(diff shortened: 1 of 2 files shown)


ishan16696:
>Going forward, the snapshot leases will be used as a heartbeat

How the `snapshot leases` can be used as heartbeat can you please explain? 
In multi-node etcd, snapshot will be taken by `backup-leader` so the snapshot leases can be used as heartbeat only for leader but not for `backup-restore` followers.

aaronfern:
Here, I don't mean a member heartbeat, but rather a heartbeat for the snapshotter or the backup component as a whole. The snapshotter can be in a failing state without the pod also failing.
The snapshot leases would be used as a heartbeat just for the `BackupReady` condition for the time being

ishan16696:
@aaronfern thanks for explaining

aaronfern:
PR is ready for review as it now has no dependencies on any other PRs

pkg/health/heartbeat/heartbeat.go
```diff
@@ -241,17 +242,33 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
...
+	if len(prevDeltaSnapshots) > 0 {
+		deltaSnap := prevDeltaSnapshots[len(prevDeltaSnapshots)-1]
+		var deltaRev int64
+		if deltaSnapLease.Spec.HolderIdentity != nil {
+			deltaRev, err = strconv.ParseInt(*deltaSnapLease.Spec.HolderIdentity, 10, 64)
+			if err != nil {
+				return err
+			}
+		}
+		if deltaSnapLease.Spec.HolderIdentity == nil || deltaRev < deltaSnap.LastRevision {
```
ishan16696, line 258: `deltaSnapLease.Spec.HolderIdentity == nil`
This condition is present inside `len(prevDeltaSnapshots)>0{....}` and if `len(prevDeltaSnapshots)>0` which means DeltaSnaposhots present then `deltaSnapLease.Spec.HolderIdentity` can't be nil. right ?

ishan16696 (changes requested):
Hi @aaronfern,
Overall PR looks good, just 2 nitpicks!!

pkg/health/heartbeat/heartbeat.go
aaronfern: The only time `HolderIdentity` can be nil when `len(prevDeltaSnapshots)>0`, is for a fresh cluster when the first delta snap is uploaded and we update the delta lease for the first time
`deltaSnapLease.Spec.HolderIdentity == nil` was meant for this condition

ishan16696: hmm, make sense thanks

pkg/health/heartbeat/heartbeat.go
```diff
@@ -241,17 +242,33 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 	}
 
 	renewedLease := deltaSnapLease.DeepCopy()
-	renewedLease.Spec.HolderIdentity = &revisionNumber
 	renewedTime := time.Now()
 	renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
 
+	//Update revision number only if revisions in existing lease is lower
+	if len(prevDeltaSnapshots) > 0 {
```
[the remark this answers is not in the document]
aaronfern: Added changes to set `HolderIdentity` of delta snapshot lease only if there are no previous delta snapshots and `HolderIdentity` is nil
PTAL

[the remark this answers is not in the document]
ishan16696: thanks for addressing this issue, I saw you have updated the lease with `deltaSnapLease.Spec.HolderIdentity=0` when `len(prevDeltaSnapshots) == 0 && deltaSnapLease.Spec.HolderIdentity == nil` .... Can you also check that this will not cause any issue for compaction job calculating no of events.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -241,17 +242,37 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 	}
 
 	renewedLease := deltaSnapLease.DeepCopy()
-	renewedLease.Spec.HolderIdentity = &revisionNumber
 	renewedTime := time.Now()
 	renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
 
+	//Update revision number only if revisions in existing lease is lower
```
ishan16696, line 248:
```suggestion
	//Update deltaSnapLease.Spec.HolderIdentity only when its value is less than latest deltaSnap.LastRevision
```
can you please update this comment with something similar or better.

pkg/health/heartbeat/heartbeat_test.go
```diff
@@ -2,10 +2,11 @@ package heartbeat_test
 
 import (
 	"context"
-	"k8s.io/utils/pointer"
 	"os"
 	"time"
 
+	//"k8s.io/utils/pointer"
```
ishan16696, line 8: can you remove this.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -175,28 +175,35 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
...
 				return nil
 			}
 		}
 
 		renewedLease := fullSnapLease.DeepCopy()
-		actor := strconv.FormatInt(fullSnapshot.LastRevision, 10)
-		renewedLease.Spec.HolderIdentity = &actor
 		renewedTime := time.Now()
 		renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
+		//Update revision number only if revisions in existing lease is lower
```
ishan16696, line 195: It is not updating the revision , it is updating the lease spec HolderIdentity. 
Can you update this comment.

ishan16696 (changes requested):
Hi @aaronfern ,
I have tested the PR with `multi-node etcd` environment it is working fine. 
Just couple more nitpicks !!!

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
aaronfern: As can be seen from [this](https://github.com/gardener/etcd-druid/blob/0469cc9a8c1e50c5aee8b08fcc632b89b9278366/controllers/compaction_lease_controller.go#L137) point onwards in the `compaction-lease-controller`, currently the controller will return at this point.
However, if delta lease `HolderIdentity` is set to 0, it will then use that while calculating revision difference which will also be an accurate result

pkg/health/heartbeat/heartbeat.go
aaronfern: Addressed all comments. PTAL

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
ishan16696: make sense, thanks

ishan16696 (approved):
LGTM!!

pkg/health/heartbeat/heartbeat.go
```diff
@@ -175,28 +175,35 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
...
 				return nil
 			}
 		}
 
 		renewedLease := fullSnapLease.DeepCopy()
-		actor := strconv.FormatInt(fullSnapshot.LastRevision, 10)
-		renewedLease.Spec.HolderIdentity = &actor
 		renewedTime := time.Now()
 		renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
+		// Update revisions in deltaSnapLease.Spec.HolderIdentity only when its value is less than latest deltaSnap.LastRevision
```
timuthy, line 195:
```suggestion
		// Update revisions in fullSnapLease.Spec.HolderIdentity only when its value is less than latest fullSnapshot.LastRevision
```

pkg/health/heartbeat/heartbeat.go
```diff
@@ -241,17 +242,37 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
...
+			if err != nil {
+				return err
+			}
+		}
+		if deltaSnapLease.Spec.HolderIdentity == nil || deltaRev < deltaSnap.LastRevision {
+			actor := strconv.FormatInt(deltaSnap.LastRevision, 10)
+			renewedLease.Spec.HolderIdentity = &actor
+			logString += " with revision " + actor
+		}
+	} else if len(prevDeltaSnapshots) == 0 && deltaSnapLease.Spec.HolderIdentity == nil {
```
timuthy, line 263: Nit
```suggestion
	} else if deltaSnapLease.Spec.HolderIdentity == nil {
```

timuthy (changes requested):
From what I can tell it looks good, thanks for the changes. I added two minor comments.

pkg/health/heartbeat/heartbeat.go
aaronfern: Done

pkg/health/heartbeat/heartbeat.go
aaronfern: Done

timuthy (approved):

PR #411, opened 2021-12-27
Decouple the readinessProbe of etcd and Snapshotter

### LINKED ISSUE #325: [Feature] Snapshotter should not proactively cut off traffic to etcd on backup failures
Currently, snapshotter sets etcd readiness probe to false when snapshotter fails, thus cutting off traffic to etcd until snapshotter becomes healthy again. There is a need to modify this behavior to allow etcd to serve traffic even if backups are failing, but raise a critical issue for the garden landscape operator to act upon at the earliest.  

This will reduce induced etcd downtimes caused by backup failures due to probable object storage or object storage network connectivity issues.

- Add new blanket metric for snapshotter failures
- Add a critical alert on https://github.com/gardener/gardener to alert garden operators 
- Change behavior of backup-restore by not setting http /healthz status to unavailable

ishan16696:
Currently, [snapshotter fails the readiness probe of etcd](https://github.com/gardener/etcd-backup-restore/blob/b6adfb0d37db26f48f6e33d5984f92068caa6675/pkg/server/backuprestoreserver.go#L391-L403) if it fails to take/upload the backup, thus cutting off incoming traffic to etcd until snapshotter becomes healthy again.
This PR decouples the [readinessProbe of etcd](https://github.com/gardener/etcd-druid/blob/master/charts/etcd/templates/etcd-statefulset.yaml#L55-L60) and snapshotter health of backup-restore.So, Backup-restore's [snapshotter failure won't fails the readinessProbe of etcd](https://github.com/gardener/etcd-backup-restore/blob/b6adfb0d37db26f48f6e33d5984f92068caa6675/pkg/server/backuprestoreserver.go#L403) and hence etcd will allow to keep serving the incoming traffic.
This PR also introduced a [Exponential-Backoff mechanism](https://github.com/gardener/etcd-backup-restore/pull/411/files?diff=unified&w=0#diff-bc7959fe46152e1e2795507b716dc9b825f1ffac96a2d8a5bfb73cf082e1ff71R21-R33) in Snapshotter(if Snapshotter fails.)

Fixes #325 

**Release note**:
```improvement operator
Failed backups won't fails the readinessProbe thus etcd is allowed to keep serving the incoming traffic.
```
```improvement operator
Added CLI flags `--backoff-multiplier`, `--backoff-attempt-limit` and `--backoff-threshold-time` to configure the exponential-backoff mechanism.
```


ishan16696:
Pipeline integration test had some issue .... I will force push again to re-run them. Meanwhile I have run the clusterBasedIntegrationTest Locally and they successfully passed. [clusterBasedIntegrationTest_DecoupleReadinessProbe.log](https://github.com/gardener/etcd-backup-restore/files/7819652/clusterBasedIntegrationTest_DecoupleReadinessProbe.log)

pkg/server/backuprestoreserver.go
```diff
@@ -72,12 +74,15 @@ func NewBackupRestoreServer(logger *logrus.Logger, config *BackupRestoreComponen
...
 	}
+	backoffExponentialConfig := brtypes.NewExponentialBackOff()
+
 	return &BackupRestoreServer{
 		logger:                  serverLogger,
 		config:                  config,
 		ownerChecker:            ownerChecker,
 		etcdProcessKiller:       etcdProcessKiller,
 		defragmentationSchedule: defragmentationSchedule,
+		backoff:                 backoffExponentialConfig,
```
abdasgupta, line 85: Should not we name it like `backoffConfig` ?

pkg/server/backuprestoreserver.go
```diff
@@ -196,6 +201,9 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 					b.logger.Info("backup-restore stops leading...")
 				}
 				handler.SetSnapshotterToNil()
+
+				// To do: For Multi-node etcd HTTP status need to be set to `StatusServiceUnavailable` only when backup-restore is in "StateUnknown".
```
abdasgupta, line 205: Why don't you implement the TODO now? Single node should never be `StateUnknown`, right? So, if single node is somehow in `StateUnknown` we can safely set HTTP status for StatusServiceUnavailable . Or am I missing something?

pkg/types/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+	// defaultThresholdAttempt defines default no. of attempts before a ThresholdTime is achieved.
+	defaultThresholdAttempt = 6
+	// defaultThresholdTime defines default Threshold Backoff time used when threshold is achieved.
+	defaultThresholdTime time.Duration = 128 * time.Second
+	// DefaultMinimunBackoff defines default minimum Backoff time.
+	DefaultMinimunBackoff time.Duration = 1 * time.Second
+)
+
+// ExponentialBackoff holds the configuration for the Backoff mechanism.
+type ExponentialBackoff struct {
```
abdasgupta, line 35: We should name it something like `ExponentialBackoffConfig`

abdasgupta (changes requested):
Overall good PR. Just few recommendations.

pkg/server/backuprestoreserver.go
ishan16696: >Single node should never be StateUnknown, right?

actually no, when `single node etcd` goes down/crash then backup-restore can go to `StateUnknown` from `StateLeader`.

> Why don't you implement the TODO now? 

yeah, ideally setting the HTTP status to `StatusServiceUnavailable` should be [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/leaderelection/leaderelection.go#L87-L105) when backup-restore goes to `StateUnknown` from any state but to do this [LeaderElector struct](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/leaderelection/leaderelection.go#L46-L55)  should have [HTTP Handler object](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L68). I wasn’t sure to do this, so for time being and for single node etcd I choose to go with above method since  single node `etcd` backup-restore can only be either in `StateLeader ` or in `StateUnknown`.

pkg/server/backuprestoreserver.go
ishan16696: I haven't made the [ExponentialBackoff struct](https://github.com/gardener/etcd-backup-restore/pull/411/files#diff-ff42c8f8fcbf5c49ef29bde3eced688de7d922868435ce048b6d40c049c6ebd8R35)  configurable as I thought in backup-restore we are already having too many configurable flags. 
so, Should I change the name from `backoff` to `backoffConfig`? Isn't the `backoffConfig` is bit misleading ? WDYT?

pkg/types/backoff.go
ishan16696: same as https://github.com/gardener/etcd-backup-restore/pull/411#discussion_r781102141

pkg/server/backuprestoreserver.go
abdasgupta: I think, `backoffConfig` is fine. Config means `configuration` here.  `backoff` sounds like a command

pkg/server/backuprestoreserver.go
abdasgupta: Okay. Got it. Thanks.

pkg/server/backuprestoreserver.go
ishan16696: ok, I will make the necessary  changes.

ishan16696: Addressed. PTAL

pkg/server/backuprestoreserver.go
```diff
@@ -72,12 +74,15 @@ func NewBackupRestoreServer(logger *logrus.Logger, config *BackupRestoreComponen
 		// Ideally this case should not occur, since this check is done at the config validaitions.
 		return nil, err
 	}
+	backoffExponentialConfig := brtypes.NewExponentialBackOffConfig()
```
shreyas-s-rao, line 77: Rename to `exponentialBackoffConfig` to have consistent naming

pkg/types/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// ExponentialBackoffConfig holds the configuration for the ExponentialBackoff mechanism.
+type ExponentialBackoffConfig struct {
+	Start              bool              `json:"start,omitempty"`
+	Multiplier         int               `json:"multiplier,omitempty"`
+	Attempt            uint64            `json:"attempt,omitempty"`
+	ThresholdAttempt   uint64            `json:"attemptLimit,omitempty"`
+	ThresholdTime      wrappers.Duration `json:"thresholdTime,omitempty"`
+	CurrentBackoffTime wrappers.Duration `json:"currentBackoff,omitempty"`
```
shreyas-s-rao, line 41: Please add description comments for each struct member

pkg/types/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+	// DefaultMinimunBackoff defines default minimum Backoff time.
+	DefaultMinimunBackoff time.Duration = 1 * time.Second
+)
+
+// ExponentialBackoffConfig holds the configuration for the ExponentialBackoff mechanism.
+type ExponentialBackoffConfig struct {
+	Start              bool              `json:"start,omitempty"`
+	Multiplier         int               `json:"multiplier,omitempty"`
+	Attempt            uint64            `json:"attempt,omitempty"`
+	ThresholdAttempt   uint64            `json:"attemptLimit,omitempty"`
```
shreyas-s-rao, line 39: Rename to `AttemptsThreshold` or `AttemptsLimit` and change the json tag accordingly.

pkg/types/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+
+const (
+	// defaultBackoffMultiplier defines default multiplier for BackOff-Exponential mechanism.
+	defaultBackoffMultiplier = 2
+	// defaultThresholdAttempt defines default no. of attempts before a ThresholdTime is achieved.
+	defaultThresholdAttempt = 6
+	// defaultThresholdTime defines default Threshold Backoff time used when threshold is achieved.
+	defaultThresholdTime time.Duration = 128 * time.Second
+	// DefaultMinimunBackoff defines default minimum Backoff time.
+	DefaultMinimunBackoff time.Duration = 1 * time.Second
```
shreyas-s-rao, line 31: Typo: `DefaultMinimumBackoff`

pkg/types/backoff.go
```diff
@@ -0,0 +1,53 @@
...
+type ExponentialBackoffConfig struct {
+	Start              bool              `json:"start,omitempty"`
+	Multiplier         int               `json:"multiplier,omitempty"`
+	Attempt            uint64            `json:"attempt,omitempty"`
+	ThresholdAttempt   uint64            `json:"attemptLimit,omitempty"`
+	ThresholdTime      wrappers.Duration `json:"thresholdTime,omitempty"`
+	CurrentBackoffTime wrappers.Duration `json:"currentBackoff,omitempty"`
+}
+
+// NewExponentialBackOffConfig creates and returns new ExponentialBackoff.
```
shreyas-s-rao, line 44:
```suggestion
// NewExponentialBackOffConfig creates and returns a new ExponentialBackoffConfig.
```

pkg/server/backuprestoreserver.go
shreyas-s-rao, line 205:
```suggestion
				// TODO: @ishan16696  For Multi-node etcd HTTP status need to be set to `StatusServiceUnavailable` only when backup-restore is in "StateUnknown".
```
This makes `TODO`s easily trackable. Can you also please raise an issue for this if necessary (so that it's not forgotten later)? Thanks.

shreyas-s-rao (changes requested):
@ishan16696 Thanks for your PR. I have a few comments/suggestions. please address them:

1. It is generally advisable to keep config objects read-only. I see that in `getNextBackoffTime()` and `resetBackoffExponential()`, you are editing the config's member values like `Attempt` and `CurrentBackoffTime`. It's better to separate the "config" from the "working object". The config can have members like `AttemptsThreshold`, `Multiplier` and `TimeThreshold`, while you can have another object called `ExponentialBackoff` with method `getNextBackoffTime()` and variables `CurrentAttempt` and `CurrentBackoffTime`. In this case, `resetBackoffExponential()` can be completely eliminated because you will never edit the config, and for every time a backoff is initiated, you will create a new `ExponentialBackoff` object for it and later destroy it.

2. Also, please document the backoff method somewhere in `pkg/types/backoff.go` - what the variables mean, and how they interact with each other to make the backoff mechanism work. You could even think of a separate package for the `backoff` mechanism if it makes sense.

3. Does it make sense to expose the backoff config as a configurable flag for etcdbrctl? After all, it's a "config" right? :P

ishan16696:
Hi @shreyas-s-rao,
I have added a new pkg for `backoff mechanism` and also expose the backoff config as a configurable flags. 
I have also updated the release note and addressed all of your comments. PTAL !!

pkg/server/backuprestoreserver.go
ishan16696: created this issue https://github.com/gardener/etcd-druid/issues/288 describing the readinessProbe for multi-node etcd.

pkg/backoff/exponentialbackoff_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"time"
+
+	"github.com/gardener/etcd-backup-restore/pkg/backoff"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("OwnerChecker", func() {
```
shreyas-s-rao, line 25: Rename the Describe function.

shreyas-s-rao (approved):
@ishan16696 Thanks for making the changes! Except for one small nitpick, LGTM 👍🏼

abdasgupta (approved):
Thanks. Looks good.

PR #431, opened 2022-02-08
Add support for custom endpoints in S3 snapstore

### LINKED ISSUE #416: Configure custom S3 endpoint
Hi, 
Is there a way to configure etcdbrctl with a custom **S3 endpoint** for snapshot/backup storage? 

I understand that the chosen path is to configure the AWS S3 auth keys and let it connect to your bucket, I'm struggling into connecting the backup utility to a custom s3/minio storage because there is no option to configure the endpoint via [ ~/.aws/credentials ] nor [ ~/.aws/conf ]. 

AWS cli allows you to set the **--endpoint-url** flag with a custom endpoint of your own, p.e.:
`aws --endpoint-url https://my.full.domain:9000 s3 mb s3://dummybucket`

..but there is no option to bypass the etcdbrctl tool and add this flag.

Greets and many thanks in advance!

sibucan:
Adds the capability to specify a different S3 endpoint within the S3 snapstore logic. This allows usage of etcd-backup-restore with non-AWS S3-compliant providers, such as [min.io](https://min.io/) or [Linode Object Storage](https://www.linode.com/products/object-storage/). It also adds a way of setting `S3ForcePathStyle`, which is needed for etcd-backup-restore to be able to save the snapshot in these S3-compatible providers. 

Note that by default when using the AWS SDK client, providing an empty string in the Endpoint field of the `aws.Config{}` struct is functionally similar to what was being done before (which was keeping it nil), and will generate the default endpoint, see: https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.22.0/aws?utm_source=gopls#Config

Last but not least, I understand there's at least [one PR ](https://github.com/gardener/etcd-backup-restore/pull/428)that may modify the logic in this one. I am willing to rebase this PR if necessary if that one gets merged first.

Fixes #416

**Release note**:

```improvement operator
* Added support for non-AWS S3-compatible providers by specifying a custom endpoint.
```

### DIFF
--- a/pkg/snapstore/s3_snapstore.go
+++ b/pkg/snapstore/s3_snapstore.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -45,6 +46,8 @@ const (
 	awsSecretAccessKey          = "AWS_SECRET_ACCESS_KEY"
 	awsAcessKeyID               = "AWS_ACCESS_KEY_ID"
 	awsRegion                   = "AWS_REGION"
+	awsEndpoint                 = "AWS_ENDPOINT"
+	awsForcePathStyle           = "AWS_FORCE_PATH_STYLE"
 	awsCredentialFile           = "AWS_APPLICATION_CREDENTIALS"
 	awsCredentialJSONFile       = "AWS_APPLICATION_CREDENTIALS_JSON"
 )
@@ -88,12 +91,18 @@ func newS3FromSessionOpt(bucket, prefix, tempDir string, maxParallelChunkUploads
 
 func getSessionOptions(prefixString string) (session.Options, error) {
 
+	var s3path *bool
+	if val, err := strconv.ParseBool(os.Getenv(awsForcePathStyle)); err != nil {
+		s3path = &val
+	}
 	// TODO: passing credentials through environment variable will be deprecated by "backup-restore v0.18.0"
 	if _, isSet := os.LookupEnv(prefixString + awsAcessKeyID); isSet {
 		return session.Options{
 			Config: aws.Config{
-				Credentials: credentials.NewStaticCredentials(os.Getenv(prefixString+awsAcessKeyID), os.Getenv(prefixString+awsSecretAccessKey), ""),
-				Region:      pointer.StringPtr(os.Getenv(awsRegion)),
+				Credentials:      credentials.NewStaticCredentials(os.Getenv(prefixString+awsAcessKeyID), os.Getenv(prefixString+awsSecretAccessKey), ""),
+				Region:           pointer.StringPtr(os.Getenv(awsRegion)),
+				Endpoint:         pointer.StringPtr(os.Getenv(awsEndpoint)),
+				S3ForcePathStyle: s3path,
 			},
 		}, nil
 	}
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-2:
Thank you @sibucan for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/snapstore/s3_snapstore.go
```diff
@@ -73,11 +76,14 @@ func newS3FromSessionOpt(bucket, prefix, tempDir string, maxParallelChunkUploads
 }
 
 func getSessionOptions(prefixString string) session.Options {
+	s3path, _ := strconv.ParseBool(os.Getenv(awsForcePathStyle))
```
timuthy, line 79:
```suggestion
        var s3path *bool
	if val, err := strconv.ParseBool(os.Getenv(awsForcePathStyle)); err != nil {
	  s3pathPtr = &val
	}
```
To keep `S3ForcePathStyle` nil as it used to be before + we have a slightly improved error handling.

timuthy (changes requested):
Thanks for the contribution. One comment from my side, the rest looks good to me.

pkg/snapstore/s3_snapstore.go
sibucan: My code looked somewhat like that, initially! 😅 However, I felt more confident about changing it to the way that it looks now after I looked at the `ParseBool()` source code (https://cs.opensource.google/go/go/+/refs/tags/go1.16:src/strconv/atob.go;l=10).

If the env var doesn't have any value (which is the default configuration when connecting to Amazon S3), then the function will return `false` and an error. The `false` is what I _really_ want from that, and the error doesn't really mean anything to me, since I expected it in that situation.

I think that if I were trying to give feedback to the user about the config being wrong, the error would be somewhat useful, but I don't think that's necessary in this case, and would require a much larger refactor than what's shown here.

sibucan: If you'd like, I can leave a comment on that line, explaining that intent to make it less confusing in the future.

timuthy: TBH, I don't see any benefits having a one liner + any comments.
With the suggestion above we:
- Pass `nil` like we do today if the env var is not set (-> no regression guarantee).
- We handle the error to some extent by keeping the `nil` value.
- One should not first look at the `ParseBool` implementation to figure out the first returned value but instead handle the error if one is returned as this is a very common pattern (see [Effective Go](https://go.dev/doc/effective_go) -> Multiple return values)

sibucan: Sounds good to me, I tested it locally and it worked. 👍

timuthy:
@ishan16696 are you fine with rebasing #428, so that we can merge this one first?

ishan16696:
I think we should go with https://github.com/gardener/etcd-backup-restore/pull/435 first then https://github.com/gardener/etcd-backup-restore/pull/436 followed by this PR as this PR also required to update the doc as well as Plamen and I don’t have to retest  https://github.com/gardener/etcd-backup-restore/pull/435  PR with 5  different storage providers.

timuthy:
@sibucan kindly asking you to rebase this PR 🙂

ishan16696:
@sibucan can you also update the [document for storage providers](https://github.com/gardener/etcd-backup-restore/blob/master/doc/usage/getting_started.md#various-ways-to-pass-credentials) accordingly.
Thanks !!

sibucan:
@timuthy I've rebased and updated the document as requested -- one thing to note is that I have not changed my code so that the AWS endpoint can be configured via a file, since this functionality is freshly-merged and I wasn't sure if my changes should also follow suit (meaning they should also be configurable via a file). I guess it can also be added in a subsequent PR if deemed necessary.

timuthy (approved):
/lgtm

PR #438, opened 2022-02-16
Don't hide error when creating K8s client

timuthy:
Instead of returning a new error, we should return the original one to not cover important information that can help to solve the issue.

**Release note**:

```improvement operator
NONE
```


ishan16696 (approved):
thanks for PR 
LGTM!!

PR #444, opened 2022-02-23
OwnerCheck shouldn't fails the ReadinessProbe right away.

### LINKED ISSUE #437: [Feature] OwnerCheck shouldn't fails the ReadinessProbe right away.
Currently,  when the [OwnerCheck](https://github.com/gardener/etcd-backup-restore/blob/03920151cfeeb50d6e8a35c7de2aa1f7c6670625/pkg/common/ownerchecker.go#L43) fails it sets the [HTTP status to StatusServiceUnavailable](https://github.com/gardener/etcd-backup-restore/blob/03920151cfeeb50d6e8a35c7de2aa1f7c6670625/pkg/server/backuprestoreserver.go#L267) right away which can leads to outages as it fails the `ReadinessProbe` of etcd. So, we need to provide some delay or backoff-mechanism such that OwnerCheck shouldn't fails the ReadinessProbe of `etcd` right away.

OwnerCheck shouldn't fails the ReadinessProbe until we are sure that OwnerCheck has failed because OwnerCheck can also fails due to DNS throttling or network delays.

ishan16696:
1. Added a [backoff mechanism](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/backoff/exponentialbackoff.go#L21-L33) with failureThreshold to confirm ownerChecks failure.
2. Added a CLI flag `--owner-check-failure-threshold` to define no. of attempts to confirm owner checks failure.

Fixes #437 

**Release note**:
```improvement operator
Added a Backoff exponential mechanism to confirm the owner checks failure.
```
```improvement operator
Added a CLI flag `--owner-check-failure-threshold`, `--owner-check-backoff-multiplier` and `--owner-check-backoff-threshold-time`
```

### DIFF
--- a/pkg/common/checkeractionwatchdog.go
+++ b/pkg/common/checkeractionwatchdog.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/gardener/etcd-backup-restore/pkg/backoff"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/clock"
 )
@@ -42,6 +43,8 @@ type Watchdog interface {
 	Start(ctx context.Context)
 	// Stop stops the goroutine started by Start.
 	Stop()
+	// Confirm gives the confirmation about checkerActionWatchdog check.
+	Confirm(context.Context, uint, *backoff.ExponentialBackoff) bool
 }
 
 // NewCheckerActionWatchdog creates a new Watchdog that regularly checks if the condition checked by the given checker
@@ -90,6 +93,36 @@ func (w *checkerActionWatchdog) Start(ctx context.Context) {
 			}
 		}
 	}()
+}
+
+// Confirm gives the confirmation that condition checked by the watchdog has failed or not.
+// It returns the boolean.
+func (w *checkerActionWatchdog) Confirm(ctx context.Context, failureThreshold uint, backoff *backoff.ExponentialBackoff) bool {
+	w.logger.Debug("Starting watchdog confirm")
+	defer w.logger.Debug("Stopping watchdog confirm")
+
+	ctx, w.cancelFunc = context.WithCancel(ctx)
+	var watchdogChecksFailCount uint = 0
+
+	for {
+		select {
+		case <-ctx.Done():
+			return false
+		case <-w.clock.After(backoff.GetNextBackoffTime()):
+			result, err := w.checker.Check(ctx)
+			if err != nil || !result {
+				watchdogChecksFailCount++
+				w.logger.Debugf("watchdog ChecksFailCount: %v", watchdogChecksFailCount)
+				if watchdogChecksFailCount >= failureThreshold {
+					w.logger.Info("watchdog check fails: confirm")
+					return true
+				}
+			} else {
+				w.logger.Info("watchdog check fails: not confirm")
+				return false
+			}
+		}
+	}
 }
 
 // Stop stops the goroutine started by Start.
--- a/pkg/common/checkeractionwatchdog.go
+++ b/pkg/common/checkeractionwatchdog.go
@@ -98,27 +98,26 @@ func (w *checkerActionWatchdog) Start(ctx context.Context) {
 // Confirm gives the confirmation that condition checked by the watchdog has failed or not.
 // It returns the boolean.
 func (w *checkerActionWatchdog) Confirm(ctx context.Context, failureThreshold uint, backoff *backoff.ExponentialBackoff) bool {
-	w.logger.Debug("Starting watchdog confirm")
-	defer w.logger.Debug("Stopping watchdog confirm")
+	w.logger.Info("Starting watchdog confirm")
+	defer w.logger.Info("Stopping watchdog confirm")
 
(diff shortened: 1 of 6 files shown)


pkg/common/checkeractionwatchdog.go
```diff
@@ -92,6 +95,36 @@ func (w *checkerActionWatchdog) Start(ctx context.Context) {
 	}()
 }
 
+// Confirm gives the confirmation that condition checked by the watchdog has failed or not.
+// It returns the boolean.
+func (w *checkerActionWatchdog) Confirm(ctx context.Context, failureThreshold uint, backoff *backoff.ExponentialBackoff) bool {
+	w.logger.Debug("Starting watchdog confirm")
+	defer w.logger.Debug("Stopping watchdog confirm")
+
+	ctx, w.cancelFunc = context.WithCancel(ctx)
```
timuthy, line 104: Please don't overwrite the `cancelFunc` here because it rather belongs to the watchdog's context which is closed by the `Stop()` function. I think it's sufficient if you directly use the context `ctx` that is passed.

timuthy (changes requested):

pkg/common/checkeractionwatchdog.go
ishan16696: yes, it’s just I forgot to remove

pkg/server/backuprestoreserver.go
```diff
@@ -318,6 +314,10 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 				}
 			}
 
+			// set server's healthz endpoint status to OK so that
+			// etcd is marked as ready to serve traffic
+			handler.SetStatus(http.StatusOK)
```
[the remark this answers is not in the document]
ishan16696: just wanted to set the readinessProbe only after `ownercheck` succeed/fail . Previously I missed to test this [PR](https://github.com/gardener/etcd-backup-restore/pull/411) with ownercheck.

plkokanov (commented):
Generally I'm a bit worried that the Confirm method will block the loop in `runEtcdProbeLoopWithSnapshotter()` and that it's called after we've already stopped the snapshotter, which means that no snapshots will be taken while waiting for the owner confirmation to finish.

Shouldn't we be using the Confirm() method inside the ownerCheckWatchdog here: 
https://github.com/gardener/etcd-backup-restore/blob/0ef71737084a13946eb0430da017f127c0892011/pkg/common/checkeractionwatchdog.go#L79

timuthy:
> Shouldn't we be using the Confirm() method inside the ownerCheckWatchdog here:

Also a good idea 👍

ishan16696:
>Shouldn't we be using the Confirm() method inside the ownerCheckWatchdog here:

make sense

pkg/common/checkeractionwatchdog.go
```diff
@@ -39,9 +40,11 @@ func (f ActionFunc) Do(ctx context.Context) {
 // Watchdog manages a goroutine that can be started and stopped.
 type Watchdog interface {
 	// Start starts a goroutine using the given context.
-	Start(ctx context.Context)
+	Start(context.Context, uint, *backoff.ExponentialBackoff)
 	// Stop stops the goroutine started by Start.
 	Stop()
+	// Confirm gives the confirmation about checkerActionWatchdog check.
+	Confirm(context.Context, uint, *backoff.ExponentialBackoff) bool
```
timuthy, line 47: Do we really need to expose this function? It can also be a private function in the `Watchdog` implementation, right?

timuthy (commented):
Only one nit, the rest looks good to me. Thank you 👍

pkg/server/backuprestoreserver.go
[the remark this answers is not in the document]
timuthy: Okay, thanks for explaining and testing.

pkg/common/checkeractionwatchdog.go
ishan16696: Actually I have also added the unit tests  which required to call `Confirm` from different pkg, that’s why it’s public.
check [here](https://github.com/gardener/etcd-backup-restore/pull/444/files?diff=unified&w=0#diff-ca1dc33ca57f024b5c11aa85848011b962f18bfb602046f3786f6b25676e7261R123)

timuthy:
/hold
There is currently a discussion with @plkokanov and @vlerenc that we can drop that retry mechanism and instead only kill the etcd process if a **different owner is detected**. It should then continue to run in case of a DNS query error.

timuthy:
@stoyanr @plkokanov kindly ask you to confirm if you are fine with the the latest suggestion https://github.com/gardener/etcd-backup-restore/pull/444#issuecomment-1050794638.

ishan16696:
As decided we don’t want to introduce a back-off or retry mechanism when ownerCheck fails. https://github.com/gardener/etcd-backup-restore/issues/437#issuecomment-1061839872
So closing this PR in favour of this [PR](https://github.com/gardener/etcd-backup-restore/pull/453).
/close

PR #446, opened 2022-02-25
Add feature parity for s3 compliant providers

timuthy:
This PR adds support for S3 compatible providers (see #431) when configuration is passed via secrets.

**Release note**:

```improvement operator
None
```

### DIFF
--- a/pkg/snapstore/s3_snapstore.go
+++ b/pkg/snapstore/s3_snapstore.go
@@ -52,10 +52,12 @@ const (
 )
 
 type awsCredentials struct {
-	AccessKeyID     string `json:"accessKeyID"`
-	Region          string `json:"region"`
-	SecretAccessKey string `json:"secretAccessKey"`
-	BucketName      string `json:"bucketName"`
+	AccessKeyID      string  `json:"accessKeyID"`
+	Region           string  `json:"region"`
+	SecretAccessKey  string  `json:"secretAccessKey"`
+	BucketName       string  `json:"bucketName"`
+	Endpoint         *string `json:"endpoint,omitempty"`
+	S3ForcePathStyle *bool   `json:"s3ForcePathStyle,omitempty"`
 }
 
 // S3SnapStore is snapstore with AWS S3 object store as backend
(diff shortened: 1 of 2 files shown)


ishan16696:
@timuthy Can you resolve the merge conflicts ?

pkg/snapstore/s3_snapstore.go
```diff
@@ -184,24 +188,41 @@ func readAWSCredentialFromDir(dirname string) (*awsCredentials, error) {
...
 			}
 			awsConfig.SecretAccessKey = string(data)
+		case "endpoint":
+			data, err := ioutil.ReadFile(dirname + "/endpoint")
+			if err != nil {
+				return nil, err
+			}
+			awsConfig.Endpoint = pointer.StringPtr(string(data))
+		case "s3ForcePathStyle":
+			data, err := ioutil.ReadFile(dirname + "/s3ForcePathStyle")
```
ishan16696, line 217:
```suggestion
			data, err := os.ReadFile(dirname + "/s3ForcePathStyle")
```

ishan16696: as now we have move to Go `v1.17`

ishan16696 (changes requested):

timuthy:
Thanks for the review @ishan16696. PTAL.

pkg/snapstore/s3_snapstore.go
```diff
@@ -141,8 +143,10 @@ func readAWSCredentialsJSONFile(filename string) (session.Options, error) {
 
 	return session.Options{
 		Config: aws.Config{
-			Credentials: credentials.NewStaticCredentials(awsConfig.AccessKeyID, awsConfig.SecretAccessKey, ""),
-			Region:      pointer.StringPtr(awsConfig.Region),
+			Credentials:      credentials.NewStaticCredentials(awsConfig.AccessKeyID, awsConfig.SecretAccessKey, ""),
+			Region:           pointer.StringPtr(awsConfig.Region),
+			Endpoint:         awsConfig.Endpoint,
```
ishan16696, line 148: What happen when there is no `Endpoint` configure as `Endpoint` used only for S3 compatible providers? Have you tested that scenario?

timuthy: Nothing happens, the Endpoint just remains `nil`. This is the same as not setting it at all, like we do today https://github.com/gardener/etcd-backup-restore/blob/c48c9973106837a87a652da131cbd6d5bf19b32a/pkg/snapstore/s3_snapstore.go#L78.

PR #453, opened 2022-03-11
Don't fail the Ownercheck when DNS resolve lookup fails.

### LINKED ISSUE #437: [Feature] OwnerCheck shouldn't fails the ReadinessProbe right away.

ishan16696:
Don't set the [HTTP  handler status to StatusServiceUnavailable](https://github.com/gardener/etcd-backup-restore/blob/03920151cfeeb50d6e8a35c7de2aa1f7c6670625/pkg/server/backuprestoreserver.go#L267) when the ownercheck's [DNS resolve lookup](https://github.com/gardener/etcd-backup-restore/blob/03920151cfeeb50d6e8a35c7de2aa1f7c6670625/pkg/common/ownerchecker.go#L59) fails.
Set the [HTTP  handler status to StatusServiceUnavailable](https://github.com/gardener/etcd-backup-restore/blob/03920151cfeeb50d6e8a35c7de2aa1f7c6670625/pkg/server/backuprestoreserver.go#L267) only when  [OwnerCheck](https://github.com/gardener/etcd-backup-restore/blob/03920151cfeeb50d6e8a35c7de2aa1f7c6670625/pkg/common/ownerchecker.go#L43) confirms that the current ownership has changed.

Fixes #437 

**Release note**:
```improvement operator
NONE
```

cc @plkokanov

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -275,17 +275,26 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 			}
 		}
 
-		// set server's healthz endpoint status to OK so that
-		// etcd is marked as ready to serve traffic
-		handler.SetStatus(http.StatusOK)
-
 		if runServerWithSnapshotter {
 			if b.ownerChecker != nil {
 				// Check if the actual owner ID matches the expected one
 				// If the check fails or returns false, take a final full snapshot if needed
 				b.logger.Debugf("Checking owner before starting snapshotter...")
 				result, err := b.ownerChecker.Check(ctx)
-				if err != nil || !result {
+				if err != nil {
+					b.logger.Errorf("ownerChecker check fails: %v", err)
+
+					// Wait for the configured interval before making another attempt
+					b.logger.Infof("Waiting for %s...", b.config.OwnerCheckConfig.OwnerCheckInterval.Duration)
+					select {
+					case <-ctx.Done():
+						b.logger.Info("Shutting down...")
+						return
+					case <-time.After(b.config.OwnerCheckConfig.OwnerCheckInterval.Duration):
+					}
+					continue
+
+				} else if !result {
 					handler.SetStatus(http.StatusServiceUnavailable)
 
 					// If the previous full snapshot doesn't exist or is not marked as final, take a final full snapshot
@@ -317,6 +326,10 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 					continue
 				}
 			}
+
+			// set server's healthz endpoint status to OK so that
+			// etcd is marked as ready to serve traffic
+			handler.SetStatus(http.StatusOK)
 
 			// The decision to either take an initial delta snapshot or
 			// or a full snapshot directly is based on whether there has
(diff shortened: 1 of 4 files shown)


pkg/server/backuprestoreserver.go
```diff
@@ -275,17 +275,26 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
...
 		}
 
-		// set server's healthz endpoint status to OK so that
-		// etcd is marked as ready to serve traffic
-		handler.SetStatus(http.StatusOK)
-
 		if runServerWithSnapshotter {
 			if b.ownerChecker != nil {
 				// Check if the actual owner ID matches the expected one
 				// If the check fails or returns false, take a final full snapshot if needed
```
plkokanov, line 281: Adapt this comment to the new changes appropriately.

pkg/server/backuprestoreserver.go
```diff
@@ -275,17 +275,26 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
...
+
+					// Wait for the configured interval before making another attempt
+					b.logger.Infof("Waiting for %s...", b.config.OwnerCheckConfig.OwnerCheckInterval.Duration)
+					select {
+					case <-ctx.Done():
+						b.logger.Info("Shutting down...")
+						return
+					case <-time.After(b.config.OwnerCheckConfig.OwnerCheckInterval.Duration):
+					}
+					continue
```
plkokanov, line 295: Why do we need this retry logic here? If the etcd gets restarted while there are connectivity problems to the DNS, it will not be able to start again.

plkokanov (changes requested):

pkg/server/backuprestoreserver.go
ishan16696: why etcd will not be able to start again?

pkg/server/backuprestoreserver.go
ishan16696: I didn’t get it ?

ishan16696: just wanted to set the readinessProbe only after `ownercheck` succeed/fail . Previously I missed to test this [PR](https://github.com/gardener/etcd-backup-restore/pull/411) with ownercheck.

plkokanov: I guess I wasn't very clear since my comment is on a line that wasn't changed in the PR. I meant that 
```suggestion
				// If the check returns false, take a final full snapshot if needed. If the check returns an error continue with normal operation.
```

pkg/server/backuprestoreserver.go
plkokanov: I meant the following scenario:
1. There are some errors with DNS on the landscape. 
2. For some other reason etcd-backup-restore is restarted
3. When etcd-backup-restore starts it will try to do the owner check on line 283
4. If the owner check returns an error, the select/case will block until the `b.config.OwnerCheckConfig.OwnerCheckInterval.Duration` has expired
5. Once the time expires, the loop is repeated, if there is yet again an error from the owner check then the select/case will block the loop again. 
6. Steps 4 and 5 will be repeated until the DNS errors are resolved and therefore `handler.SetStatus(http.StatusOK)` is not called and etcd-backup-restore does not become ready and snapshotter is not started

ishan16696: make sense.

plkokanov (approved):
/lgtm atleast form my side, seems like the snapshotter unit tests are failing for some reaosn though.

aaronfern (approved):
/lgtm

PR #455, opened 2022-03-16
To capture the Full snapshot time taken and compression time taken

### LINKED ISSUE #454: [Feature] Need to capture the Full snapshot time and compression time.
Currently , In [FullSnapshot](https://github.com/gardener/etcd-backup-restore/blob/1e3169cd7b7ea8d482f95ecd5fc4e23a6aed92e5/pkg/etcdutil/etcdutil.go#L250) we are just capturing  [upload time of a full snapshot](https://github.com/gardener/etcd-backup-restore/blob/1e3169cd7b7ea8d482f95ecd5fc4e23a6aed92e5/pkg/etcdutil/etcdutil.go#L269-L278).
Apart from upload time, to find the root cause of certain scenarios(mention below) we need more information:
1. Need to capture the time taken by [full snapshot api call](https://github.com/gardener/etcd-backup-restore/blob/1e3169cd7b7ea8d482f95ecd5fc4e23a6aed92e5/pkg/etcdutil/etcdutil.go#L251).
2. Need to capture the time taken by compression algorithm to [compress](https://github.com/gardener/etcd-backup-restore/blob/1e3169cd7b7ea8d482f95ecd5fc4e23a6aed92e5/pkg/etcdutil/etcdutil.go#L258) the full snapshot.

We encountered an error when `backup-restore` unable to take a full snapshot due to `compression failed: context deadline exceeded`.

logs from that scenario:
```console
2022-03-04 04:11:32 | {"log":"Taking scheduled snapshot for time: 2022-03-04 04:11:32.722506055 +0000 UTC","severity":"INFO"}
2022-03-04 04:11:32 | {"log":"Renewed delta snapshot lease with revision 16150919546 at time 2022-03-04 04:11:32.714797883 +0000 UTC m=+13681.961386812","severity":"INFO"}
2022-03-04 04:11:37 | {"log":"Successfully opened snapshot reader on etcd","severity":"INFO"}
2022-03-04 04:11:37 | {"log":"start compressing the snapshot using gzip Compression Policy","severity":"INFO"}
2022-03-04 04:11:37 | {"log":"{\"level\":\"info\",\"ts\":\"2022-03-04T04:11:37.613Z\",\"caller\":\"clientv3/maintenance.go:200\",\"msg\":\"opened snapshot stream; downloading\"}\n"}
2022-03-04 04:19:37 | {"log":"compression failed: context deadline exceeded","severity":"ERR"}
```

There could be 2 possibilities:
1. Either `etcd` was too busy and `etcd` also took time in responding to Snapshot API call.
2. Compression of snapshots was slow.

ishan16696:
1. Log the time taken by [compression](https://github.com/gardener/etcd-backup-restore/blob/1e3169cd7b7ea8d482f95ecd5fc4e23a6aed92e5/pkg/etcdutil/etcdutil.go#L258) to compress the full snapshot.
2. Log the time taken by [full snapshot api call](https://github.com/gardener/etcd-backup-restore/blob/1e3169cd7b7ea8d482f95ecd5fc4e23a6aed92e5/pkg/etcdutil/etcdutil.go#L251)
3. Increase the default snapshot time taken.

Fixes #454 

```improvement operator
NONE
```

### DIFF
--- a/pkg/etcdutil/etcdutil.go
+++ b/pkg/etcdutil/etcdutil.go
@@ -248,17 +248,24 @@ func GetEtcdEndPointsSorted(ctx context.Context, clientMaintenance client.Mainte
 
 // TakeAndSaveFullSnapshot takes full snapshot and save it to store
 func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceCloser, store brtypes.SnapStore, lastRevision int64, cc *compressor.CompressionConfig, suffix string, isFinal bool, logger *logrus.Entry) (*brtypes.Snapshot, error) {
+	startTime := time.Now()
 	rc, err := client.Snapshot(ctx)
 	if err != nil {
 		return nil, &errors.EtcdError{
 			Message: fmt.Sprintf("failed to create etcd snapshot: %v", err),
 		}
 	}
+	timeTaken := time.Since(startTime)
+	logger.Infof("Total time taken by Snapshot API: %f seconds.", timeTaken.Seconds())
+
 	if cc.Enabled {
+		startTimeCompression := time.Now()
 		rc, err = compressor.CompressSnapshot(rc, cc.CompressionPolicy)
 		if err != nil {
 			return nil, fmt.Errorf("unable to obtain reader for compressed file: %v", err)
 		}
+		timeTakenCompression := time.Since(startTimeCompression)
+		logger.Infof("Total time taken in full snapshot compression: %f seconds.", timeTakenCompression.Seconds())
 	}
 	defer rc.Close()
 
@@ -266,7 +273,6 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
 
 	// Then save the snapshot to the store.
 	snapshot := snapstore.NewSnapshot(brtypes.SnapshotKindFull, 0, lastRevision, suffix, isFinal)
-	startTime := time.Now()
 	if err := store.Save(*snapshot, rc); err != nil {
 		timeTaken := time.Since(startTime)
 		metrics.SnapshotDurationSeconds.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull, metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(timeTaken.Seconds())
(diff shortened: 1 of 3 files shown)


pkg/etcdutil/etcdutil.go
```diff
@@ -248,25 +248,31 @@ func GetEtcdEndPointsSorted(ctx context.Context, clientMaintenance client.Mainte
...
+		timeTakenCompression := time.Since(startTimeCompression)
+		logger.Infof("Total time taken in full snapshot compression: %f seconds.", timeTakenCompression.Seconds())
 	}
 	defer rc.Close()
 
 	logger.Infof("Successfully opened snapshot reader on etcd")
 
 	// Then save the snapshot to the store.
 	snapshot := snapstore.NewSnapshot(brtypes.SnapshotKindFull, 0, lastRevision, suffix, isFinal)
-	startTime := time.Now()
```
aaronfern, line 269: Do we not want any logs for how long a snapshot upload took?

ishan16696: hmm ...basically this can deduce from above information.
we have `timeTaken` which is timeTaken in api call + compression time + upload time.
we log the  Snapshot API call time, log compress time, log the total time taken. I think we this information we can Calculate the snapshot upload timee.

aaronfern: Right, makes sense 
Thanks

aaronfern (approved):
Thanks for the PR!
/lgtm

ishan16696:
@unmarshall Can you take a look at this PR?.

pkg/etcdutil/etcdutil.go
```diff
@@ -248,25 +248,31 @@ func GetEtcdEndPointsSorted(ctx context.Context, clientMaintenance client.Mainte
 
 // TakeAndSaveFullSnapshot takes full snapshot and save it to store
 func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceCloser, store brtypes.SnapStore, lastRevision int64, cc *compressor.CompressionConfig, suffix string, isFinal bool, logger *logrus.Entry) (*brtypes.Snapshot, error) {
+	startTime := time.Now()
```
unmarshall, line 251: In general this is fine. If you wish to improve the readability of the code then you can do the following (I am assuming that you will need to measure time in a LOT of places):

- Break down this function into 3 parts which can then be independently tested (unit/live)
- Introduce new utility methods:
```go
func measureRunTime(name string) func() {
     start := time.Now()
     return dumpRunTime(name, start)
}

func dumpRunTime(name string, start time.Time) func() {
       return func() {
             logger.Infof("Total run time for %s: is %f seconds", time.Since(start).Seconds())
       }
}
```
You can improve it by passing in the `sink` for the logs to the `measureRunTime` but right not you are only printing the logs to stdout so not an issue.

To use the utility function all you have to do is in the beginning of your function have:
```go
defer measureRunTime(<name>)()
```

This is not a mandatory thing to do and therefore even without this change you can merge this PR.

unmarshall (commented):
A minor comment but in general even without this change (which i leave it to you) its fine.

pkg/etcdutil/etcdutil.go
ishan16696: hmm ... make sense to me.
We can take this as an improvement later as we need to update the time with in lot of places.

ishan16696: right now, I'm merging the PR without this change as we need to release thee backup-restore today hope thats fine.

unmarshall: Yes as i said this is not a mandatory change (only good to have). Please go ahead with the merge.

PR #467, opened 2022-04-19
Throw Fatal error to avoid potential deadlock condition.

ishan16696:
It has been observed that etcdbr in etcd-events pod got stuck with this last error log:
```
2022-04-14 21:05:16 | {"log":"failed to create clientset: Get \"https://api.gcp-us1.garden.internal.canary.k8s.ondemand.com:443/api?timeout=32s\": dial tcp: lookup api.gcp-us1.garden.internal.canary.k8s.ondemand.com on 10.243.0.10:53: read udp 10.243.131.151:50810-\u003e10.243.0.10:53: i/o timeout","severity":"ERR"}
```
so, I think it should’ve return throw `Fatal error` to avoid potential deadlock condition.

Fixes #

**Release note**:
```bugfix operator
Throw Fatal error to avoid edge case potential deadlocks.
```

### DIFF
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -310,12 +310,13 @@ func RenewMemberLeasePeriodically(ctx context.Context, stopCh chan struct{}, hco
 
 	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
 	if err != nil {
-		logger.Errorf("failed to create clientset: %v", err)
+		logger.Fatalf("failed to create clientset: %v", err)
 		return
 	}
+
 	hb, err := NewHeartbeat(logger, etcdConfig, clientSet)
 	if err != nil {
-		logger.Errorf("failed to initialize new heartbeat: %v", err)
+		logger.Fatalf("failed to initialize new heartbeat: %v", err)
 		return
 	}
 	hb.heartbeatTimer = time.NewTimer(hconfig.HeartbeatDuration.Duration)
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -171,7 +171,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 				b.logger.Infof("Creating snapstore from provider: %s", b.config.SnapstoreConfig.Provider)
 				ss, err = snapstore.GetSnapstore(b.config.SnapstoreConfig)
 				if err != nil {
-					b.logger.Errorf("failed to create snapstore from configured storage provider: %v", err)
+					b.logger.Fatalf("failed to create snapstore from configured storage provider: %v", err)
 					return
 				}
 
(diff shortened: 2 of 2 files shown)


ishan16696:
Once approved, I will open an another PR for master branch.

ishan16696:
And Please let me know if there could be some other potential deadlock condition.

cc @aaronfern

vpnachev:
> Once approved, I will open an another PR for master branch.

JFYI: It should be in reverse order - first contribute the fix in master and then cherry-pick it to release/hotfix branches.

ishan16696:
yes ideally it should be the case but right now my first priority is to unblocked the patch release and test it out with `single node etcd`.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -310,12 +310,13 @@ func RenewMemberLeasePeriodically(ctx context.Context, stopCh chan struct{}, hco
 
 	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
 	if err != nil {
-		logger.Errorf("failed to create clientset: %v", err)
+		logger.Fatalf("failed to create clientset: %v", err)
```
unmarshall, line 313: The issue seems to be that while creating a clientSet, api server could not be reached within the timeout time.  Would it make sense to retry with a fixed backoff before completely giving up?
Also it would be a bit cleaner and easier to test if `RenewMemberLeasePeriodically` returns an error. The caller can then choose to exit in case this method returns an error by calling `go log.Fatal(heartbeat.RenewMemberLeasePeriodically(<method arguments>)`

ishan16696: It is edge case occurs when `Etcds` are getting updated to newer version. It's about `fail fast` to avoid deadlock.
Retrying with backoff kindof make sense but returning an error doesn’t make sense.
Basically, leader-election is thinking `RenewMemberLeasePeriodically` is running and when leader backup-restore lost election due to some error it will try to close `RenewMemberLeasePeriodically` but it was not running which cause deadlock, so returning an error doesn’t help. check [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L211-L222)

unmarshall: In the end you need to fail fast. This you can also achieve without putting `logger.Fatalf` inside the `RenewMemberLeasePeriodically` function. This function only returns an error if it has given up. The caller catches the error and then calls `go log.Fatal(heartbeat.RenewMemberLeasePeriodically(<method arguments>)`. In this case `runServer` will return an error and your etcdbr will be forced to fail fast. This is the same pattern that is also followed in `http.ListenAndServe`

ishan16696: no, I'm open to change.
I'll try to do it and let you know to review it.

ishan16696: Hey @unmarshall ,
I will add this change when I will open a PR on master branch right now I don’t want to make too many changes. Hope that’s fine

unmarshall: yes that should be fine.

Issue #466, opened 2022-04-13
Safe-guard against wrongly mounted volumes


abdasgupta:
**Describe the bug:**
We found that there is a bug in csi-driver that can try to attempt to attach two disks to the same node at same time. And so, a race condition may happen and one node may be attached with a volume that is not intended for it. So, it's possible that volume used by ETCD of one shoot is mounted as volume used for ETCD of another shoot. We need to place a check in ETCD backup-restore, so that if this unfortunate mount happens, ETCD BR will not start at all

**Expected behavior:**
ETCD BR will not start ETCD and throw error if volume of any other shoot is mounted as ETCD volume.

**How To Reproduce (as minimally and precisely as possible):**

**Logs:**

**Screenshots (if applicable):**

**Environment (please complete the following information):**
 - Etcd version/commit ID :
 - Etcd-backup-restore version/commit ID:
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]:

**Anything else we need to know?**:

vlerenc:
Thank you @abdasgupta, though I wouldn't call it a bug, but actually a safe-guard/protection, if that ever happens (the bug is with Kubernetes/CSI that leads to such cases of wrongly mounted volumes against which we want a safe-guard as the consequences would be dire).
/title Safe-guard against wrongly mounted volumes
/remove bug
/add enhancement

ishan16696:
closing this issue as it is fixed in this PR https://github.com/gardener/etcd-backup-restore/pull/468 for master branch
/close.

PR #468, opened 2022-04-20
Added a file containing shoot identifier to safe guard against wrong volume mount.

abdasgupta:
This PR checks whether wrong volume is mounted to ETCD container and if so, it doesn't start the ETCD. It writes shoot identifier to a file in the mounted volume. Then it checks if the mounted volume contains file with the same shoot identifier that is passed to it as POD Namespace.

Fixes #
https://github.com/gardener/etcd-backup-restore/issues/466

**Release note**:

```improvement operator
- ETCD won't restart from the PVC if it is wrongly mounted to the pod
```

### DIFF
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -43,6 +43,11 @@ import (
 func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int64) error {
 	start := time.Now()
 	dataDirStatus, err := e.Validator.Validate(mode, failBelowRevision)
+	if dataDirStatus == validator.WrongVolumeMounted {
+		metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
+		return fmt.Errorf("won't initialize ETCD because wrong ETCD volume is mounted: %v", err)
+	}
+
 	if dataDirStatus == validator.DataDirectoryStatusUnknown {
 		metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
 		return fmt.Errorf("error while initializing: %v", err)
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"time"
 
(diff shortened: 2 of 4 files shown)


pkg/initializer/validator/datavalidator.go
```diff
@@ -88,6 +89,35 @@ func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirSta
...
+			}
+		}
+
+		// read the content of the file safe_guard and match it with the environment variable
+		content, err := os.ReadFile(path)
+		if err != nil {
+			return WrongVolumeMounted, fmt.Errorf("can't read the content of the `safe_guard` file to determine if a wrong volume is mounted: %v", err)
+		}
+		if string(content) != namespace {
+			return WrongVolumeMounted, fmt.Errorf("wrong volume is mounted. The shoot name derived from namespcae is %s and the content of `safe_guard` file is %s", namespace, string(content))
```
dguendisch, line 117:
```suggestion
			return WrongVolumeMounted, fmt.Errorf("wrong volume is mounted. The shoot name derived from namespace is %s and the content of `safe_guard` file at %s is %s", namespace, path, string(content))
```

pkg/initializer/initializer.go
```diff
@@ -43,6 +43,11 @@ import (
 func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int64) error {
 	start := time.Now()
 	dataDirStatus, err := e.Validator.Validate(mode, failBelowRevision)
+	if dataDirStatus == validator.WrongVolumeMounted {
+		metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
+		return fmt.Errorf("won't initialize ETCD because wrong ETCD volume is mounted : %v", err)
```
dguendisch, line 48:
```suggestion
		return fmt.Errorf("won't initialize ETCD because wrong ETCD volume is mounted: %v", err)
```

dguendisch (dismissed):
Thank you! Only two nits, rest lgtm (although I cannot really test this locally, so I hope an etcd-br maintainer will review as well)

pkg/initializer/initializer.go
abdasgupta: Done

pkg/initializer/validator/datavalidator.go
abdasgupta: Done

dguendisch (commented):
/lgtm

aaronfern (approved):
I tested this and it's working as expected.

/lgtm

PR #470, opened 2022-04-25
Added a file containing shoot identifier to safe guard against wrong volume mount.

abdasgupta:
This PR checks whether wrong volume is mounted to ETCD container and if so, it doesn't start the ETCD. It writes shoot identifier to a file in the mounted volume. Then it checks if the mounted volume contains file with the same shoot identifier that is passed to it as POD Namespace.

Fixes #
https://github.com/gardener/etcd-backup-restore/issues/466

**Release note**:

```improvement operator
ETCD won't restart from the PVC if it is wrongly mounted to the pod
```

### DIFF
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/initializer/validator/datavalidator.go
```diff
@@ -88,6 +89,35 @@ func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirSta
 }
 
 func (d *DataValidator) sanityCheck(failBelowRevision int64) (DataDirStatus, error) {
+	mntDataDir := path.Dir(d.Config.DataDir)
+	path := mntDataDir + "/" + safeGuard
+	namespace := os.Getenv(podNamespace)
+	if namespace == "" {
+		d.Logger.Warn("POD_NAMESPACE environment variable is not set. The variable is used to safe guard against wrong volume mount")
```
aaronfern, line 96: Should we change this to `d.Logger.Fatal`?
The volume would still be mounted if the `POD_NAMESPACE` env var is not present

wdyt?

abdasgupta: We shouldn't be changing it to Fatal. This warning is raised in case of unit tests. In Unit test or if you run ETCD BR standalone, you may not provide the Env variable, so a warning is raised

aaronfern (approved):

PR #476, opened 2022-05-06
Fixes copy operation for Swift snapstore

plkokanov:
**What this PR does / why we need it**:
Fixes the retrieval of credentials when using the copy operation with the `Swift` snapstore.

Without this fix we get the following error when trying to copy backups:
```
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

time="2022-05-06T10:29:51Z" level=info msg="etcd-backup-restore Version: v0.15.2"
time="2022-05-06T10:29:51Z" level=info msg="Git SHA: c9b5c3d0"
time="2022-05-06T10:29:51Z" level=info msg="Go Version: go1.17.9"
time="2022-05-06T10:29:51Z" level=info msg="Go OS/Arch: linux/amd64"
time="2022-05-06T10:29:51Z" level=fatal msg="Could not get source and destination snapstores: could not get source snapstore: Missing input for argument [auth_url]"
```

This is because the `getClientOpts` function was exiting too soon if `isSource==true` and was never trying to read the file containing the source store credentials. 

I've added the following code for backwards compatibility:
```
return &clientconfig.ClientOpts{
	EnvPrefix: envPrefixSource,
}, nil
```

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Fixed retrieval of credentials during copy operation for backups stored in Swift snapstore.
```

PR #477, opened 2022-05-09
Fixes copy operation for Swift snapstore

ishan16696:
cherry-pick of PR: https://github.com/gardener/etcd-backup-restore/pull/476

Fixes the retrieval of credentials when using the copy operation with the `Swift` snapstore.

Without this fix we get the following error when trying to copy backups:
```
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

time="2022-05-06T10:29:51Z" level=info msg="etcd-backup-restore Version: v0.15.2"
time="2022-05-06T10:29:51Z" level=info msg="Git SHA: c9b5c3d0"
time="2022-05-06T10:29:51Z" level=info msg="Go Version: go1.17.9"
time="2022-05-06T10:29:51Z" level=info msg="Go OS/Arch: linux/amd64"
time="2022-05-06T10:29:51Z" level=fatal msg="Could not get source and destination snapstores: could not get source snapstore: Missing input for argument [auth_url]"
```

This is because the `getClientOpts` function was exiting too soon if `isSource==true` and was never trying to read the file containing the source store credentials. 

I've added the following code for backwards compatibility:
```
return &clientconfig.ClientOpts{
	EnvPrefix: envPrefixSource,
}, nil
```

Fixes #

**Release note**:
```improvement operator
Fixed retrieval of credentials during copy operation for backups stored in Swift snapstore.
```

### DIFF
--- a/pkg/snapstore/swift_snapstore.go
+++ b/pkg/snapstore/swift_snapstore.go
@@ -100,14 +100,9 @@ func NewSwiftSnapStore(config *brtypes.SnapstoreConfig) (*SwiftSnapStore, error)
 }
 
 func getClientOpts(isSource bool) (*clientconfig.ClientOpts, error) {
-	if isSource {
-		return &clientconfig.ClientOpts{
-			EnvPrefix: envPrefixSource,
-		}, nil
-	}
-
-	if _, isSet := os.LookupEnv(swiftCredentialJSONFile); isSet {
-		if filename := os.Getenv(swiftCredentialJSONFile); filename != "" {
+	prefix := getEnvPrefixString(isSource)
+	if _, isSet := os.LookupEnv(prefix + swiftCredentialJSONFile); isSet {
+		if filename := os.Getenv(prefix + swiftCredentialJSONFile); filename != "" {
 			clientOpts, err := readSwiftCredentialsJSON(filename)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


aaronfern (approved):

pkg/snapstore/swift_snapstore.go
```diff
@@ -100,14 +100,9 @@ func NewSwiftSnapStore(config *brtypes.SnapstoreConfig) (*SwiftSnapStore, error)
...
-		return &clientconfig.ClientOpts{
-			EnvPrefix: envPrefixSource,
-		}, nil
-	}
-
-	if _, isSet := os.LookupEnv(swiftCredentialJSONFile); isSet {
-		if filename := os.Getenv(swiftCredentialJSONFile); filename != "" {
+	prefix := getEnvPrefixString(isSource)
+	if _, isSet := os.LookupEnv(prefix + swiftCredentialJSONFile); isSet {
+		if filename := os.Getenv(prefix + swiftCredentialJSONFile); filename != "" {
```
unmarshall, line 105: `os.LookupEnv` already returns the value set against a key in the environment. Is there a reason that you wish to get the value again via `os.Getenv`?

ishan16696: yes, you are right ... will fix this at all places in separate  PR.

Issue #469, opened 2022-04-22
[Feature] Backup-restore should take care of releasing the lock over etcd db file which stops new etcd process to come up


ishan16696:
**Feature (What you would like to be added):**
1. It has been observed that some times due to abruptly termination of `etcd container` the lock on the `etcd db` file was still held by previous etcd process. And this stop new etcd process to come up as [new etcd process unable to access db file](https://github.com/etcd-io/etcd/blob/v3.4.13/etcdserver/backend.go#L69-L98) as it is unable to [open bolt db](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/mvcc/backend/backend.go#L158-L165).

2. And In other case, abnormal termination of `etcd container` leads to the database directory lock not being released and prevents the backup-restore to hang while opening the database for verification on etcd container restart.

In both of the cases, Backup-restore should detect this intervene and take care of releasing the lock over bolt db. This can also avoid manual intervention.

**Motivation (Why is this needed?):**
It has been observed that some times due to abruptly crash/shutdown of `etcd container` the lock on the `etcd db` file was still held by previous etcd process.
Logs of etcd:
```
etcdserver: another etcd process is using "/var/etcd/data/new.etcd/member/snap/db" and holds the file lock, or loading backend file is taking >10 seconds
```
Till now this scenario is very rarely observed but in `multi-node etcd-backup-restore` the chances of this kind of scenarios can also be increased.

**Approach/Hint to the implement solution (optional):**
A pod restart resolves the issue.

ishan16696:
This issue had been observed again when @seshachalam-yv was creating a load in `multi-node etcd` zonal cluster to test it.
Initially all 3 member of etcd cluster went into this issue but then @seshachalam-yv informed me that 2 out of 3 pods member recovered without doing anything.
```
etcd-main-0                                   1/2     Running            0          7m2s
etcd-main-1                                   2/2     Running            0          29m
etcd-main-2                                   2/2     Running            0          29m
```

But `etcd-main-0` didn’t recovered. Logs of `etcd` container of `etcd-main-0` pod:
```
2022-06-06 13:03:22.806358 W | etcdserver: another etcd process is using "/var/etcd/data/new.etcd/member/snap/db" and holds the file lock, or loading backend file is taking >10 seconds
2022-06-06 13:03:22.806378 W | etcdserver: waiting for it to exit before starting...
```
[...]

PR #480, opened 2022-05-23
Added a timer while opening a boltDB and improve error handling for verifying database.

ishan16696:
Till now to verify `boltDB`, backup-restore used to open it and while opening `boltDB` when it was unable to acquire the lock over database file then backup-restore used to ends up waiting indefinitely for other process to release the lock over database file.
This PR mitigate this issue by introducing a timeout setting to open boltDB, so backup-restore won't have to wait indefinitely, if it fails to acquire lock it can throw error after timeout occurs.

**Note**: Till now we haven’t decided what action needs to be taken when backup-restore throws error in this case but I think `etcd-druid` can take care of this case as it seems to be just like restoration of single member in `multi-node etcd`.

Fixes  # [469](https://github.com/gardener/etcd-backup-restore/issues/469) partially  

I have set `timeoutToOpenBoltDB` to `120 * time.Second`. I choose randomly but I'm open for change. 

**Release note**:
```improvement operator
Introducing a timeout `timeoutToOpenBoltDB` to open boltDB within a given time, so backup-restore won't have to wait for ever.
```

### DIFF
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -48,6 +48,11 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		return fmt.Errorf("won't initialize ETCD because wrong ETCD volume is mounted: %v", err)
 	}
 
+	if dataDirStatus == validator.FailToOpenBoltDBError {
+		metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
+		return fmt.Errorf("failed to initialize since another process still holds the file lock")
+	}
+
 	if dataDirStatus == validator.DataDirectoryStatusUnknown {
 		metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
 		return fmt.Errorf("error while initializing: %v", err)
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -46,11 +46,6 @@ const (
 )
 
 var (
-	// A map of valid files that can be present in the snap folder.
-	validFiles = map[string]bool{
-		"db": true,
-	}
-
 	// ErrPathRequired is returned when the path to a Bolt database is not specified.
 	ErrPathRequired = errors.New("path required")
 
@@ -79,6 +74,10 @@ func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirSta
 	if mode == Full {
 		d.Logger.Info("Checking for data directory files corruption...")
 		if err := d.checkForDataCorruption(); err != nil {
+			if errors.Is(err, bolt.ErrTimeout) {
+				d.Logger.Errorf("another etcd process is using %v and holds the file lock", d.backendPath())
+				return FailToOpenBoltDBError, err
+			}
 			d.Logger.Infof("Data directory corrupt. %v", err)
 			return DataDirectoryCorrupt, nil
 		}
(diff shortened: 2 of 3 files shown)


ishan16696:
Added a release note.
PR is up for review.

pkg/initializer/validator/datavalidator.go
```diff
@@ -376,9 +386,9 @@ func getLatestEtcdRevision(path string) (int64, error) {
 		return -1, fmt.Errorf("unable to stat backend db file: %v", err)
 	}
 
-	db, err := bolt.Open(path, 0400, &bolt.Options{ReadOnly: true})
+	db, err := bolt.Open(path, 0400, &bolt.Options{Timeout: timeoutToOpenBoltDB, ReadOnly: true})
 	if err != nil {
-		return -1, fmt.Errorf("unable to open backend boltdb file: %v", err)
+		return -1, err
```
aaronfern, line 391: I don't think change is needed here, but no strong opinion.

pkg/initializer/validator/datavalidator.go
```diff
@@ -79,6 +78,10 @@ func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirSta
 	if mode == Full {
 		d.Logger.Info("Checking for data directory files corruption...")
 		if err := d.checkForDataCorruption(); err != nil {
+			if err.Error() == ErrTimeout.Error() {
```
aaronfern, line 81: This is just a string comparison, isn't it?
would `err.Error()` return `"timeout"` in this case?

pkg/initializer/validator/datavalidator.go
ishan16696: actually it is needed, so that we can able to distinguish between the errors and detect that it is related to opening of boltdb as we want to send status as `FailToOpenBoltDBError`

pkg/initializer/validator/datavalidator.go
ishan16696: yes, it is just a string comparison.
Initially I tried to use [errors.Is()](https://pkg.go.dev/errors#Is) and [errors.As()](https://pkg.go.dev/errors#As) but they didn’t work as expected, so I used this.

pkg/initializer/validator/datavalidator.go
aaronfern: Fair enough

pkg/initializer/validator/datavalidator.go
aaronfern: I guess my question here is, will err return an error with just the string `"timeout"`? 
that's the only way this condition evaluates to true, isn't it?

unmarshall: Actually what you have defined is a `Sentinal` error and direct comparison with errors.Is(err, ErrTimeout) should work.
The issue is that you have now defined a duplicate error with the same name `ErrTimeout` that is also defined in bbolt/errors.go. So when bbolt returns a sentinal error `ErrTimeout` from `verifyDb` then you compare that against the sentinal error that you have defined in your file. This will never work.

If you wish to have your own error then in `verifyDb` method you should check if the error returned by `bolt.Open` is `ErrTimeout` as defined in `bolt` package and if yes then instead of propagating that error you wrap that error into a custom error and then return that instead.

ishan16696: thanks @unmarshall ... I will update

pkg/initializer/initializer.go
```diff
@@ -48,6 +48,11 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		return fmt.Errorf("won't initialize ETCD because wrong ETCD volume is mounted: %v", err)
 	}
 
+	if dataDirStatus == validator.FailToOpenBoltDBError {
+		metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
+		return fmt.Errorf("failed to initialize since another process is still holds the file lock")
```
aaronfern, line 53:
```suggestion
		return fmt.Errorf("failed to initialize as another process still holds the file lock")
```

aaronfern (approved):

Just a small nit, otherwise /lgtm

abdasgupta (approved):
Looks good to me

PR #482, opened 2022-05-26
Added logic for single node restoration after abrupt shutdown.

### LINKED ISSUE #481: [Feature] Single node restoration for ETCD multinode cluster
Single nodes should be able to come back up seamlessly in a multi node cluster setup.
Currently, in the multinode cluster, we are seeing that single node can't restore properly if it faces a sudden shutdown. Single nodes are unable to come back up due to revision check that is blocking it from running. That's why, we need to remove the revision check so that single nodes can come back up.
We need to determine the cluster size and if it's more than one, skip revision check.

cc @ishan16696 @aaronfern @timuthy @unmarshall

abdasgupta:
If a node abruptly shutdown in ETCD multi node cluster, it checks revision before coming back up. If the revision number is lesser than the latest snapshot, it fails to restore. Whereas this check is necessary for a single node cluster, it's not necessary for a multinode setup. This PR takes care of both type of clusters.

Fixes #481 

**Release note**:

```noteworthy operator
- An initial-cluster field is now expected in the ETCD config
```

### DIFF
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -141,6 +141,11 @@ func (d *DataValidator) sanityCheck(failBelowRevision int64) (DataDirStatus, err
 		return DataDirectoryValid, nil
 	}
 
+	if d.OriginalClusterSize > 1 {
+		d.Logger.Info("Skipping check for revision consistency of etcd member as it will get in sync with etcd leader.")
+		return DataDirectoryValid, nil
+	}
+
 	etcdRevision, err := getLatestEtcdRevision(d.backendPath())
 	if err != nil && errors.Is(err, bolt.ErrTimeout) {
 		d.Logger.Errorf("another etcd process is using %v and holds the file lock", d.backendPath())
@@ -154,7 +159,7 @@ func (d *DataValidator) sanityCheck(failBelowRevision int64) (DataDirStatus, err
 	etcdRevisionStatus, latestSnapshotRevision, err := d.checkEtcdDataRevisionConsistency(etcdRevision, failBelowRevision)
 
 	// if etcd revision is inconsistent with latest snapshot revision then
-	//   check the etcd revision consistency by starting an embedded etcd since the WALs file can have uncommited data which it was unable to flush to Bolt DB.
+	//   check the etcd revision consistency by starting an embedded etcd since the WALs file can have uncommited data which it was unable to flush to Bolt DB
 	if etcdRevisionStatus == RevisionConsistencyError {
 		d.Logger.Info("Checking for Full revision consistency...")
 		fullRevisionConsistencyStatus, err := d.checkFullRevisionConsistency(dataDir, latestSnapshotRevision)
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -28,6 +29,7 @@ import (
 	"github.com/gardener/etcd-backup-restore/pkg/leaderelection"
 	"github.com/gardener/etcd-backup-restore/pkg/metrics"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
+	"github.com/ghodss/yaml"
 
 	"github.com/gardener/etcd-backup-restore/pkg/defragmentor"
 	"github.com/gardener/etcd-backup-restore/pkg/errors"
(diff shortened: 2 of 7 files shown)


aaronfern:
This PR adds support for single node restoration only from revision mismatches and not from data/PV corruption
Is this right?

abdasgupta:
> This PR adds support for single node restoration only from revision mismatches and not from data/PV corruption
> Is this right?

Yes. Only that.

pkg/initializer/validator/datavalidator.go
```diff
@@ -142,6 +142,11 @@ func (d *DataValidator) sanityCheck(failBelowRevision int64) (DataDirStatus, err
 		return DataDirectoryValid, nil
 	}
 
+	if d.OriginalClusterSize > 1 {
+		d.Logger.Info("Skipping check for revision consistency, The revision will be updated by the other members of the cluster.")
```
timuthy, line 146:
```suggestion
		d.Logger.Info("Skipping check for revision consistency. The revision will be updated by the other members of the cluster.")
```

test/e2e/integration/cloud_backup_test.go
```diff
@@ -37,13 +37,14 @@ import (
...
-	etcdArgs := []string{"--name=etcd",
+	etcdArgs := []string{"--name=etcd1",
 		"--advertise-client-urls=http://0.0.0.0:2379",
 		"--listen-client-urls=http://0.0.0.0:2379",
+		"--initial-cluster=etcd1=http://0.0.0.0:2380",
+		"--initial-advertise-peer-urls=http://0.0.0.0:2380",
 		"--initial-cluster-state=new",
 		"--initial-cluster-token=new",
-		"--log-output=stdout",
-		"--data-dir=" + os.Getenv("ETCD_DATA_DIR")}
```
timuthy, line 46: Can you explain why this was removed?

pkg/server/backuprestoreserver.go
```diff
@@ -88,6 +91,35 @@ func NewBackupRestoreServer(logger *logrus.Logger, config *BackupRestoreComponen
...
+
+	config := map[string]interface{}{}
+	if err := yaml.Unmarshal([]byte(configYML), &config); err != nil {
+		b.logger.Fatalf("Unable to unmarshal etcd config yaml file: %v", err)
+		return err
+	}
+
+	initialClusterMap, err := types.NewURLsMap(fmt.Sprint(config["initial-cluster"]))
+	if err != nil {
+		b.logger.Fatal("Please provide initial cluster value for embedded ETCD")
```
timuthy, line 120: Please add a **release note** that this a `initial-cluster` field is now expected in the config.

timuthy (changes requested):

test/e2e/integration/cloud_backup_test.go
```diff
@@ -127,6 +129,29 @@ var _ = Describe("CloudBackup", func() {
...
 		}
+		// Create and place a ETCD config yaml
+		outfile := "/tmp/etcd.conf.yaml"
+		etcdConfigYaml := `# Human-readable name for this member.
+    name: etcd1
+    data-dir: ` + os.Getenv("ETCD_DATA_DIR") + `
+    metrics: extensive
+    snapshot-count: 75000
+    enable-v2: false
+    quota-backend-bytes: 8589934592
```
ishan16696, line 140:
```suggestion
    quota-backend-bytes: 1073741824
```
as it is just for testing we can decrease the quota bytes?

ishan16696 (changes requested):

test/e2e/integration/cloud_backup_test.go
abdasgupta: We are already providing `data directory` through config file. It's creating conflict if it's also provided by the cmd line arguments

pkg/server/backuprestoreserver.go
abdasgupta: Done

pkg/initializer/validator/datavalidator.go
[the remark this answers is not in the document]
abdasgupta: Done

test/e2e/integration/cloud_backup_test.go
abdasgupta: Done

ishan16696:
Integration tests again went into this issue https://github.com/gardener/etcd-backup-restore/issues/483

aaronfern (approved):

ishan16696:
Do not merge this PR. I have some doubts.
/do-not-merge

ishan16696 (approved):

abdasgupta:
Addresses #489

PR #487, opened 2022-06-03
Add support for etcd cluster scale up

aaronfern:
This PR adds support for scaling up an etcd cluster from 1 -> 3
Etcd backup restore will now on bootstrap detect if there is an etcd cluster already running and add itself to it as a learner. It will then promote itself to a full voting member when its logs have caught up with the leader

Fixes gardener/etcd-druid#349

To be reviewed in conjuncture with https://github.com/gardener/etcd-druid/pull/366

**Release note**:

```noteworthy operator
Etcd can now scale up itself from a single member cluster to a multi member cluster
```


pkg/initializer/initializer.go
```diff
@@ -193,3 +238,102 @@ func (e *EtcdInitializer) removeDir(dirname string) error {
...
+		Endpoints:         []string{"http://etcd-main-peer.default.svc.cluster.local:2380"}, //TODO: use ETCD_ENDPOINT env var passed by druid
+		InsecureTransport: true,                                                             //TODO: is it right to use insecure transport?
+	})
+	cli, _ := clientFactory.NewCluster()
+	r.Logger.Info("Etcd client created")
+
+	// List members in cluster
+	memListCtx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
+	etcdMemberList, err := cli.MemberList(memListCtx)
+	cancel()
```
ishan16696, line 317:
```suggestion
	defer cancel()
```

pkg/initializer/initializer.go
```diff
@@ -193,3 +238,102 @@ func (e *EtcdInitializer) removeDir(dirname string) error {
...
+	//Create etcd client
+	clientFactory := etcdutil.NewFactory(brtypes.EtcdConnectionConfig{
+		Endpoints:         []string{"http://etcd-main-peer.default.svc.cluster.local:2380"}, //TODO: use ETCD_ENDPOINT env var passed by druid
+		InsecureTransport: true,                                                             //TODO: is it right to use insecure transport?
+	})
+	cli, _ := clientFactory.NewCluster()
+	r.Logger.Info("Etcd client created")
+
+	// List members in cluster
+	memListCtx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
ishan16696, line 315: use this https://github.com/gardener/etcd-backup-restore/blob/master/pkg/types/etcdconnection.go#L46 or https://github.com/gardener/etcd-backup-restore/blob/master/pkg/types/etcdconnection.go#L29 in place of timeout

pkg/server/backuprestoreserver.go
```diff
@@ -162,6 +165,57 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+		clientFactory := etcdutil.NewFactory(brtypes.EtcdConnectionConfig{
+			Endpoints:         []string{"http://etcd-main-peer.default.svc:2380"}, //[]string{os.Getenv("ETCD_ENDPOINT")}, //TODO: Pass this via env var
+			InsecureTransport: true,
+		})
+		cli, _ := clientFactory.NewCluster()
+
+		//List all members in the etcd cluster
+		//Member URL will appear in the memberlist call response as soon as the member has been added to the cluster
+		//However, the name of the member will appear only if the member has started
+		ctx2, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
ishan16696, line 180: same as above

pkg/server/backuprestoreserver.go
```diff
@@ -162,6 +165,57 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+		etcdList, memListErr := cli.MemberList(ctx2)
+		cancel()
```
ishan16696, line 182:
```suggestion
		defer cancel()
```

ishan16696 (changes requested):
I take a initial look, direction/design seems correct to me  although I have few doubts.
Some nitpicks:

pkg/initializer/initializer.go
```diff
@@ -42,6 +46,42 @@ import (
...
+		defer cancel()
+		memberURL := getMemberURL()
+		if memberURL == "" {
+			e.Logger.Warn("Could not fetch member URL")
+		}
+		for {
+			//Create etcd client
+			clientFactory := etcdutil.NewFactory(brtypes.EtcdConnectionConfig{
+				Endpoints:         []string{"http://etcd-main-peer.default.svc:2380"}, //TODO: use ETCD_ENDPOINT env var passed by druid
+				InsecureTransport: true,
```
ishan16696, line 62: Whenever we want to connect to `embedded etcd` or its corresponding `etcd` we set `InsecureTransport: true` to get etcd client.
But here you need to connect to a etcd member of etcd cluster running in a different pod to get etcd client then I guess you should try to have secure connection. WDYT?

pkg/server/backuprestoreserver.go
```diff
@@ -162,6 +165,57 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+		}
+
+		//TODO: Simplify logic below
+		var promoted bool
+		promoted = false
+		for _, y := range etcdList.Members {
+			b.logger.Info("CHECK: ", y.Name, " : ", os.Getenv("POD_NAME"))
+			if y.Name == os.Getenv("POD_NAME") {
+				b.logger.Info("Promoting member ", y.Name)
+				ctx2, cancel := context.WithTimeout(context.TODO(), 15*time.Second)
```
ishan16696, line 196: AFAIU ctx2 will get cancel after 15sec, it means learner will only have 15sec to get in sync with leader.
Can we increase this time?

ishan16696 (changes requested):

pkg/server/httpAPI.go
```diff
@@ -432,6 +438,69 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+	}, curSts)
+	if errSts != nil {
+		h.Logger.Warn("error fetching etcd sts ", errSts)
+	}
+
+	config["initial-cluster-state"] = "new"
+	//TODO: achieve this without an sts?
+	if *curSts.Spec.Replicas > 1 && *curSts.Spec.Replicas > curSts.Status.UpdatedReplicas {
+		config["initial-cluster-state"] = "existing"
+	}
```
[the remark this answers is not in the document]
aaronfern: Done

pkg/initializer/initializer.go
```diff
@@ -42,6 +46,42 @@ import (
...
+			}
+			if err2 == nil || strings.Contains(rpctypes.ErrGRPCPeerURLExist.Error(), err2.Error()) {
+				break
+				//TODO: why not just return here?
+			}
+			e.Logger.Info("Could not as member as learner due to: ", err2)
+			e.Logger.Info("Trying again in 5 seconds")
+			timer := time.NewTimer(5 * time.Second)
+			<-timer.C
+			timer.Stop()
```
abdasgupta: Should we use `retry.OnError()` from `"k8s.io/client-go/util/retry"` package?

pkg/initializer/initializer.go
abdasgupta: Why does it need Endpoint to be passed by druid?

pkg/initializer/initializer.go
```diff
@@ -199,3 +244,102 @@ func (e *EtcdInitializer) removeDir(dirname string) error {
...
+	}
+
+	cli.Close()
+
+	return false
+}
+
+func getMemberURL() string {
+	//end := strings.Split(os.Getenv("ETCD_ENDPOINT"), "//") //TODO: use ETCD_ENDPOINT env var passed by druid
+	memberURL := "http://" + os.Getenv("POD_NAME") + ".etcd-main-peer.default.svc:2380"
```
abdasgupta: Can you form the URL from here https://github.com/gardener/etcd-backup-restore/blob/608b7c470b50a1203071f049dd5c28b31b35e9f5/pkg/server/backuprestoreserver.go#L106 ?

pkg/server/backuprestoreserver.go
abdasgupta: You may use retry.OnError instead of for loop. You may use the already available config file to form the URL

pkg/server/httpAPI.go
```diff
@@ -432,6 +438,69 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+		for _, y := range initialcluster {
+			// Add member to `initial-cluster` only if already a cluster member
+			z := strings.Split(y, "=")
+			if membrs[z[0]] || z[0] == podName {
+				cluster = cluster + y + ","
+			}
+		}
+
+		config["initial-cluster"] = cluster[:len(cluster)-1]
+	}
```
abdasgupta, line 503: Can we edit the config file backuprestoreserver.go instead of httpapi.go ?

abdasgupta (commented):
We may need a discussion on my comments

pkg/member/member_add.go
```diff
@@ -0,0 +1,158 @@
...
+		//Member URL will appear in the memberlist call response as soon as the member has been added to the cluster as a learner
+		//However, the name of the member will appear only if the member has started running
+		memListCtx, memListCtxcancel := context.WithTimeout(context.TODO(), brtypes.DefaultEtcdConnectionTimeout)
+		etcdList, memListErr := cli.MemberList(memListCtx)
+		memListCtxcancel()
+
+		if memListErr != nil {
+			logger.Info("error listing members: ", memListErr)
+			cli.Close()
+			continue
```
unmarshall, line 121: is it possible to have unrecoverable errors, if yes then you will end up in an infinite loop. I would highly recommend not to do retries in this file. Return proper errors and then let the caller handle the retry.
If you really wish to retry in this file, then have 2 functions one which does not retry and one which retries and delegates to the function which does not retry.

pkg/member/member_add.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		//TODO: Simplify logic below
+		var promoted bool
+		promoted = false
+		for _, y := range etcdList.Members {
+			if y.Name == os.Getenv("POD_NAME") {
```
unmarshall, line 128: There is no need to look out `POD_NAME` from an environment so many times. Once the environment is set its immutable, doing it just once is good enough.

Also this function should take pod name as an argument, the caller should decide from where it wishes to get the pod name and not this function, makes unit testing difficult.

pkg/member/member_add.go
```diff
@@ -0,0 +1,158 @@
...
+				_, memPromoteErr := cli.MemberPromote(memPromoteCtx, y.ID)
+				cancel()
+				if memPromoteErr == nil {
+					//Exit if member is successfully promoted or if member is not a learner
+					promoted = true
+					logger.Info("Member promoted ", y.Name, " : ", y.ID)
+					break
+				}
+				if strings.Contains(rpctypes.ErrGRPCMemberNotLearner.Error(), memPromoteErr.Error()) {
+					//Exit if member is already part of the cluster
```
unmarshall, line 141: If there is an error why do you have to loop through the rest of the members? The intent of this method is to only promote one member

pkg/member/member_add.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		if promoted {
+			break
+		}
+
+		//Timer here so that the member promote loop doesn't execute too frequently
+		logger.Info("Member still catching up logs from leader. Retrying promotion...")
+		timer := time.NewTimer(RetryPeriod)
+		<-timer.C
+		timer.Stop()
```
unmarshall, line 156: This is not a nice way of handling backoff. Needs to be improved everywhere where this pattern is used.

pkg/server/httpAPI.go
```diff
@@ -432,6 +437,71 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+		h.Logger.Errorf("failed to create clientset: %v", err)
+		return
+	}
+	curSts := &appsv1.StatefulSet{}
+	errSts := clientSet.Get(context.TODO(), client.ObjectKey{
+		Namespace: podNS,
+		Name:      podName[:strings.LastIndex(podName, "-")],
+	}, curSts)
+	if errSts != nil {
+		h.Logger.Warn("error fetching etcd sts ", errSts)
```
unmarshall, line 452: again a proper HttpResponse  with a proper code should be returned.

pkg/server/httpAPI.go
```diff
@@ -432,6 +437,71 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+	//One reason why we might want to have a strict ordering when members are joining the cluster
+	//addmember subcommand achieves this by making sure the pod with the previous index is running before attempting to add itself as a learner
+	etcdConn := *h.EtcdConnectionConfig
+	etcdConn.Endpoints = []string{fmt.Sprintf("%s://%s:%s", protocol, domaiName, peerPort)} //[]string{"http://etcd-main-peer.default.svc.cluster.local:2380"} //TODO: pass via env var
+	clientFactory := etcdutil.NewFactory(etcdConn)
+	cli, err := clientFactory.NewCluster()
+	if err != nil {
+		h.Logger.Warnf("Error with NewCluster() : %v", err)
+	}
+	ctx, cancel := context.WithTimeout(context.TODO(), brtypes.DefaultEtcdConnectionTimeout)
```
unmarshall, line 475: pass context after getting it from `req.Context()`

pkg/server/httpAPI.go
```diff
@@ -432,6 +437,71 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+	defer cancel()
+	memList, err := cli.MemberList(ctx)
+	noOfMembers := 0
+	if err != nil {
+		h.Logger.Warnf("Error with MemberList() : %v", err)
```
unmarshall, line 480: If there is an error then, do you assumed that there will be 0 members? Is that a correct assumption?

pkg/member/member_add.go
```diff
@@ -0,0 +1,158 @@
...
+func AddMemberAsLearner(logger *logrus.Logger) error {
+	//Add member as learner to cluster
+	memberURL := getMemberURL()
+	if memberURL == "" {
+		logger.Warn("Could not fetch member URL")
+	}
+	for {
+		//Create etcd client
+		//TODO: Use secure transport
+		clientFactory := etcdutil.NewFactory(brtypes.EtcdConnectionConfig{
```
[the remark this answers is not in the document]
ishan16696: Hi @unmarshall ,
Basically `NewFactory()` takes the `EtcdConnectionConfig` and returns the `clientFactory`. The thing is `EtcdConnectionConfig` can vary according to the use case. so, If we keep using the same factory this might leads to ambiguity and can leads to human error that dev don’t notice about `EtcdConfiguration` used in factory creation and ends up using same factory again and again.  I think with increasing code base it is better to create clientFactory again with `EtcdConnectionConfig`.

pkg/server/httpAPI.go
aaronfern: Yes, if there's an error here, the assumption is that it is a member bootstrap or a cluster is waking up from hibernation, so we would not want to manipulate the `initial-cluster` string

pkg/server/backuprestoreserver.go
```diff
@@ -195,6 +196,9 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 	handler := b.startHTTPServer(etcdInitializer, b.config.SnapstoreConfig.Provider, b.config.EtcdConnectionConfig, nil)
 	defer handler.Stop()
 
+	// Promotes member if it is a learner
+	member.PromoteMember(context.TODO(), b.logger)
```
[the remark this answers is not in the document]
aaronfern: 1. that's a thing I missed. Will use a proper context
2. A member will be allowed to be promoted as soon as its logs have caught up to the leader. There is no way to know exactly whether a member is ready to be promoted, so we immediately try to promote a member and we keep on trying until it succeeds. The promote call will succeed only when its logs have caught up to the leader

ishan16696:
Can you also add a release note and remove this line from PR `This is a WIP PR`.

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+
+// AddMember interface defines the functionalities needed to add a member to the etcd cluster
+type AddMember interface {
+	// AddMemberAsLearner add a new member as a learner to the etcd cluster
+	AddMemberAsLearner(*logrus.Logger, *brtypes.EtcdConnectionConfig)
+
+	// IsMemberInCluster checks is the current members peer URL is already part of the etcd cluster
+	IsMemberInCluster(*logrus.Logger, *brtypes.EtcdConnectionConfig)
+
+	// PromoteMember promotes an etcd member from a learner to a voting member of the cluster. This will succeed only if its logs are caught up with the leader
```
ishan16696, line 38:
```suggestion
	// PromoteMember promotes an etcd member from a learner to a voting member of the cluster. This will succeed if and only if:  learner is in a healthy state and the learner is in sync with leader.
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -276,3 +279,63 @@ func GetBackupLeaderEndPoint(endPoints []string, port uint) (string, error) {
...
+
+// GetEtcdSvcEndpoint returns the endpoint to the etcd client service
+func GetEtcdSvcEndpoint() (string, error) {
+	var inputFileName string
+	etcdConfigForTest := os.Getenv("ETCD_CONF")
+	if etcdConfigForTest != "" {
+		inputFileName = etcdConfigForTest
+	} else {
+		inputFileName = "/var/etcd/config/etcd.conf.yaml"
+		//inputFileName = "/Users/I544000/etcd.conf.yaml"
```
ishan16696, line 302: can you also remove this line

pkg/server/backuprestoreserver.go
```diff
@@ -190,11 +192,21 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
 	}
-	etcdInitializer := initializer.NewInitializer(restoreOpts, snapstoreConfig, b.logger.Logger)
+	etcdInitializer := initializer.NewInitializer(restoreOpts, snapstoreConfig, b.config.EtcdConnectionConfig, b.logger.Logger)
 
 	handler := b.startHTTPServer(etcdInitializer, b.config.SnapstoreConfig.Provider, b.config.EtcdConnectionConfig, nil)
 	defer handler.Stop()
 
+	// Promotes member if it is a learner
+	backoff := retry.DefaultBackoff
+	backoff.Steps = 20 //TODO: find an apt value here
```
ishan16696, line 202: Are we only trying for 20 times to call a `PromoteMember` ?

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+	//Add member as learner to cluster
+	logger.Info("Starting Add Member")
+	memberURL, err := getMemberURL(logger)
+	if memberURL == "" || err != nil {
+		logger.Info("Error fetching etcd member URL")
+		return fmt.Errorf("Could not fetch member URL : %v", err)
+	}
+
+	cli, err := m.clientFactory.NewCluster()
+	defer cli.Close()
```
unmarshall, line 72: if `m.clientFactory.NewCluster` returns an error then when your return from this method as a result of a return statement in line number 75 then it will execute the `defer cli.close` and that will be an error as cli is nil.
you should check the error first and if it is not nil then you should have defer statement.

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		logger.Info("Error creating etcd client to add member")
+		return fmt.Errorf("failed to build etcd cluster client : %v", err)
+	}
+
+	memAddCtx, cancel := getContext(context.TODO(), etcdTimeout)
```
unmarshall, line 78: this method is not so useful and therefore it should be removed. Also passing `context.TODO` instead of taking parent context that should have been passed to the method needs to be corrected.

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+	defer cancel()
+	_, err = cli.MemberAddAsLearner(memAddCtx, []string{memberURL})
+
+	if err == nil {
```
unmarshall, line 82: This is not required and should be removed.

ishan16696 (changes requested):

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+	for _, y := range etcdMemberList.Members {
+		if y.Name == podName {
+			return true, err
+		}
+	}
+
+	return false, err
+}
+
+func getMemberURL(logger *logrus.Logger) (string, error) {
```
unmarshall, line 152: As discussed before this is not the correct place to read the configuration file. It should only be read just once and not again and again. This method is currently called whenever `AddMemberAsListener` and `UpdateMemberPeerAddress` methods are invoked.

unmarshall (commented):
I have only review some of the code. In general code quality needs to be improved quite a bit. I would not recommended pushing this code as is and making a release.

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+// NewMember holds the configuration for the mechanism of adding a new member to the cluster.
+type NewMember struct {
+	clientFactory client.Factory
+}
+
+// NewMemberConfig returns new ExponentialBackoff.
+func NewMemberConfig(etcdConnConfig *brtypes.EtcdConnectionConfig) *NewMember {
+	etcdConn := *etcdConnConfig
+	svcEndpoint, err := miscellaneous.GetEtcdSvcEndpoint()
+	if svcEndpoint == "" || err != nil {
```
ishan16696, line 51: I think you should return the err or print the error or distinguish between two cases `svcEndpoint == "" || err != nil`. Let’s suppose `err!=nil` then you are setting `svcEndpoint = "http://127.0.0.1:2379"` and then if we got some error due http instead of https then we thought it is some certificates issue which will give us false error.

pkg/member/member_add.go
```diff
@@ -0,0 +1,264 @@
...
+
+	logger.Info("Could not as member as learner due to: ", err)
+	return fmt.Errorf("Could not as member as learner due to: %v", err)
+}
+
+// IsMemberInCluster checks is the current members peer URL is already part of the etcd cluster
+func (m *NewMember) IsMemberInCluster(logger *logrus.Logger) (bool, error) {
+	// Check if an etcd is already available
+	backoff := retry.DefaultBackoff
+	backoff.Steps = 2
```
ishan16696, line 103: are we trying only 2 times here ?

pkg/server/httpAPI.go
```diff
@@ -393,7 +397,9 @@ func (h *HTTPHandler) serveLatestSnapshotMetadata(rw http.ResponseWriter, req *h
 
 func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	inputFileName := "/var/etcd/config/etcd.conf.yaml"
+	//inputFileName := "/Users/I544000/etcd.conf.yaml"
 	outputFileName := "/etc/etcd.conf.yaml"
+	//outputFileName := "/Users/I544000/etcd.conf.yaml"
```
ishan16696, line 402: remove these lines as well

ishan16696 (changes requested):

pkg/server/httpAPI.go
```diff
@@ -450,6 +461,86 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+	//INITIAL_CLUSTER served via the etcd config must be tailored to the number of members in the cluster at that point. Else etcd complains with error "member count is unequal"
+	//One reason why we might want to have a strict ordering when members are joining the cluster
+	//addmember subcommand achieves this by making sure the pod with the previous index is running before attempting to add itself as a learner
+	svcEndpoint, err := miscellaneous.GetEtcdSvcEndpoint()
+	if svcEndpoint == "" || err != nil {
+		svcEndpoint = "http://127.0.0.1:2379"
+	}
+	etcdConn.Endpoints = []string{svcEndpoint}
+	clientFactory := etcdutil.NewFactory(etcdConn)
+	cli, err := clientFactory.NewCluster()
```
ishan16696, line 474: I guess you are not closing the client.
`defer cli.Close()`

pkg/server/httpAPI.go
```diff
@@ -450,6 +461,86 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		logger.Warnf("Error with NewCluster() : %v", err)
```
ishan16696, line 476: Should we return from here as we are unable to get the client.

ishan16696 (changes requested):
some more nitpicks :)

pkg/server/backuprestoreserver.go
aaronfern: Updated this to have continuously retry

pkg/member/member_add.go
aaronfern: Yup, makes sense
Updated, and also updated `miscellaneous/GetEtcdSvcEndpoint()`
PTAL

pkg/member/member_add.go
aaronfern: This retry is for `ProbeEtcd()`. I think 2 attempts should suffice to see if a cluster is already up or not
wdyt? do you recommend another value here?

pkg/server/httpAPI.go
```diff
@@ -450,6 +461,86 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	h.Logger.Info("Served config for ETCD instance.")
 }
 
+func getInitialCluster(ctx context.Context, initialCluster string, etcdConn brtypes.EtcdConnectionConfig, protocol string, clientPort string, logger logrus.Entry, podName string) string {
+	//INITIAL_CLUSTER served via the etcd config must be tailored to the number of members in the cluster at that point. Else etcd complains with error "member count is unequal"
+	//One reason why we might want to have a strict ordering when members are joining the cluster
+	//addmember subcommand achieves this by making sure the pod with the previous index is running before attempting to add itself as a learner
+	svcEndpoint, err := miscellaneous.GetEtcdSvcEndpoint()
+	if svcEndpoint == "" || err != nil {
```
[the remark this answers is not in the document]
aaronfern: Yup, makes sense
Updated, and also updated miscellaneous/GetEtcdSvcEndpoint()
PTAL

pkg/server/httpAPI.go
aaronfern: Yup, missed that 
Thanks for pointing that out

pkg/server/httpAPI.go
aaronfern: Updated
PTAL

pkg/member/member_control.go
```diff
@@ -0,0 +1,262 @@
...
+	defer cancel()
+	_, err := cli.MemberPromote(memPromoteCtx, member.ID) //Member promote call will succeed only if member is in sync with leader, and will error out otherwise
+	if err == nil {                                       //Member successfully promoted
+		m.logger.Info("Member promoted ", member.Name, " : ", member.ID)
+		return nil
+	} else if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCMemberNotLearner)) { //Member is not a learner
+		if member.PeerURLs[0] == "http://localhost:2380" { //[]string{"http://localhost:2380"}[0] {
+			// Already existing clusters have `http://localhost:2380` as the peer address. This needs to explicitly updated to the new address
+			// TODO: Remove this peer address updation logic on etcd-br v0.20.0
+			err = m.updateMemberPeerAddress(ctx, member.ID)
```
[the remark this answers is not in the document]
aaronfern: I've logged this error
Do we really need a new var here since we return `nil` there anyway?

pkg/member/member_add.go
ishan16696: make sense

pkg/member/member_control.go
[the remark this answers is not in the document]
ishan16696: ok, it make sense. Just one doubt:
If we fail to update the PeerAddress then  Is it ok to return nil ?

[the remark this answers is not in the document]
aaronfern: I think so, yes
Failure to update peer address shouldn't result in a failure imo, it would just retry the next time the pod restarts

ishan16696 (approved):

PR #492, opened 2022-06-20
Update Alpine to 3.15.4

timuthy:
```improvement operator
The Alpine base image has been updated to `3.15.4`.
```


timuthy:
/close
As discussed, we'll switch to a distroless base image.

PR #496, opened 2022-06-28
Fix the integration tests to run them locally.

ishan16696:
I was unable to run the integration tests locally as it was not getting the `REGION` ENV from file `${HOME}/.aws/config` which leads to the failure in creation of S3 bucket.
This PR fixes this issue to run the integration tests locally.

Fixes #

**Release note**:
```improvement operator
None
```

### DIFF
--- a/.ci/integration_test
+++ b/.ci/integration_test
@@ -134,6 +134,11 @@ function create_etcd_data_directory() {
   export ETCD_DATA_DIR=${TEST_DIR}/etcd-data
   mkdir -p ${ETCD_DATA_DIR}
 }
+
+function get_aws_existing_region() {
+  export REGION=`cat ${HOME}/.aws/config | tail -n 1 | awk '{print $3}'`
+}
+
 #############################
 #        AWS Setup          #
 #############################
@@ -182,6 +187,8 @@ function setup-aws-infrastructure() {
   echo "Setting up AWS infrastructure..."
   if [[ "${USE_EXISTING_AWS_SECRET}" == "1" ]]; then
     create_aws_secret
+  else
+    get_aws_existing_region
   fi
   create_s3_bucket
   echo "AWS infrastructure setup completed."
--- a/.ci/integration_test
+++ b/.ci/integration_test
@@ -136,7 +136,7 @@ function create_etcd_data_directory() {
 }
 
 function get_aws_existing_region() {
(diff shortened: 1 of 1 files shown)


.ci/integration_test
```diff
@@ -134,6 +134,11 @@ function create_etcd_data_directory() {
   export ETCD_DATA_DIR=${TEST_DIR}/etcd-data
   mkdir -p ${ETCD_DATA_DIR}
 }
+
+function get_aws_existing_region() {
+  export REGION=`cat ${HOME}/.aws/config | tail -n 1 | awk '{print $3}'`
```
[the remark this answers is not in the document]
ishan16696: I got this error `grep: invalid option -- P`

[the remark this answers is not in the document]
ishan16696: Is this fine `cat ${HOME}/.aws/config | grep -e "^.*region.*$" | awk '{print $3}'` ?

.ci/integration_test
```diff
@@ -134,6 +134,11 @@ function create_etcd_data_directory() {
   export ETCD_DATA_DIR=${TEST_DIR}/etcd-data
   mkdir -p ${ETCD_DATA_DIR}
 }
+
+function get_aws_existing_region() {
+  export REGION=`cat ${HOME}/.aws/config | grep -e "^.*region.*$" | awk '{print $3}'`
```
timuthy, line 139: Concerning the AWS documentation https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html the default config file looks like the following:
```
[default]
region=us-west-2
output=json
```

There are not white spaces and thus  `awk '{print $3}'` won't work. If you want to cover both options then you can normalize the field before extracting the region information.

ishan16696: >  There are no white spaces and thus awk '{print $3}' won't work

ok, but in my local `~/.aws/config` file have spaces and it is working fine.
```
[default]
region = eu-west-1
```
but yeah thanks .... I will try to generalise this.

timuthy: Please see https://github.com/gardener/etcd-backup-restore/pull/496/files#r947653005: We should try to solve the issue for the community and thus appreciate the generalization you offered https://github.com/gardener/etcd-backup-restore/pull/496/files#r949200282. That should only be a small change.

ishan16696: done. PTAL

timuthy (approved):
/lgtm
Thank you!

PR #501, opened 2022-07-12
Temp fix: skip single member restoration in case of data-dir invalid.

ishan16696:
By not letting it trigger the the restoration this PR fixes the issues of wrong restoration of single member in a multi-node etcd cluster in case of data-dir found to be invalid. Please refer [here](https://github.com/gardener/etcd-druid/issues/361#issuecomment-1174907483).
This PR also fixes one bug in Scale feature in func: [IsMemberInCluster](https://github.com/gardener/etcd-backup-restore/blob/a6ee3fd56bc0401fe54f048fe8166750f677a777/pkg/member/member_control.go#L164) as it shouldn't return the error, it should return the nil.

Fixes https://github.com/gardener/etcd-druid/issues/361 (Temporary fix)

**Release note**:
```bugFix operator
Temp fix: skip the single member restoration if data-dir found to be invalid.
```
```bugFix operator
Fixed a bug in Scaleup feature in func: `IsMemberInCluster()` which can cause Scaleup feature to get fail.
```

### DIFF
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -68,6 +68,9 @@ func (d *DataValidator) backendPath() string { return filepath.Join(d.snapDir(),
 func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirStatus, error) {
 	status, err := d.sanityCheck(failBelowRevision)
 	if status != DataDirectoryValid {
+		if d.OriginalClusterSize > 1 && !miscellaneous.IsBackupBucketEmpty(d.Config.SnapstoreConfig, d.Logger) {
+			return DataDirStatusUnknownInMultiNode, nil
+		}
 		return status, err
 	}
 
--- a/pkg/initializer/validator/datavalidator.go
+++ b/pkg/initializer/validator/datavalidator.go
@@ -68,6 +68,7 @@ func (d *DataValidator) backendPath() string { return filepath.Join(d.snapDir(),
 func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirStatus, error) {
 	status, err := d.sanityCheck(failBelowRevision)
 	if status != DataDirectoryValid {
+		// TODO: To be removed when backup-restore supports restoration of single member in multi-node etcd cluster.
 		if d.OriginalClusterSize > 1 && !miscellaneous.IsBackupBucketEmpty(d.Config.SnapstoreConfig, d.Logger) {
 			return DataDirStatusUnknownInMultiNode, nil
 		}
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -36,6 +36,7 @@ import (
 	"go.etcd.io/etcd/pkg/types"
 	"gopkg.in/yaml.v2"
 
+	appsv1 "k8s.io/api/apps/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	fake "sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -46,6 +47,10 @@ const (
 	NoLeaderState uint64 = 0
 	// EtcdConfigFilePath is the file path where the etcd config map is mounted.
 	EtcdConfigFilePath string = "/var/etcd/config/etcd.conf.yaml"
+	// ClusterStateNew defines the "new" state of etcd cluster.
+	ClusterStateNew = "new"
+	// ClusterStateExisting defines the "existing" state of etcd cluster.
+	ClusterStateExisting = "existing"
 )
 
 // GetLatestFullSnapshotAndDeltaSnapList returns the latest snapshot
@@ -424,4 +429,32 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
 		return true
 	}
 	return false
+}
+
+// GetInitialClusterState returns the cluster state
+func GetInitialClusterState(ctx context.Context, logger logrus.Entry, podName string, podNS string) string {
+	clusterState := "new"
+
+	//Read sts spec for updated replicas to toggle `initial-cluster-state`
+	clientSet, err := GetKubernetesClientSetOrError()
+	if err != nil {
+		logger.Errorf("failed to create clientset: %v", err)
+		return clusterState
+	}
+	curSts := &appsv1.StatefulSet{}
+	errSts := clientSet.Get(ctx, client.ObjectKey{
+		Namespace: podNS,
+		Name:      podName[:strings.LastIndex(podName, "-")],
+	}, curSts)
+	if errSts != nil {
+		logger.Warn("error fetching etcd sts ", errSts)
+		return clusterState
+	}
+
+	//TODO: achieve this without an sts?
+	if *curSts.Spec.Replicas > 1 && *curSts.Spec.Replicas > curSts.Status.UpdatedReplicas {
+		clusterState = "existing"
+	}
+
+	return clusterState
 }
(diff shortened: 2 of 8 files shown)


pkg/initializer/validator/datavalidator.go
```diff
@@ -68,6 +68,9 @@ func (d *DataValidator) backendPath() string { return filepath.Join(d.snapDir(),
 func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirStatus, error) {
 	status, err := d.sanityCheck(failBelowRevision)
 	if status != DataDirectoryValid {
+		if d.OriginalClusterSize > 1 && !miscellaneous.IsBackupBucketEmpty(d.Config.SnapstoreConfig, d.Logger) {
```
aaronfern, line 71: Why does it matter here if the bucket is empty or not? Why not just return `DataDirStatusUnknownInMultiNode` here if cluster size > 1?

aaronfern: Right, makes sense
Thanks for explaining :)

pkg/initializer/validator/datavalidator.go
```diff
@@ -68,6 +68,10 @@ func (d *DataValidator) backendPath() string { return filepath.Join(d.snapDir(),
 func (d *DataValidator) Validate(mode Mode, failBelowRevision int64) (DataDirStatus, error) {
 	status, err := d.sanityCheck(failBelowRevision)
 	if status != DataDirectoryValid {
+		// TODO: To be removed when backup-restore supports restoration of single member in multi-node etcd cluster.
+		if d.OriginalClusterSize > 1 && !miscellaneous.IsBackupBucketEmpty(d.Config.SnapstoreConfig, d.Logger) {
```
timuthy, line 72: I suggest to add this check to the `sanityCheck` function as this one contains the main logic.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -349,3 +355,52 @@ func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
...
+	}
+
+	if baseSnap == nil && (deltaSnapList == nil || len(deltaSnapList) == 0) {
+		logger.Infof("No snapshot found. BackupBucket is empty")
+		return true
+	}
+	return false
+}
+
+// GetInitialClusterState returns the cluster state
```
timuthy, line 380:
```suggestion
// GetInitialClusterState returns the cluster state, either `new` or `existing`.
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -349,3 +355,52 @@ func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
...
+func GetInitialClusterState(ctx context.Context, logger logrus.Entry, podName string, podNS string) string {
```
timuthy, line 381: We should pass the Kubernetes client to this function instead of re-creating it every time it is called. The creation and usage of a new client is expensive (mainly due to discovery calls), i.e. users of this function should create a client before and re-use it wherever it is possible. 

Please also add unit tests.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -349,3 +355,52 @@ func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
 		}
 	}
 }
+
+// IsBackupBucketEmpty checks whether the backup bucket is empty or not.
+func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logrus.Logger) bool {
```
timuthy, line 360: Please add unit tests.

timuthy (changes requested):
Thanks for the fix @ishan16696 👍

pkg/miscellaneous/miscellaneous.go
ishan16696: > The creation and usage of a new client is expensive (mainly due to discovery calls), i.e. users of this function should create a client before and re-use it wherever it is possible.

yes, agree with you but to address this issue is out of scope of this PR, we can pick this up as separate issue.
Here I can't do much, we have to create new client as `GetInitialClusterState()` is only called once.

timuthy: Hm, I don't think there is much change required.
We just have to amend the function to accept a Kubernetes client and create this client when the BackupRestore server is created. This will also allow you to write unit tests for this function with a fake or mock client.

ishan16696: done. PTAL

pkg/miscellaneous/miscellaneous.go
ishan16696: done

pkg/initializer/validator/datavalidator.go
ishan16696: I tried but it is getting little complicated, so let's not invest time in it as we are anyway going to remove this logic in upcoming releases. Hope that's fine.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -349,3 +355,52 @@ func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
...
+
+	if baseSnap == nil && (deltaSnapList == nil || len(deltaSnapList) == 0) {
+		logger.Infof("No snapshot found. BackupBucket is empty")
+		return true
+	}
+	return false
+}
+
+// GetInitialClusterState returns the cluster state, either `new` or `existing`.
+func GetInitialClusterState(ctx context.Context, logger logrus.Entry, clientSet client.Client, podName string, podNS string) string {
```
aaronfern, line 386: Can you please use the constants `ClusterStateNew` and `ClusterStateExisting` that you created instead of hardcoding?

aaronfern (changes requested):

pkg/miscellaneous/miscellaneous.go
ishan16696: done. PTAL. Also added one more unit tests

aaronfern (approved):
Thanks for the changes!
/lgtm

timuthy (changes requested):
Thanks for the improvements 👍

pkg/miscellaneous/miscellaneous.go
```diff
@@ -349,3 +355,50 @@ func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
...
+	//Read sts spec for updated replicas to toggle `initial-cluster-state`
+	curSts := &appsv1.StatefulSet{}
+	errSts := clientSet.Get(ctx, client.ObjectKey{
+		Namespace: podNS,
+		Name:      podName[:strings.LastIndex(podName, "-")],
+	}, curSts)
+	if errSts != nil {
```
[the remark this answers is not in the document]
aaronfern: The assumption here is that in most cases the value is `new`, which is the default case.
We will only need `ClusterStateExisting` in the case of a new member joining the cluster, which is ideally not a case we would see often. Even then, if we are adding a new member, we'd like everything to be up and running smoothly
An error here results in defaulting to `ClusterStateNew` which results in a new node not being allowed to join the cluster and affect the quorum

pkg/miscellaneous/miscellaneous_test.go
```diff
@@ -293,6 +297,154 @@ var _ = Describe("Miscellaneous Tests", func() {
...
+	})
+
+	Describe("Get the Initial ClusterState", func() {
+		var (
+			sts *appsv1.StatefulSet
+		)
+		BeforeEach(func() {
+			os.Setenv("STS_NAME", "etcd-test")
+			os.Setenv("POD_NAME", "etcd-test-0")
+			os.Setenv("NAMESPACE", "test_namespace")
```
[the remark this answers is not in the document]
ishan16696: removed it

pkg/miscellaneous/miscellaneous_test.go
```diff
@@ -293,6 +297,154 @@ var _ = Describe("Miscellaneous Tests", func() {
 		})
 	})
 
+	Describe("Check Emptiness of backup-bucket", func() {
```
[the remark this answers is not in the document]
ishan16696: added but in different way. PTAL

timuthy (approved):
/lgtm

timuthy:
@ishan16696 thanks for taking care. We can merge the PR after it's been rebased 🙂

PR #503, opened 2022-07-12
Added retry mechanism for member peer URL update

aaronfern:
This PR adds a retry mechanism to the member peer URL update mechanism
We notices that in a few cases (such as when running `etcd-backup-restore` locally), the peer URL update call failed due to the etcd clientset applying changes to one particular index while the etcd database has already moved to a further index. We need this retry mechanism to retry such calls so that the change is made on the correct index

Fixes #

**Release note**:

```improvement operator

```


pkg/member/member_control.go
```diff
@@ -258,11 +258,17 @@ func (m *memberControl) doPromoteMember(ctx context.Context, member *etcdserverp
 		m.logger.Info("Member promoted ", member.Name, " : ", member.ID)
 		return nil
 	} else if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCMemberNotLearner)) { //Member is not a learner
-		if member.PeerURLs[0] == "http://localhost:2380" { //[]string{"http://localhost:2380"}[0] {
+		if member.PeerURLs[0] == "http://localhost:2380" {
 			// Already existing clusters have `http://localhost:2380` as the peer address. This needs to explicitly updated to the new address
 			// TODO: Remove this peer address updation logic on etcd-br v0.20.0
-			err = m.updateMemberPeerAddress(ctx, member.ID)
-			m.logger.Errorf("Could not update member peer URL : %v", err)
+			err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
```
ishan16696, line 264: What If it won't able to update `Peer Address` in default no. of retry ?

aaronfern: Then ideally there is no action and usual br operation proceeds
Imo we shouldn't fail in case this happens

aaronfern: I don't see an issue here with the default number of retries. That should be sufficient imo

ishan16696: ok

pkg/member/member_control.go
```diff
@@ -258,11 +258,17 @@ func (m *memberControl) doPromoteMember(ctx context.Context, member *etcdserverp
...
+				return err != nil
+			}, func() error {
+				return m.updateMemberPeerAddress(ctx, member.ID)
+			})
+			if err != nil {
+				m.logger.Warnf("Could not update member peer URL : %v", err)
```
ishan16696, line 270: Instead of warning ... can you log the error.

ishan16696 (changes requested):

pkg/member/member_control.go
unmarshall: The peer URL is updated from `initial-advertise-peer-urls` which is used to communicate to other peers. If the member's peerURL cannot be updated then will this member be able to communicate to its peers? If not then will it impact on leader election and any other operation that requires quorum response?

pkg/member/member_control.go
```diff
@@ -258,11 +258,17 @@ func (m *memberControl) doPromoteMember(ctx context.Context, member *etcdserverp
 		m.logger.Info("Member promoted ", member.Name, " : ", member.ID)
 		return nil
 	} else if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCMemberNotLearner)) { //Member is not a learner
-		if member.PeerURLs[0] == "http://localhost:2380" { //[]string{"http://localhost:2380"}[0] {
+		if member.PeerURLs[0] == "http://localhost:2380" {
```
unmarshall, line 261: The name of the interface should be `MemberControl` and not `ControlMember`. The latter is a verb and is more apt for a function name and not for an interface. Also the struct is named as `memberControl`

pkg/member/member_control.go
aaronfern: This case can happen in a single-node case, for clusters that have existed for a long time, so the only impact I can see is that an existing cluster might not be able to scale up in case the peer URL could not be updated
In the multi node case, we replace failed nodes with new nodes (that would already have the updated peer URL and would not need any action), so that should not cause an impact on quorum responses.

pkg/member/member_control.go
aaronfern: Done

pkg/member/member_control.go
aaronfern: Updated the interface name to `Control` after an ad-hoc consultation

ishan16696 (approved):
LGTM. Please rebase this PR.

timuthy (commented):
Please add a release note.

PR #504, opened 2022-07-13
Added support to run etcd client calls only when in multi-node case

aaronfern:
This PR adds support to allow etcd calls only if `backup-restore` determines that it is in a multi-node setup.
This removes dependencies in a single node case of requiring network policies for etcd calls to succeed 

Fixes #

**Release note**:

```improvement operator

```


pkg/miscellaneous/miscellaneous.go
```diff
@@ -338,6 +338,36 @@ func ProbeEtcd(ctx context.Context, clientFactory etcdClient.Factory, logger *lo
...
 
+// IsMultiNode determines whether a pod is part of a multi node setup or not
+// This is determined by checking the `initial-cluster` of the etcd configmap to check the number of members expected.
+func IsMultiNode() bool {
+	var inputFileName string
+	etcdConfigForTest := os.Getenv("ETCD_CONF")
+	if etcdConfigForTest != "" {
+		return false
+	}
+	inputFileName = "/var/etcd/config/etcd.conf.yaml"
```
timuthy, line 349: I see this path being duplicated many times in the code. Can you create a constant and make the code use this?

timuthy, line 349: This should be transferred into a utility function. I also see this code being duplicated multiple times in the code base.

timuthy (changes requested):
Looks good to me in general. I added some comments to improve DRY. Please add a proper release note as well.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -338,6 +338,36 @@ func ProbeEtcd(ctx context.Context, clientFactory etcdClient.Factory, logger *lo
 	return nil
 }
 
+// IsMultiNode determines whether a pod is part of a multi node setup or not
+// This is determined by checking the `initial-cluster` of the etcd configmap to check the number of members expected.
+func IsMultiNode() bool {
```
timuthy, line 343: Maybe we can align this more with that is already used here https://github.com/gardener/etcd-backup-restore/blob/a6ee3fd56bc0401fe54f048fe8166750f677a777/pkg/server/backuprestoreserver.go#L142 to not have logic duplication all over the place.

pkg/miscellaneous/miscellaneous.go
aaronfern: Done

aaronfern: Done

timuthy: Can you please also make other areas in the code use this utility?

- https://github.com/gardener/etcd-backup-restore/blob/a6ee3fd56bc0401fe54f048fe8166750f677a777/pkg/miscellaneous/miscellaneous.go#L295
- https://github.com/gardener/etcd-backup-restore/blob/a6ee3fd56bc0401fe54f048fe8166750f677a777/pkg/member/member_control.go#L74
- https://github.com/gardener/etcd-backup-restore/blob/a6ee3fd56bc0401fe54f048fe8166750f677a777/pkg/server/backuprestoreserver.go#L103

aaronfern: Done

pkg/miscellaneous/miscellaneous.go
aaronfern: Done

pkg/miscellaneous/miscellaneous.go
```diff
@@ -41,7 +42,8 @@ import (
 
 const (
 	// NoLeaderState defines the state when etcd returns LeaderID as 0.
-	NoLeaderState uint64 = 0
+	NoLeaderState      uint64 = 0
+	etcdConfigFilePath string = "/var/etcd/config/etcd.conf.yaml"
```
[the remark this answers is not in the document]
aaronfern: Done

pkg/miscellaneous/miscellaneous.go
```diff
@@ -296,11 +297,11 @@ func GetEnvVarOrError(varName string) (string, error) {
 // GetEtcdSvcEndpoint returns the endpoint to the etcd client service
 func GetEtcdSvcEndpoint() (string, error) {
 	var inputFileName string
-	etcdConfigForTest := os.Getenv("ETCD_CONF")
-	if etcdConfigForTest != "" {
+
+	inputFileName = GetConfigFilePath()
+	if inputFileName != EtcdConfigFilePath {
```
timuthy, line 302: Can you explain why this check is required or was not there before?

pkg/server/backuprestoreserver.go
```diff
@@ -200,19 +192,21 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
-		case <-ctx.Done():
-			b.logger.Info("Context cancelled. Stopping retry promoting member")
-			return ctx.Err()
-		default:
-		}
-		m := member.NewMemberControl(b.config.EtcdConnectionConfig)
-		err := m.PromoteMember(ctx)
-		if err == nil {
-			break
+	if miscellaneous.IsMultiNode(b.logger) {
```
timuthy, line 195: Should we rather use the `restoreOpts.OriginalClusterSize` here, so that we don't have to decode the `configYML` again?

timuthy (changes requested):
Nice, thanks for the improvements 👍

pkg/miscellaneous/miscellaneous.go
aaronfern: So ideally if we are in test-mode (ETCD_CONF is set) then we return an empty service endpoint so that the default one is used (`http://127.0.0.1:2379`)
We send a proper service endpoint only when running in a cluster

timuthy: Hm, I can't really judge this case but it seems very implicit and also confusing (because why should we then supprt setting a config via `ETCD_CONF` ?!). At a minimum a comment would be nice why such a check is performed.

aaronfern: Yes, I understand, and we will be picking this up and finding a better way to implement this while refactoring the code base
For the time being I have added a comment here

pkg/server/backuprestoreserver.go

PR #506, opened 2022-07-21
Don't update peer URL when promoting member

aaronfern:
This PR updates member promote logic to not attempt to update peer URL while promoting a member.
Promote member is called only when in multi node case and in that case updated member peer URLs are already used since these are new nodes/pods trying to join the cluster

Fixes #

**Release note**:

```improvement operator
No attempt is made to update member Peer URL when trying to promote a member
```


timuthy (changes requested):
Looks good 👍 Please add a proper RELEASE NOTE so that we can merge the PR.

aaronfern:
Thanks for the review @timuthy!
Added a release note, PTAL

timuthy (approved):

Issue #517, opened 2022-08-04
[BUG] Two snapshotters running after network disruption


frenchtoasters:
**Describe the bug:**
When there is a network disruption between the etcd and etcdbrctl containers it is possible to have two instances of the snapshotter taking and uploading snapshots to the snapstore. 

**Expected behavior:**
The snapshotter should always be stopped when the state moves from `Leader` to `UnknownState`. 

**How To Reproduce (as minimally and precisely as possible):**
* Create cluster
* Trigger `context deadline exceeded` when etcdbrctl connects to running etcd 
* Before the next leaderElection check interval fix connection issues
* See that snapshotter is not stopped and second one starts

**Logs:**
```
time="2022-08-03T05:20:57Z" level=info msg="Successfully saved delta snapshot at: Incr-16537142-16537653-1659504056.gz" actor=snapshotter
time="2022-08-03T05:20:57Z" level=info msg="Stopping delta snapshot..." actor=snapshotter
time="2022-08-03T05:20:57Z" level=info msg="Resetting delta snapshot to run after 5m0s." actor=snapshotter
{"level":"warn","ts":"2022-08-03T05:22:45.581Z","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"passthrough:///https://etcd-0.etcd.kube-system.svc.cluster.local:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
time="2022-08-03T05:22:45Z" level=error msg="failed to get status of etcd endPoint: https://etcd-0.etcd.kube-system.svc.cluster.local:2379 with error: context deadline exceeded" actor=leader-elector
time="2022-08-03T05:22:45Z" level=error msg="failed to elect the backup-restore leader: context deadline exceeded" actor=leader-elector
time="2022-08-03T05:22:45Z" level=info msg="backup-restore stops leading..." actor=backup-restore-server
time="2022-08-03T05:22:45Z" level=info msg="Setting status to : 503" actor=backup-restore-server
time="2022-08-03T05:22:45Z" level=info msg="backup-restore is in: UnknownState" actor=leader-elector
time="2022-08-03T05:22:45Z" level=info msg="waiting for Re-election..." actor=leader-elector
time="2022-08-03T05:22:45Z" level=info msg="Closing defragmentor." actor=backup-restore-server
[...]

PR #518, opened 2022-08-04
Remove redef of ssrStopCh

frenchtoasters:
**What this PR does / why we need it**: I have found that in some cases where the connection between etcd and etcdbrctl is disrupted it is possible that the snapshotter is not stopped and when the connection is restored a new one is started and there are now two instances of a snapshotter running. Ensuring that we are sending the stop signal to the same `ssrStopCh` that is passed to the snapshotter when it was created. 

Fixes https://github.com/gardener/etcd-backup-restore/issues/517

**Special notes for your reviewer**: Looking at the logs you can see that snapshotter is first stopped at `2022-08-03T05:22:45Z` then at `2022-08-03T05:22:51Z` etcdbrctl detects the etcd node is the `Leader`, creates another snapshotter at `2022-08-03T05:22:51Z`. Then at `2022-08-03T05:23:05Z` it fails to get the status of etcd because of a `context deadline exceeded` this updates the backup-restore state to `UnknownState` and then on the next interval for the leaderElection check it changes the state back to `Leader`. However at `2022-08-03T05:23:05Z` only the defragmentor is closed nothing else, and at `2022-08-03T05:23:10Z` we create a new snapshotter. 

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -228,7 +228,6 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 
 	leaderCallbacks := &brtypes.LeaderCallbacks{
 		OnStartedLeading: func(leCtx context.Context) {
-			ssrStopCh = make(chan struct{})
 			var err error
 			var defragCallBack defragmentor.CallbackFunc
 			if runServerWithSnapshotter {


gardener-robot-ci-3:
Thank you @frenchtoasters for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

ishan16696:
Hi @frenchtoasters ,
thanks reporting this bug. I haven't tested/verify the bug yet.  Let me comment as I have few things in my mind.
I can agree with the bug but not with the fix. 
Let  me explain why `ssrStopCh` is redefine, suppose we removed this redef `ssrStopCh` now take this scenario:
1. there is a backup-leader --> starts snapshotter,defragmentor etc --> lost election --> backup-leader moves to unknown state --> closes snapshotter using the `ssrStopCh`.
2. again it becomes the backup-leader --> backup-leader moves from unknown state to leader --> starts snapshotter etc ==> now here's the thing we started our snapshotter again with same `ssrStopCh` channel which is already closed in previous leader-ection lost ==> now that's why it is important to redef the `ssrStopCh`.
3. now backup-leader lost the election you will not be able to close the snapshotter as `ssrStopCh` is already closed.
I tested this long time back ..... this is what I remembered.

ishan16696:
ok, I have few things in my mind to fix this although not sure if it will work or not. Let me tests that and I will let you know.

frenchtoasters:
@ishan16696 thanks for the response, laying that out really helps. Whats interesting to me though is that the channel can only handle the request once. Maybe that is cause the assumption was that if the snapshotter is triggered to stop the whole application would be getting restarted? 

My first thought after looking at what you laid out was that maybe we would want something like a `ssrStartCh` along with the `ssrStopCh`. That would allow for a pod that followed the path of Leader -> Follower -> Leader, all without being restarted, to not fall into this state. Is that kinda along the lines of what you were thinking? Or do we need to always create a new instance of the snapshotter when we start Leading?

ishan16696:
Hi @frenchtoasters ,

>  we would want something like a ssrStartCh along with the ssrStopCh

hmm .... Interesting but Can we be sure that adding more channels will help us and might not lead to more race condition. 

> do we need to always create a new instance of the snapshotter when we start Leading?
 
yes, it is important to create a new snapshotter .... it may not sound very obvious for single member etcd but it is important for multi-node etcd, please refer: https://github.com/gardener/etcd-backup-restore/blob/master/doc/concepts/leader_election.md

ishan16696:
> Is that kinda along the lines of what you were thinking? 

I was thinking to check the whether the [snapshotter is active](https://github.com/gardener/etcd-backup-restore/blob/f6e1b93e58cf6e6c1d341a22e7e47c9e6cfd7d8a/pkg/server/backuprestoreserver.go#L265) or not before starting another go routine of snapshotter .... but I haven’t this yet.

ishan16696:
closing this PR as it is not fixing the issue, please see this comment: https://github.com/gardener/etcd-backup-restore/pull/518#issuecomment-1206000534
@frenchtoasters  Let's continue our discussion on issue itself. Hope that’s fine.
/close

PR #522, opened 2022-08-26
Added the new metrics needed to be exposed for multi-node etcd.

ishan16696:
This PR adds new metrics which required to be exposed via backup-restore to get more information about multi-node etcd which will be helpful in debugging/fine-tuning the multi-node etcd.

Fixes # https://github.com/gardener/etcd-druid/issues/358

Please see https://github.com/gardener/etcd-druid/pull/414

**Release note**:
```improvement operator
Added new metrics for multi-node etcd: `etcdbr_defragmentation_duration_seconds`, `etcdbr_restoration_duration_seconds` , `etcdbr_cluster_size` , `etcdbr_is_learner `, `etcdbr_is_learner_count_total `, `etcdbr_add_learner_duration_seconds `, `etcdbr_member_remove_duration_seconds `, `etcdbr_member_promote_duration_seconds `.
```

### DIFF
--- a/pkg/leaderelection/leaderelection.go
+++ b/pkg/leaderelection/leaderelection.go
@@ -21,7 +21,9 @@ import (
 
 	"github.com/gardener/etcd-backup-restore/pkg/errors"
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
+	"github.com/gardener/etcd-backup-restore/pkg/metrics"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
@@ -149,6 +151,7 @@ func (le *LeaderElector) Run(ctx context.Context) error {
 				// If learner(non-voting member) is present in a cluster(size>1)
 				// then promote the learner to voting member.
 				if isLearner && le.PromoteCallback != nil {
+					metrics.IsLearner.With(prometheus.Labels{}).Set(1)
 					le.logger.Info("member is a learner(non-voting) member in the cluster...")
 					le.PromoteCallback.Promote(ctx, le.logger)
 				}
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -112,11 +112,13 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
 			m.logger.Infof("Member %s already part of etcd cluster", memberURL)
 			return nil
 		}
+		metrics.IsLearnerCountTotal.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Inc()
 		metrics.AddLearnerDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
 		return fmt.Errorf("error while adding member as a learner: %v", err)
 	}
 
-	metrics.HasLearner.With(prometheus.Labels{}).Set(1)
+	metrics.IsLearnerCountTotal.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededTrue}).Inc()
+	metrics.IsLearner.With(prometheus.Labels{}).Set(1)
 	metrics.AddLearnerDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededTrue}).Observe(time.Since(start).Seconds())
 	m.logger.Infof("Added member %v to cluster as a learner", strconv.FormatUint(response.Member.GetID(), 16))
 	return nil
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -11,8 +11,10 @@ import (
 
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
+	"github.com/gardener/etcd-backup-restore/pkg/metrics"
 	"github.com/gardener/etcd-backup-restore/pkg/miscellaneous"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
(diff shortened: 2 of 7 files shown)


ishan16696:
FYI.  new metric: `etcdbr_defragmentation_duration_seconds ` in multi-node etcd right now it is not working as expected.
/do-not-merge

ishan16696:
> FYI. new metric: etcdbr_defragmentation_duration_seconds in multi-node etcd right now might not work as expected.
I will verify it again.

It is working fine

aaronfern (changes requested):

pkg/metrics/metrics.go
```diff
@@ -118,23 +132,26 @@ var (
...
-	// RestorationDurationSeconds is metric to expose the duration required to restore the data directory from snapshots in seconds.
+
+	// RestorationDurationSeconds is metric to expose the duration required to restore the etcd member.
 	RestorationDurationSeconds = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Namespace: namespaceEtcdBR,
-			Name:      "restoration_duration_seconds",
-			Help:      "Total latency distribution of restoring from snapshot.",
+			Subsystem: subsystemRestore,
+			Name:      "duration_seconds",
```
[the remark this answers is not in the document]
ishan16696: It is fine, it will append the `subsystemRestore ` as prefix: so it will be `etcdbr_restoration_duration_seconds `

pkg/metrics/metrics.go
```diff
@@ -30,13 +30,22 @@ const (
 	ValueSucceededTrue = "true"
 	// ValueSucceededFalse is value False for metric label failed.
 	ValueSucceededFalse = "false"
+	// ValueRestoreSingleMemberInMN is value for metric of single member restoration in multi-node.
+	ValueRestoreSingleMemberInMN = "single_member"
```
shreyas-s-rao, line 34: I would recommend to use the name `ValueRestoreSingleMemberInMultiNode` instead.

pkg/metrics/metrics.go
```diff
@@ -168,6 +185,66 @@ var (
...
+		prometheus.HistogramOpts{
+			Namespace: namespaceEtcdBR,
+			Name:      "member_remove_duration_seconds",
+			Help:      "Total latency to remove the etcd member from the cluster.",
+		},
+		[]string{LabelSucceeded},
+	)
+
+	// MemberPromoteDurationSeconds is metric to expose duration required to promote the learner to the voting member.
+	MemberPromoteDurationSeconds = prometheus.NewHistogramVec(
```
shreyas-s-rao, line 240: Thanks for adding this very important metric!

pkg/member/member_control.go
```diff
@@ -103,15 +105,19 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
...
 	if err != nil {
 		if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCPeerURLExist)) || errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCMemberExist)) {
 			m.logger.Infof("Member %s already part of etcd cluster", memberURL)
 			return nil
 		}
+		metrics.AddLearnerDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(time.Since(start).Seconds())
 		return fmt.Errorf("error while adding member as a learner: %v", err)
 	}
 
+	metrics.HasLearner.With(prometheus.Labels{}).Set(1)
```
shreyas-s-rao, line 119: If the member is adding itself as a learner, shouldn't you also set `IsLearner` metric to true here?

pkg/server/backuprestoreserver.go
```diff
@@ -297,6 +298,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 			if restoreOpts.OriginalClusterSize > 1 {
 				m := member.NewMemberControl(b.config.EtcdConnectionConfig)
 				if err := m.PromoteMember(ctx); err == nil {
+					metrics.IsLearner.With(prometheus.Labels{}).Set(0)
```
shreyas-s-rao, line 301: Also set `HasLearner` to false, since the learner was promoted to member here.

pkg/leaderelection/leaderelection.go
```diff
@@ -149,6 +151,7 @@ func (le *LeaderElector) Run(ctx context.Context) error {
 				// If learner(non-voting member) is present in a cluster(size>1)
 				// then promote the learner to voting member.
 				if isLearner && le.PromoteCallback != nil {
+					metrics.IsLearner.With(prometheus.Labels{}).Set(1)
```
shreyas-s-rao, line 154: Also set `HasLearner` to true, since learner is present in the cluster.

shreyas-s-rao (changes requested):
@ishan16696 thanks for the much-needed PR. I have some suggestions. PTAL, thanks.

pkg/server/backuprestoreserver.go
aaronfern: Can you rather set this to `false` within `PromoteMember` of `member_control.go` so that we set the metric in all cases of member promotion?

pkg/miscellaneous/miscellaneous.go
```diff
@@ -442,17 +442,22 @@ func DoPromoteMember(ctx context.Context, member *etcdserverpb.Member, cli etcdC
 	memPromoteCtx, cancel := context.WithTimeout(ctx, brtypes.DefaultEtcdConnectionTimeout)
 	defer cancel()
 
+	start := time.Now()
 	//Member promote call will succeed only if member is in sync with leader, and will error out otherwise
 	_, err := cli.MemberPromote(memPromoteCtx, member.ID)
 	if err == nil {
 		//Member successfully promoted
+		metrics.HasLearner.With(prometheus.Labels{}).Set(0)
```
aaronfern, line 450: I see that `HasLearner` is set to `1` when `CheckIfLearnerPresent` is called which will return true when a learner is present in the cluster, but is set to `0` only when `DoPromoteMember` is called
This has a potential that a member might set its `HasLearner` metric to `1` and never set it back to `0` if it is already a voting member but finds another learner in the cluster

aaronfern (changes requested):
I had another look. In general seems fine and also working fine so far

Just a couple of comments

ishan16696:
I have introduced new metrics `etcdbr_is_learner_count_total` and removed this metrics:  `HasLearner` as it overlapped with `IsLearner `.

aaronfern (approved):
/lgtm

ishan16696:
Pipeline is having some issues ... I will re-push again.

shreyas-s-rao (approved):
@ishan16696 Thanks for making the requested changes. 
/lgtm

PR #528, opened 2022-09-01
Updated the logics to accommodate with transient quorum loss scenario.

abdasgupta:

Fixes #
https://github.com/gardener/etcd-druid/issues/436

Recovery from a transient quorum loss is supported by ETCD multinode cluster. But so far, we were extra taking action to restore single node and scale up ETCD cluster. The action was blocking normal path of recovering from transient quorum loss. I updated the logic in this PR in such a way so that recovery from transient quorum loss, single node restoration and scale up.

Following are the details of how I am testing transient quorum loss manually.

1. A three node cluster is bootstrapped using ETCD CR and druid. Before bootstrap, make sure that the backup bucket is empty.

2. After the cluster is properly bootstrapped, put some data in ETCD using ETCDCTL PUT command.

3. Wait few minutes so that incremental snapshot is taken.

4. Applied a webhook that doesn't let new pods created.

5. Deleted two pods to trigger quorum loss by Druid.

6.  After waiting some time, removed the webhook.

7. Two pods joined the cluster sequentially.

More details are in first comment.

```noteworthy operator
While recovering from the quorum loss, ETCD pods joins the cluster sequentially.
Pods join to the cluster sequentially in scale-up case as well.
```


abdasgupta:
1. Bootstrap of the three node ETCD cluster:
NAME                         READY   STATUS    RESTARTS        AGE
etcd-druid-6d985cf99-jtq98   1/1     Running   3 (4h37m ago)   8h
etcd-main-0                  2/2     Running   0               71s
etcd-main-1                  2/2     Running   0               71s
etcd-main-2                  2/2     Running   0               71s

2. Put some data in one of the member
```
$ k exec -it etcd-main-0 -c etcd /bin/sh                                                                                                                             ✹
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
/ # etcdctl --cacert=/var/etcd/ssl/client/ca/ca.crt --cert=/var/etcd/ssl/client/client/tls.crt --key=/var/etcd/ssl/client/client/tls.key --endpoints=https://etcd-main-
local:2379 put foo bar
```

3. After two pods are deleted from the cluster
NAME                         READY   STATUS    RESTARTS        AGE
etcd-druid-6d985cf99-jtq98   1/1     Running   3 (4h39m ago)   8h
etcd-main-0                  1/2     Running   0               3m4s

4. Eventually, remaining pod is also deleted
NAME                         READY   STATUS    RESTARTS        AGE
etcd-druid-6d985cf99-jtq98   1/1     Running   3 (4h39m ago)   8h

5. Druid start one pod first to recover from quorum loss

NAME                         READY   STATUS    RESTARTS        AGE
etcd-druid-6d985cf99-jtq98   1/1     Running   3 (4h40m ago)   8h
etcd-main-0                  2/2     Running   0               29s

6. Second pod joins the cluster

NAME                         READY   STATUS    RESTARTS        AGE
etcd-druid-6d985cf99-jtq98   1/1     Running   3 (4h41m ago)   8h
etcd-main-0                  2/2     Running   0               62s
etcd-main-1                  2/2     Running   0               29s
etcd-main-2                  1/2     Running   0               29s

7. Third pod joins the cluster
NAME                         READY   STATUS    RESTARTS        AGE
etcd-druid-6d985cf99-jtq98   1/1     Running   3 (4h41m ago)   8h
etcd-main-0                  2/2     Running   0               83s
etcd-main-1                  2/2     Running   0               50s
etcd-main-2                  2/2     Running   0               50s

I restarted etcd-main-2 at step 7, the cluster recovered properly.

When I restarted etcd-main-1 while it was yet to join the cluster properly, the cluster crashed. Ideally, at that point, a quorum loss should have been triggered. We need to look more deeply, why that didn't happen.

aaronfern:
With this PR, is #456 still needed?

pkg/server/backuprestoreserver.go
```diff
@@ -194,7 +193,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 	defer handler.Stop()
 
 	// Promotes member if it is a learner
-	if restoreOpts.OriginalClusterSize > 1 {
+	/*if restoreOpts.OriginalClusterSize > 1 {
```
aaronfern, line 196: With this block commented out, where are you promoting the member? I might've missed something

pkg/initializer/initializer.go
```diff
@@ -46,22 +54,95 @@ import (
...
+					retry.OnError(backOff, func(err error) bool {
+						return err != nil
+					}, func() error {
+						// check whether the cluster size without the learner is just one less than the member index found from the POD_NAME
+						size, err := m.ClusterSizeExcludingLearner(ctx)
+						if err != nil {
+							return err
+						}
+
+						// trick to add member as learner sequentially
```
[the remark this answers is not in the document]
abdasgupta: Yes, it's beneficial. Also, it's recommended in ETCD Doc

pkg/server/backuprestoreserver.go
abdasgupta: I really need your help here. I don't understand why these lines are here. We are anyway promoting members in L310

aaronfern: The `PromoteMember` in L310 is mainly targeted at cases when there is single member restoration. 
In that case, etcd-br will not restart, so these lines will not be executed, hence it's added there.

For new members, will `le.CurrentState == StateFollower` turn out to be true? If that is the case, then we can consider removing this block and let the `leaderelector` handle promotion, however we will still need the part where we update members (`UpdateMember()`) for existing clusters to have their peer URLs updated

pkg/server/httpAPI.go
```diff
@@ -503,12 +507,39 @@ func (h *HTTPHandler) GetClusterState(ctx context.Context, clusterSize int, clie
...
-	// addmember subcommand achieves this by making sure the pod with the previous index is running before attempting to add itself as a learner
+func getInitialCluster(ctx context.Context, initialCluster string, etcdConn brtypes.EtcdConnectionConfig, protocol string, clientPort string, logger logrus.Entry, podName string) string {
+	//INITIAL_CLUSTER served via the etcd config must be tailored to the number of members in the cluster at that point. Else etcd complains with error "member count is unequal"
+	//One reason why we might want to have a strict ordering when members are joining the cluster
+	//addmember subcommand achieves this by making sure the pod with the previous index is running before attempting to add itself as a learner
+
+	//Test for quorum loss
+	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+	if err != nil {
+		logger.Errorf("failed to create clientset: %v", err)
```
ishan16696, line 518: Isn't it is better to return error or fatal error here ?

pkg/initializer/initializer.go
```diff
@@ -88,13 +169,15 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 
 	metrics.ValidationDurationSeconds.With(prometheus.Labels{metrics.LabelSucceeded: metrics.ValueSucceededTrue}).Observe(time.Since(start).Seconds())
 
+	// bootstrap of the first member after quorum loss case as well as the bootstrap case come to this point
+	// so, check for quorum loss case and skip restoration using removing and adding member
 	if dataDirStatus != validator.DataDirectoryValid {
-		if dataDirStatus == validator.DataDirStatusInvalidInMultiNode || (e.Validator.OriginalClusterSize > 1 && dataDirStatus == validator.DataDirectoryCorrupt) || (e.Validator.OriginalClusterSize > 1 && isEtcdMemberPresent) {
+		if (dataDirStatus == validator.DataDirStatusInvalidInMultiNode || (e.Validator.OriginalClusterSize > 1 && dataDirStatus == validator.DataDirectoryCorrupt) || (e.Validator.OriginalClusterSize > 1 && isEtcdMemberPresent)) && !quorumLossCaseFound {
```
ishan16696, line 175: > .... &&`! quorumLossCaseFound ` 

Why we required this condition in case of quorumLoss ?
AFAIU, if there is a quorum loss then controller will scale-down the cluster to 0 then scale-up to 1.
So, IMO  `OriginalClusterSize` will handle this case ... isn't ?

pkg/initializer/initializer.go
```diff
@@ -46,22 +54,95 @@ import (
...
+					backOff := wait.Backoff{
+						Steps:    100,
+						Duration: 30 * time.Second,
+						Factor:   5.0,
+						Jitter:   0.1,
+					}
+
+					retry.OnError(backOff, func(err error) bool {
+						return err != nil
+					}, func() error {
```
ishan16696, line 128:
```suggestion
					retry.OnError(backOff, AnyError, func() error {
```
please see [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/errors/error.go#L38)

ishan16696 (changes requested):

ishan16696:
> While recovering from the quorum loss, all the ETCD pods joins the cluster sequentially.

What I can determine from your logic it is also true for scale-up feature, so I guess you can update the release note.

pkg/initializer/initializer.go
abdasgupta: Nope. It's not handling

pkg/initializer/initializer.go
abdasgupta: Done

pkg/initializer/initializer.go
abdasgupta: Done

pkg/server/httpAPI.go
abdasgupta: Done

pkg/member/member_control.go
```diff
@@ -308,3 +311,27 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 
 	return miscellaneous.CheckIfLearnerPresent(learnerCtx, cli)
 }
+
+// ClusterSizeExcludingLearnerdetermines the cluster size without considering the learners
+func (m *memberControl) ClusterSizeExcludingLearner(ctx context.Context) (int, error) {
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/miscellaneous/miscellaneous.go
```diff
@@ -197,6 +197,20 @@ func GetKubernetesClientSetOrError() (client.Client, error) {
 	return cl, nil
 }
 
+// GetKubernetesClientSetWithSchemeOrError creates and returns a kubernetes clientset or an error if creation fails
```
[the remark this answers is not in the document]
abdasgupta: Done

go.mod
```diff
@@ -53,6 +53,11 @@ require (
 	sigs.k8s.io/controller-runtime v0.8.3
 )
 
+require (
+	github.com/gardener/etcd-druid/api v0.6.0
```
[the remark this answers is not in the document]
abdasgupta: Any suggestion on how to do it? I don't see it can fetch higher version

ashwani2k:
@abdasgupta  Thanks for creating the issue to link this PR. 
Few comments on the overall structure which can help us to align this with other PRs. 

> What this PR does / why we need it:

Section is empty ?? 
Can you please elaborate how the PR fixes the https://github.com/gardener/etcd-druid/issues/436#issue-1378486604. 
The special note for reviewer is same as the motivation part in https://github.com/gardener/etcd-druid/issues/436#issue-1378486604, so it will immensely help the readers of this PR to elaborate the approach used to fix the issue highlighted in the motivation section. 
Also though this is fresh in everyone's memory on how it solves, but if someone were to look at it lets say 6 months down the line then they will actually read this section to ascertain if they should check the code or not for what they are looking for. 

Some reference PRs in multi-node scope where PR description is well captured. 
https://github.com/gardener/etcd-backup-restore/pull/482
https://github.com/gardener/etcd-backup-restore/pull/532
https://github.com/gardener/etcd-backup-restore/pull/411 


Also while reading the release notes, I couldn't get the gist of the what's being introduced with this PR to allow transient quorum loss to recover. 
>While recovering from the quorum loss, ETCD pods joins the cluster sequentially.
>Pods join to the cluster sequentially in scale-up case as well.

Which of the above 2 statements highlights what has been introduced new with this PR? I believe only the second one as it says as well. However to me it looked like 2 isolated statements which might be understood by someone involved in the implementation/discussion but may be not to me. 
 @ishan16696, @aaronfern, Do you read/see it differently? 

While reading it, I also checked the above listed PR for their release notes and they looked more apt. (but may be that's just me 🤔 )
https://github.com/gardener/etcd-backup-restore/pull/482
https://github.com/gardener/etcd-backup-restore/pull/532
https://github.com/gardener/etcd-backup-restore/pull/509 
https://github.com/gardener/etcd-backup-restore/pull/411

ashwani2k:
> Which of the above 2 statements highlights what has been introduced new with this PR? I believe only the second one as it says as well. However to me it looked like 2 isolated statements which might be understood by someone involved in the implementation/discussion but may be not to me.

Also, If the answer is both are valid then i'll propose we can combine them.  
One more thing which came to me on a reading it again, as we are fixing transient quorum loss so we should highlight this as well in the release notes to avoid confusion with the other story of permanent quorum loss. 

Also shouldn't the categorization for release note be `bug` instead of noteworthy and operator as I believe the operator will have nothing to act on here as this will automatically recover or am I missing something? 
```
-->
```noteworthy operator
While recovering from the quorum loss, ETCD pods joins the cluster sequentially.
Pods join to the cluster sequentially in scale-up case as well.
```

ishan16696:
@abdasgupta I guess we can close this PR ?

ishan16696:
We have found that to recover from the **transient quorum loss scenario** there is no requirement of adding additional logic as PVC/data-dir of each cluster member in the etcd cluster are intact, hence **transient quorum loss**  get recover without need of any manual intervention.
To recover from the **permanent quorum loss scenario** we have added playbook doc [here](https://github.com/gardener/etcd-druid/blob/master/docs/operation/Recover_From_Etcd_Permanent_Quorum_Loss.md). Please follow the steps mentioned in playbook to recover the cluster from permanent quorum loss.

Hence, I'm closing this PR.
/close

PR #530, opened 2022-09-02
Enhance etcd lease to add peer url enabled status as an annotation

### LINKED ISSUE #529: [Feature] Enhance etcd lease to add peer url enabled status as an annotation
Add annotation to etcd lease to contain peerUrlTLSEnabled boolean value.

Currently for a single node etcd cluster, the member peer URL is not TLS enabled. If this cluster is scaled from 1 to 3 then the new members that are started have peer certs configured and already communicate via peer URL that is TLS enabled. Since the first (existing) member's peer URL is not TLS enabled, the new members fail to communicate and enter into `CrashLoopBackOff`.

[etcd-druid/PR#6533](https://github.com/gardener/gardener/pull/6533) addresses this shortcoming. However, it requires that the lease is updated with peerURLTLSEnabled annotation which it will depend upon to determine if corrective action is required (which involves deleting STS).

unmarshall:
Adds an annotation to etcd lease which indicates if the peer url is TLS enabled.

Fixes #529 

**Release note**:

```improvement operator
Adds an annotation to etcd lease which indicates if the peer url is TLS enabled.
```

### DIFF
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -317,9 +317,9 @@ func RenewMemberLeasePeriodically(ctx context.Context, stopCh chan struct{}, hco
 		return fmt.Errorf("failed to create clientset: %v", err)
 	}
 
-        metadata := map[string]string{
-           PeerUrlTLSEnabledKey: strconv.FormatBool(peerUrlTLSEnabled)
-        }
+	metadata := map[string]string{
+		PeerUrlTLSEnabledKey: strconv.FormatBool(peerUrlTLSEnabled),
+	}
 
 	hb, err := NewHeartbeat(logger, etcdConfig, clientSet, metadata)
 	if err != nil {
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -317,8 +317,9 @@ func RenewMemberLeasePeriodically(ctx context.Context, stopCh chan struct{}, hco
 		return fmt.Errorf("failed to create clientset: %v", err)
 	}
 
-	metadata := make(map[string]string)
-	metadata[PeerUrlTLSEnabledKey] = strconv.FormatBool(peerUrlTLSEnabled)
+        metadata := map[string]string{
+           PeerUrlTLSEnabledKey: strconv.FormatBool(peerUrlTLSEnabled)
+        }
 
 	hb, err := NewHeartbeat(logger, etcdConfig, clientSet, metadata)
 	if err != nil {
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -34,9 +34,10 @@ import (
 )
 
 const (
-	podName              = "POD_NAME"
-	podNamespace         = "POD_NAMESPACE"
-	PeerUrlTLSEnabledKey = "member.etcd.gardener.cloud/tls-enabled"
+	podName      = "POD_NAME"
+	podNamespace = "POD_NAMESPACE"
+	// PeerURLTLSEnabledKey is the name of the annotation that will be added to the lease and will indicate whether TLS has been enabled for peer URL
+	PeerURLTLSEnabledKey = "member.etcd.gardener.cloud/tls-enabled"
 )
 
 // Heartbeat contains information to perform regular heart beats in a Kubernetes cluster.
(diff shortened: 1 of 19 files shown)


pkg/server/backuprestoreserver.go
```diff
@@ -210,6 +216,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 			}
 			miscellaneous.SleepWithContext(ctx, retryTimeout)
 		}
+		peerUrlTLSEnabled = true
```
timuthy, line 219: I think it's not safe to assume that `peerUrlTLS` is enabled when using multi-node is used. It's true for Gardener but not in general.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -129,6 +132,7 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 	renewedMemberLease.Spec.HolderIdentity = &memberID
 	renewedTime := time.Now()
 	renewedMemberLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
+	renewedMemberLease.Annotations = hb.metadata
```
timuthy, line 135: Please only overwrite the `PeerUrlTLSEnabledKey`, not the entire annotations.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -306,14 +310,17 @@ func DeltaSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, k8s
...
+func RenewMemberLeasePeriodically(ctx context.Context, stopCh chan struct{}, hconfig *brtypes.HealthConfig, logger *logrus.Entry, etcdConfig *brtypes.EtcdConnectionConfig, peerUrlTLSEnabled bool) error {
 
 	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
 	if err != nil {
 		return fmt.Errorf("failed to create clientset: %v", err)
 	}
 
-	hb, err := NewHeartbeat(logger, etcdConfig, clientSet)
+	metadata := make(map[string]string)
+	metadata[PeerUrlTLSEnabledKey] = strconv.FormatBool(peerUrlTLSEnabled)
```
timuthy, line 321:
```suggestion
        metadata := map[string]string{
           PeerUrlTLSEnabledKey: strconv.FormatBool(peerUrlTLSEnabled)
        }
```

timuthy (changes requested):
As far as I can tell, it generally looks good. I added a few comments but haven't tested it though.

ishan16696:
Can you also mention/describe in the PR that why this change is required ?

pkg/server/backuprestoreserver.go
unmarshall: Ok you are correct.

pkg/health/heartbeat/heartbeat.go
unmarshall: Fixed now

timuthy (approved):
/lgtm
/needs second-opinion
@aaronfern can you have a look as well?

pkg/server/backuprestoreserver.go
unmarshall: changed it

timuthy (approved):
Thank you! 🙂

ishan16696:
Hi @unmarshall ,
Can you please do a force push as pipeline checks are not passing and I'm unable to trigger a new build ?

unmarshall:
> Hi @unmarshall , Can you please do a force push as pipeline checks are not passing and I'm unable to trigger a new build ?

Done

unmarshall:
> While testing I've noticed that while the correct annotation is being set in the single node case, when in the multi node case it's sets the annotation to `false` irrespective of whether TLS is enabled or not
> 
> Can you please relook this?

Ok let me check this once. As you have indicated that reading the config file will be better in this case as this is the single source of truth. I really wish the reading is just done once. Perhaps this could be improved going futher.

unmarshall:
>While testing I've noticed that while the correct annotation is being set in the single node case, when in the multi node case it's sets the annotation to false irrespective of whether TLS is enabled or not

I found the issue. If you trace back the initial point is the command to start the server. So if you see init.go then you will see a method NewBackupRestoreComponentConfig which initializes RestorationConfig by invoking brtypes.NewRestorationConfig  which initializes InitialAdvertisePeerURLs to default value which is http://localhost:2380/
Since we are using this only for multi-node so it gives an incorrect result. I will now correct it and only depend upon the config file.

aaronfern (approved):
Thanks for the changes @unmarshall!
/lgtm

PR #532, opened 2022-09-13
Fix the ProbeEtcd timeout.

ishan16696:
We have observed that currently `ProbeEtcd` is using the longer timeout which is set using this flag `--etcd-connection-timeout` and in our landscapes `--etcd-connection-timeout` is set to `5mins`.  (check [here](https://github.com/gardener/etcd-druid/blob/master/pkg/component/etcd/statefulset/values_helper.go#L247))
Due to this longer Timeout `ProbeEtcd` is keep waiting instead of failing the probe and meanwhile when `ProbeEtcd` is waiting if corresponding etcd comes-up then `ProbeEtcd` get succeeds this might leads to some race-conditions between the go-routines as `ProbeEtcd` is suppose to fail.
This PR fixes this issue by using the shorter Timeout of `5sec` in `ProbeEtcd`.

Fixes #

cc @timuthy 

**Release note**:
```bugfix operator
To avoid potential race-condition between go-routines updated `probeEtcd func()` to use shorter timeout.
```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -363,7 +363,7 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 			err = b.probeEtcd(ctx)
 		}
 		if err != nil {
-			b.logger.Errorf("Failed to probe etcd: %v", err)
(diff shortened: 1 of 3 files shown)


timuthy (approved):
/lgtm

PR #534, opened 2022-09-14
Member Peer URL not updated when scaled to multi-node, scale identification is also not correct

### LINKED ISSUE #533: [BUG] Member Peer URL not updated when scaled to multi-node, scale identification is also not correct
**Describe the bug:**
In a single node etcd cluster today, the peer URL is not TLS enabled. When the etcd cluster is scaled from 1 to 3 then the other nodes only talk over TLS but since the first member cannot talk TLS, the other 2 members fail to join the cluster. The pods go into `CrashLoopBackOff`. To correct that we earlier merge #530, however that is not sufficient. To ensure that the TLS is now enabled for peer URL, etcd-druid will restart (implemented as delete sts + create sts) 2 times. Since the etcd configuration has already changed, the restart causes the new configuration to be available to the etcd member.  In the new configuration the cluster size is already 3. In backuprestoreserver.go the call to `UpdateMemberPeerURL` is only made when the cluster size == 1. This results in the peer URL never getting TLS enabled.

Another issue is that the way we identify if this is a scale up case is incorrect. We depend upon the following check:
```go
	if *curSts.Spec.Replicas > 1 && *curSts.Spec.Replicas > curSts.Status.UpdatedReplicas {
		return pointer.StringPtr(ClusterStateExisting), nil
	}
```
This check is not deterministic as `UpdatedReplicas` only means that the sts controller has updated all replicas to have the current revision. It does not mean they are ready. As soon as the replicas are changed from 1 to 3, UpdateReplicas also change very fast to 3, resulting in the failure of the above check, making this check timing dependent and thus non-deterministic. The state that is returned is `ClusterStateNew` which causes one of the members to start a new cluster and once it does that its clusterID does not match with the existing clusterID and it fails to join the cluster.

**Expected behavior:**

Peer URL should be TLS enabled always when etcd cluster is scaled from 1 to 3. Also the identification that this is not a bootstrap case but a scale up case should be based on time insensitive conditions.

**Logs:**

 - Etcd version/commit ID :

unmarshall:

Member peer URL should always be updated. This will ensure that member peer URL is correct updated even if there was a recreation of sts and as a consequence a new configmap having a cluster size > 1 is available.
It also fixes the way we identify scale up scenario. 

Fixes #533 

**Release note**:

```improvement operator
Always update member peer URL, changed the way scale-up of etcd cluster is identified.
```

### DIFF
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -418,6 +418,8 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
 	return true
 }
 
+const scaledToMultiNodeAnnotationKey = "gardener.cloud/scaled-to-multi-node"
+
 // GetInitialClusterStateIfScaleup checks if it is the Scale-up scenario and returns the cluster state either `new` or `existing`.
 func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, clientSet client.Client, podName string, podNS string) (*string, error) {
 	//Read sts spec for updated replicas to toggle `initial-cluster-state`
@@ -431,7 +433,7 @@ func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, c
 		return nil, err
 	}
 
-	if metav1.HasAnnotation(curSts.ObjectMeta, "gardener.cloud/scaled-to-multi-node") {
+	if metav1.HasAnnotation(curSts.ObjectMeta, scaledToMultiNodeAnnotationKey) {
 		return pointer.StringPtr(ClusterStateExisting), nil
 	}
 
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -39,6 +39,8 @@ import (
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	"go.etcd.io/etcd/pkg/types"
 	"gopkg.in/yaml.v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 
 	appsv1 "k8s.io/api/apps/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -417,24 +419,27 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
 }
 
 // GetInitialClusterStateIfScaleup checks if it is the Scale-up scenario and returns the cluster state either `new` or `existing`.
-func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, clientSet client.Client, podName string, podNS string) string {
+func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, clientSet client.Client, podName string, podNS string) (*string, error) {
 	//Read sts spec for updated replicas to toggle `initial-cluster-state`
 	curSts := &appsv1.StatefulSet{}
-	errSts := clientSet.Get(ctx, client.ObjectKey{
+	err := clientSet.Get(ctx, client.ObjectKey{
 		Namespace: podNS,
 		Name:      podName[:strings.LastIndex(podName, "-")],
 	}, curSts)
-	if errSts != nil {
-		logger.Warn("error fetching etcd sts ", errSts)
-		return ClusterStateNew
+	if err != nil {
+		logger.Warn("error fetching etcd sts ", err)
+		return nil, err
 	}
 
-	//TODO: achieve this without an sts?
+	if metav1.HasAnnotation(curSts.ObjectMeta, "gardener.cloud/scaled-to-multi-node") {
+		return pointer.StringPtr(ClusterStateExisting), nil
+	}
+
+	// fallback - preserves backward compatibility
 	if *curSts.Spec.Replicas > 1 && *curSts.Spec.Replicas > curSts.Status.UpdatedReplicas {
-		return ClusterStateExisting
+		return pointer.StringPtr(ClusterStateExisting), nil
 	}
-
-	return ""
+	return nil, nil
 }
 
 // DoPromoteMember promotes a given learner to a voting member.
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -224,22 +224,22 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 			}
 			_ = miscellaneous.SleepWithContext(ctx, retryTimeout)
 		}
-	} else {
-		// when OriginalClusterSize = 1
-		m := member.NewMemberControl(b.config.EtcdConnectionConfig)
-		err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
-			return err != nil
-		}, func() error {
-			cli, err := etcdutil.NewFactory(*b.config.EtcdConnectionConfig).NewCluster()
-			if err != nil {
-				return err
-			}
-			_, err = m.UpdateMemberPeerURL(ctx, cli)
-			return err
-		})
+	}
+
+	// when OriginalClusterSize = 1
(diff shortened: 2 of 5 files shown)


pkg/server/backuprestoreserver.go
```diff
@@ -224,22 +224,22 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
-			_, err = m.UpdateMemberPeerURL(ctx, cli)
-			return err
-		})
+	}
+
+	// when OriginalClusterSize = 1
+	m := member.NewMemberControl(b.config.EtcdConnectionConfig)
+	err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
+		return err != nil
+	}, func() error {
```
ishan16696, line 233:
```suggestion
	err := retry.OnError(retry.DefaultBackoff, AnyError, func() error {
```
Please see [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/errors/error.go#L38)

pkg/server/backuprestoreserver.go
```diff
@@ -224,22 +224,22 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
-			cli, err := etcdutil.NewFactory(*b.config.EtcdConnectionConfig).NewCluster()
-			if err != nil {
-				return err
-			}
-			_, err = m.UpdateMemberPeerURL(ctx, cli)
-			return err
-		})
+	}
+
+	// when OriginalClusterSize = 1
```
ishan16696, line 229: if we are doing the `UpdateMemberPeerURL` irrespective of clusterSize then I guess you can remove this comment.

pkg/server/backuprestoreserver.go
```diff
@@ -224,22 +224,22 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+	m := member.NewMemberControl(b.config.EtcdConnectionConfig)
+	err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
+		return err != nil
+	}, func() error {
+		cli, err := etcdutil.NewFactory(*b.config.EtcdConnectionConfig).NewCluster()
 		if err != nil {
-			b.logger.Error("unable to update the member")
+			return err
```
ishan16696, line 236: can you please log the error `fmt.Errorf("unable to create etcd cluster client ",err)` as well , so we can get to know the reason behind if we see this error log `unable to update the member`

ishan16696 (changes requested):
Hi @unmarshall ,
thanks for PR.
Have you also tested the single member restoration feature with these changes ?

pkg/server/backuprestoreserver.go
unmarshall: Ah yes, i forgot. Changed.

pkg/server/backuprestoreserver.go
unmarshall: removed.

pkg/server/backuprestoreserver.go
unmarshall: i discussed this with Aaron. He did not want to return an error in this. Can you check with him once?

aaronfern: This line in inside a `retry.OnError` function, so imo it's okay if it is not logged here as it will be anyway logged in line [239](https://github.com/gardener/etcd-backup-restore/pull/534/files#diff-df6f7abbd9069d0a874a18b5717de5b0fcd68dac32184fdcdc2a26019be2296fR239)
Having said that, can you please update [line 239](https://github.com/gardener/etcd-backup-restore/pull/534/files#diff-df6f7abbd9069d0a874a18b5717de5b0fcd68dac32184fdcdc2a26019be2296fR239) to print the error along with the log?

pkg/miscellaneous/miscellaneous.go
```diff
@@ -417,24 +419,27 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
...
-		return ClusterStateNew
+	if err != nil {
+		logger.Warn("error fetching etcd sts ", err)
+		return nil, err
 	}
 
-	//TODO: achieve this without an sts?
-	if *curSts.Spec.Replicas > 1 && *curSts.Spec.Replicas > curSts.Status.UpdatedReplicas {
-		return ClusterStateExisting
+	if metav1.HasAnnotation(curSts.ObjectMeta, "gardener.cloud/scaled-to-multi-node") {
```
aaronfern, line 434: Please define a constant for this annotation and use that constant in the check here

pkg/server/backuprestoreserver.go
```diff
@@ -224,22 +224,19 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
 		}
+		_, err = m.UpdateMemberPeerURL(ctx, cli)
+		return err
+	})
+	if err != nil {
+		b.logger.Error("unable to update the member")
```
aaronfern, line 239:
```suggestion
		b.logger.Error("unable to update the member: ", err)
```

aaronfern (changes requested):
Thanks for the PR @unmarshall!
Few suggestions from me, looks good otherwise

pkg/server/backuprestoreserver.go
unmarshall: done

pkg/miscellaneous/miscellaneous.go
unmarshall: done

unmarshall:
> Have you also tested the single member restoration feature with these changes ?
@ishan16696  no i have not tested that explicitly. Let us sync up tomorrow and test it out.

pkg/server/backuprestoreserver.go
```diff
@@ -224,24 +210,28 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
-		m := member.NewMemberControl(b.config.EtcdConnectionConfig)
-		err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
-			return err != nil
-		}, func() error {
-			cli, err := etcdutil.NewFactory(*b.config.EtcdConnectionConfig).NewCluster()
-			if err != nil {
-				return err
-			}
-			_, err = m.UpdateMemberPeerURL(ctx, cli)
+	}
```
abdasgupta, line 213: We don't need this block of promoting member. We are already promoting member downwards. Can we remove it?

abdasgupta (changes requested):

pkg/server/backuprestoreserver.go
unmarshall: Yes this was removed in this PR. Maybe i did not understand your suggestion. If you view the file you will see that this block was removed. Now we always want to promote a member irrespective of cluster size.

ishan16696 (changes requested):
I have tested the PR for single member restoration. It is working fine.
Just a one nitpicks.

pkg/server/backuprestoreserver.go
```diff
@@ -62,7 +62,7 @@ var (
 	// runServerWithSnapshotter indicates whether to start server with or without snapshotter.
 	runServerWithSnapshotter bool = true
 	retryTimeout                  = 5 * time.Second
-	peerURLTLSEnabled        bool
+	//peerURLTLSEnabled        bool
```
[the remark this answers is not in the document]
unmarshall: done

pkg/server/backuprestoreserver.go
ishan16696: I think what @abdasgupta is saying that we are promoting the learner in leader-election as well. check [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/leaderelection/leaderelection.go#L153-L157).
@aaronfern can we remove this from here or may be @abdasgupta you can address this in your PR ?

abdasgupta: Better if you remove PromoteMember from here. But keep the UpdateMemberPeerURL

abdasgupta (approved):

ishan16696 (approved):

aaronfern (approved):
/lgtm

PR #537, opened 2022-09-21
To decrease the likelihood of race condition while closing the snapshotter.

ishan16696:
It has been have observed that there is no need to close the snapshotter in [OnStoppedLeading func(){....}](https://github.com/gardener/etcd-backup-restore/blob/ab224ee318896e90c72b4a6158844bc400feed9e/pkg/server/backuprestoreserver.go#L265-L273) as snapshotter will get close by [handleSsrStopRequest](https://github.com/gardener/etcd-backup-restore/blob/ab224ee318896e90c72b4a6158844bc400feed9e/pkg/server/backuprestoreserver.go#L599) whenever context get canceled due to loss of leadership. Although In [OnStoppedLeading](https://github.com/gardener/etcd-backup-restore/blob/ab224ee318896e90c72b4a6158844bc400feed9e/pkg/server/backuprestoreserver.go#L265-L273) we are taking the mutex lock before accessing the resource but still I would suggest it is better to remove it as it decreases the likelihood of race condition while closing the snapshotter.

This PR also adds some useful logging.

Fixes #

**Release note**:
```improvement operator
Decreases the likelihood of potential race condition between the go-routines while closing the snapshotter.
```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -263,14 +263,8 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 			}
 		},
 		OnStoppedLeading: func() {
-			// stops the running snapshotter
 			if runServerWithSnapshotter {
-				ssr.SsrStateMutex.Lock()
-				defer ssr.SsrStateMutex.Unlock()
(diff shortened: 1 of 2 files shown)


aaronfern (approved):
/lgtm

PR #540, opened 2022-09-21
UpdateMemberPeerURL only logs an error in case of failure

### LINKED ISSUE #539: [BUG] `UpdateMemberPeerURL` should just log an error message instead of returning the error
**Describe the bug:**
In recent `etcd-backup-restore` releases, a change was introduced where `etcd-backup-restore` attempts to update the member URL. This is needed as members need a specific peer URL and older clusters may still have an outdated URL.
Whilst `etcd-backup-restore` reattempts the `UpdateMemberPeerURL` call in case of failures, it does so only a finite number of times and then [returns an error](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L228-L230). This error propagates upwards and causes [a fatal error](https://github.com/gardener/etcd-backup-restore/blob/master/cmd/server.go#L55) which then causes the container to crash.
This behaviour can severely impact restoration of clusters that have a lot of delta snapshots as the backuprestoreserver can now return this error before etcd recovers from snapstore and can cause restoration to fail.

I would propose to Just log an error in the logs and let `backup-restore` continue. If the cluster is restoring, the restoration logic will take care of [updating the member URL](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/restorer/restorer.go#L130)

**Expected behavior:**
`UpdateMemberPeerURL` not causing the container to crash in event of an error

aaronfern:
This PR updated the functionality of etcd member peer URL updation to only log an error in case of failure.
This is needed as sometimes etcd restoration takes longer to complete and etcd member peer URL updation failure will then cause the container to crash before the restoration can complete

Fixes #539 

**Release note**:

```improvement operator
[bug-fix] backup-restore does not return error when it fails to update PeerURL of member.
```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -226,7 +226,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 		return nil
 	})
 	if err != nil {
-		return fmt.Errorf("failed to update member peer url: %w", err)
+		b.logger.Errorf("failed to update member peer url: %v", err)
 	}
 
 	peerURLTLSEnabled := b.isPeerURLTLSEnabled(memberPeerURL)


pkg/server/backuprestoreserver.go
```diff
@@ -226,7 +226,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 		return nil
 	})
 	if err != nil {
-		return fmt.Errorf("failed to update member peer url: %w", err)
+		b.logger.Errorf("failed to update member peer url: %w", err)
```
ishan16696, line 229: pipeline checks fails with this `pkg/server/backuprestoreserver.go:229:3: Errorf call has error-wrapping directive %w, which is only supported by Errorf`
I think this is due to this issue: https://github.com/golang/go/issues/47641
```suggestion
		b.logger.Errorf("failed to update member peer url: %v", err)
```

ishan16696 (changes requested):

pkg/server/backuprestoreserver.go
aaronfern: Thanks for pointing out
Updated

ishan16696 (approved):

ishan16696:
Please add a appropriate release note:
`Fix a bug: backup-restore shouldn’t return error when it fails to update PeerURL of member.`

ishan16696:
I'm unable to retrigger the pipeline, please do the force push to re-trigger the pipeline checks.

PR #546, opened 2022-10-06
Removes the closing of snapshotter during initialization.

ishan16696:
It has been observed that whenever etcd goes down or restarted then [initialization](https://github.com/gardener/etcd-backup-restore/blob/bd4f088085e43aaa23101e1b98699a2b626532d1/pkg/server/httpAPI.go#L225) request closes the snapshotter but there is no need of that as snapshotter will get close by [handleSsrStopRequest](https://github.com/gardener/etcd-backup-restore/blob/bd4f088085e43aaa23101e1b98699a2b626532d1/pkg/server/backuprestoreserver.go#L593) if `backup-leader` moves to `UnkownState` and cancels the context as etcd is down.
This PR removes the closing of `snapshotter` during initialization phase to removes the redundancy of closing of snapshotter.

Fixes #

@unmarshall 

**Release note**:
```improvement operator
Removes the redundant closing of snapshotter during initialization.
```

### DIFF
--- a/pkg/server/httpAPI.go
+++ b/pkg/server/httpAPI.go
@@ -28,7 +28,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
 	etcdclient "github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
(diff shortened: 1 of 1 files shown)


ishan16696:
PR is ready for review.
@shreyas-s-rao It would be nice if you also take a look as we are removing some legacy code in this PR.

cc @aaronfern @unmarshall @abdasgupta

pkg/server/httpAPI.go
```diff
@@ -233,15 +232,7 @@ func (h *HTTPHandler) serveInitialize(rw http.ResponseWriter, req *http.Request)
...
-			// This is needed to stop the currently running snapshotter.
-			if h.Snapshotter != nil {
-				h.SetStatus(http.StatusServiceUnavailable)
-				atomic.StoreUint32(&h.AckState, HandlerAckWaiting)
-				h.Logger.Info("Changed handler state.")
-				h.ReqCh <- emptyStruct
-				h.Logger.Info("Waiting for acknowledgment...")
-				<-h.AckCh
-			}
+			h.SetStatus(http.StatusServiceUnavailable)
```
aaronfern, line 235: Shouldn't the status be set to `Unavailable` only when `h.Snapshotter` is present?

ishan16696: it doesn't matter.
For single node case: we know `h.Snapshotter!=nil` is always true as it is the etcd-leader.
For multi-node: we are not using the `/healthz` endpoint as readinessProbe, so it is not used anywhere.

abdasgupta (approved):
Looks good

aaronfern (approved):

ishan16696:
waiting for @shreyas-s-rao review as it changes legacy code.
/hold

shreyas-s-rao (approved):
@ishan16696 Thanks for the PR. LGTM
/lgtm

Issue #553, opened 2022-11-02
[BUG] Race-condition between closing the snapshotter and taking out-of-sync snapshot.


ishan16696:
**Describe the bug:**
We have observed the race-condition is happening while closing the `snapshotter` if out-of-sync snapshot is also get triggers at the same time which leads to deadlock.
When [snapshotEventHandler](https://github.com/gardener/etcd-backup-restore/blob/84da9e978dac1509cf277e1e03bb58384f9e2671/pkg/snapshot/snapshotter/snapshotter.go#L600) get close and [out-of-schedule full snapshot](https://github.com/gardener/etcd-backup-restore/blob/84da9e978dac1509cf277e1e03bb58384f9e2671/pkg/snapshot/snapshotter/snapshotter.go#L200) triggers at the same time then [TriggerFullSnapshot](https://github.com/gardener/etcd-backup-restore/blob/84da9e978dac1509cf277e1e03bb58384f9e2671/pkg/snapshot/snapshotter/snapshotter.go#L200) takes  SsrStateMutex and tries to send a signal to this [fullSnapshotReqCh](https://github.com/gardener/etcd-backup-restore/blob/84da9e978dac1509cf277e1e03bb58384f9e2671/pkg/snapshot/snapshotter/snapshotter.go#L208) channel where nobody is reading it as [snapshotEventHandler](https://github.com/gardener/etcd-backup-restore/blob/84da9e978dac1509cf277e1e03bb58384f9e2671/pkg/snapshot/snapshotter/snapshotter.go#L600) already got closed, hence TriggerFullSnapshot never release SsrStateMutex as well.

**RCA of bug:**
It has been observed that during the maintenance window of cluster defragmentation of etcd started.
```
2022-11-01 04:13:00 | {"log":"Defragmenting etcd member[https://etcd-main-0.etcd-main-peer.shoot--garden--aws-eu9.svc:2379]","severity":"INFO"}
2022-11-01 04:13:00 | {"log":"Starting the defragmentation on etcd leader","severity":"INFO"}
2022-11-01 04:13:00 | {"log":"Starting the defragmentation as all members of etcd cluster are in healthy state","severity":"INFO"}
2022-11-01 04:13:00 | {"log":"All etcd members endPoints: [https://etcd-main-0.etcd-main-peer.shoot--garden--aws-eu9.svc:2379]","severity":"INFO"}
```

[...]

PR #554, opened 2022-11-03
Removes taking full snapshot after defragmentation to avoid race-condintion.

ishan16696:
This PR removes triggering the out-of-schedule full snapshot after defragmentation as triggering the out-of-schedule full snapshot and closing of snapshotter can cause potential deadlock situation.

temporary fixes: [issue: 553](https://github.com/gardener/etcd-backup-restore/issues/553)

**Release note**:
```bugfix operator
To avoid race-condition between closing of snapshotter and taking snapshot after defrag, removing the out-of-schedule full snapshot triggered after defragmentation of etcd.
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -241,6 +241,7 @@ func (ssr *Snapshotter) stop() {
 		ssr.deltaSnapshotTimer.Stop()
 		ssr.deltaSnapshotTimer = nil
 	}
+	ssr.SetSnapshotterInactive()
 	ssr.closeEtcdClient()
 }
 
@@ -697,9 +698,7 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 
 		case <-stopCh:
 			ssr.logger.Info("Closing the Snapshot EventHandler.")
-			leaseUpdateCancel()
 			ssr.cleanupInMemoryEvents()
-			ssr.SetSnapshotterInactive()
 			return nil
 		}
 	}
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -232,7 +232,7 @@ func (ssr *Snapshotter) TriggerDeltaSnapshot() (*brtypes.Snapshot, error) {
 // not have any effect.
 func (ssr *Snapshotter) stop() {
 	ssr.logger.Info("Closing the Snapshotter...")
-	ssr.SsrStateMutex.Lock()
+
 	if ssr.fullSnapshotTimer != nil {
 		ssr.fullSnapshotTimer.Stop()
 		ssr.fullSnapshotTimer = nil
@@ -242,9 +242,20 @@ func (ssr *Snapshotter) stop() {
 		ssr.deltaSnapshotTimer = nil
 	}
 	ssr.closeEtcdClient()
+}
 
+// SetSnapshotterInactive set the snapshotter state to Inactive.
+func (ssr *Snapshotter) SetSnapshotterInactive() {
+	ssr.SsrStateMutex.Lock()
+	defer ssr.SsrStateMutex.Unlock()
 	ssr.SsrState = brtypes.SnapshotterInactive
-	ssr.SsrStateMutex.Unlock()
+}
+
+// SetSnapshotterActive set the snapshotter state to active.
+func (ssr *Snapshotter) SetSnapshotterActive() {
+	ssr.SsrStateMutex.Lock()
+	defer ssr.SsrStateMutex.Unlock()
+	ssr.SsrState = brtypes.SnapshotterActive
 }
 
 func (ssr *Snapshotter) closeEtcdClient() {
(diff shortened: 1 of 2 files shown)


ishan16696:
Can anyone of you review this PR @aaronfern @abdasgupta ?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -242,9 +242,20 @@ func (ssr *Snapshotter) stop() {
...
+// SetSnapshotterInactive set the snapshotter state to Inactive.
+func (ssr *Snapshotter) SetSnapshotterInactive() {
+	ssr.SsrStateMutex.Lock()
+	defer ssr.SsrStateMutex.Unlock()
 	ssr.SsrState = brtypes.SnapshotterInactive
-	ssr.SsrStateMutex.Unlock()
+}
+
+// SetSnapshotterActive set the snapshotter state to active.
+func (ssr *Snapshotter) SetSnapshotterActive() {
```
abdasgupta, line 255: Who is calling this?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -242,9 +242,20 @@ func (ssr *Snapshotter) stop() {
 		ssr.deltaSnapshotTimer = nil
 	}
 	ssr.closeEtcdClient()
+}
 
+// SetSnapshotterInactive set the snapshotter state to Inactive.
+func (ssr *Snapshotter) SetSnapshotterInactive() {
```
unmarshall, line 248: So there are a few things that should be corrected:

**Method:** snapshotter.snapshotEventHandler
```go
defer leaseUpdateCancel()
...
case <-stopCh:
	ssr.logger.Info("Closing the Snapshot EventHandler.")
	leaseUpdateCancel()
	ssr.cleanupInMemoryEvents()
	return nil
}
```
function `leaseUpdateCancel` is called twice. Not necessary.
2. So consider the following:
    1. Backup-restore loses leadership. As a result the leCtx gets cancelled. In method `handleSsrStopRequest` an empty struct will be sent over the ssr stop channel.
    2. As a result ssr.snapshotEventHandler will exit and the status of the ssr will still be active.
    3. Now ssr.TriggerFullSnapshot is invoked. It will be able to take the lock as currently no one is holding it and its status would still be active. It will now send a message on `ssr.fullSnapshotAckCh` with no receiver. This will block.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -242,9 +242,20 @@ func (ssr *Snapshotter) stop() {
...
+	ssr.SsrStateMutex.Lock()
+	defer ssr.SsrStateMutex.Unlock()
 	ssr.SsrState = brtypes.SnapshotterInactive
-	ssr.SsrStateMutex.Unlock()
+}
+
+// SetSnapshotterActive set the snapshotter state to active.
+func (ssr *Snapshotter) SetSnapshotterActive() {
+	ssr.SsrStateMutex.Lock()
+	defer ssr.SsrStateMutex.Unlock()
```
unmarshall, line 257: instead of using mutex to change a field in a struct you should perhaps use `atomic.Bool.CompareAndSwap` instead. Usage of `sync.Mutex`

pkg/snapshot/snapshotter/snapshotter.go
unmarshall: If this method is not used then please remove it as it is dead code.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: I think atomic operations are costly operations as it involves system call, so if we can achieve something without kernel involvement then why not use it, that's why I used mutex instead of atomic ops. WDYT?

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: > function leaseUpdateCancel is called twice. Not necessary.

removed

> * Backup-restore loses leadership. As a result the leCtx gets cancelled. In method handleSsrStopRequest an empty struct will be sent over the ssr stop channel.
> * As a result ssr.snapshotEventHandler will exit and the status of the ssr will still be active.
 >* Now ssr.TriggerFullSnapshot is invoked. It will be able to take the lock as currently no one is holding it and its status would still be active. It will now send a message on ssr.fullSnapshotReqCh with no receiver. This will block. Please double check if this can happen.

as first safety measure I have removed the invokation of `ssr.TriggerFullSnapshot` after defrag call.
as secondary safety measure, I'm setting the snapshottter state to `Inactive` before exiting snapshotEventHandler loop.

ishan16696: > It is also incorrect to create a channel in OnStartedLeading callback. It should be created as part of snapshotter.NewSnapshotter call. The channel belongs to snapshotter and should not be created outside.

but we are passing the responsibility of closing the snapshottter to `handleSsrStopRequest`, that's why we are creating channel in OnStartedLeading.

pkg/snapshot/snapshotter/snapshotter.go
unmarshall: https://www.sixt.tech/golangs-atomic - atomics are supported by native cpu architectures and are generally faster as compared to mutex. They do not completely replace mutex as such but there is an overlap.

ishan16696:
@unmarshall and I decided for now we can go with this temporary fix for time being as proper fix might need some refactoring  of snapshotter logic and proper testing.

abdasgupta (approved):
Good PR

PR #555, opened 2022-11-10
Remove owner checks

plkokanov:
This PR removes the owner checks from the `backup-restore` that were used to restart the `etcd` process that runs in the source `Seed` cluster during "bad case" control plane migration. As part of the simplification efforts from https://github.com/gardener/gardener/issues/6302 it was decided that "bad case" control plane migration will most likely never be used when the HA gardener topic is ready.

Part of https://github.com/gardener/gardener/issues/6302

This PR depends on https://github.com/gardener/etcd-druid/pull/461 as we do not want to accidentally set owner check flags without the `backup-restore` actually supporting them.

**Release note**:

```improvement operator
Removed owner checks that were used to restart the `etcd` process that runs in the source `Seed` cluster during "bad case" control plane migration.
```


pkg/server/init.go
```diff
@@ -36,9 +36,7 @@ func NewBackupRestoreComponentConfig() *BackupRestoreComponentConfig {
 		SnapstoreConfig:          snapstore.NewSnapstoreConfig(),
 		CompressionConfig:        compressor.NewCompressorConfig(),
 		RestorationConfig:        brtypes.NewRestorationConfig(),
-		OwnerCheckConfig:         brtypes.NewOwnerCheckConfig(),
 		DefragmentationSchedule:  defaultDefragmentationSchedule,
-		EtcdProcessName:          defaultEtcdProcessName,
```
ishan16696, line 41: Please also remove `defaultEtcdProcessName` whereever it is defined.

ishan16696 (changes requested):
Overall looks good, just 1 nit

pkg/server/init.go
ishan16696: It is define here: https://github.com/gardener/etcd-backup-restore/blob/d57caede0507002c95181d755055959a23dd3595/pkg/server/types.go#L25

ishan16696 (approved):
As we are about to start working on refactoring of snapshotter logic, our requirement is to decrease the code base, so I'm approving this PR. Whatever the review comments are left, I will address them by myself in separate PR. 
LGTM!!

PR #564, opened 2022-12-22
Fix the bug of falsely wrong annotation added to etcd-member lease of TLS not enabled.

### LINKED ISSUE #562: [BUG] Falsely wrong annotation is added to etcd-member lease of TLS not enabled
**Describe the bug:**
A falsely wrong annotation has been added to etcd-member lease of TLS not enabled `member.etcd.gardener.cloud/tls-enabled: "false"` this annotation is read by etcd-druid which hence druid thinks this is [peerUrlTLSChangedToEnabled](https://github.com/gardener/etcd-druid/blob/6f084d8ba14985c80ae1b3edd853db851dab7957/controllers/etcd_controller.go#L612-L617) case, therefore druid unnecessary recreates etcd statefulset to handle this `PeerURL TLS changed to enable` case.

**RCA:**
When [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/0bba7b21a1bf057eab367f07942391f52a69e4c2/pkg/server/backuprestoreserver.go#L208) returns an error (may be due to service endpoint not available or restoration is going on i.e etcd is not up) then `memberPeerURL` is set to empty string after that [isPeerURLTLSEnabled](https://github.com/gardener/etcd-backup-restore/blob/0bba7b21a1bf057eab367f07942391f52a69e4c2/pkg/server/backuprestoreserver.go#L218) is called which takes `memberPeerURL` as an argument which is empty hence func `isPeerURLTLSEnabled()` returns false, hence wrong annotation of `tls-enabled: "false"` is set to lease-object.

Logs of backup-restore:
```
k logs etcd-main-0 -c backup-restore | grep "failed to update member peer url"                        
time="2022-12-14T13:54:35Z" level=error msg="failed to update member peer url: error listing members: context deadline exceeded" actor=backup-restore-server
```

**Expected behavior:**

1. Mimic the quorum loss scenario, then tried to recover from quorum loss using the steps as mentioned in the [playbook](https://github.com/gardener/etcd-druid/blob/master/docs/operation/Recover_From_Etcd_Permanent_Quorum_Loss.md).
2. From moving 0->1 replica, etcd is successfully restored.
3. But when I removed the ignore annotation from etcd CR and add reconcile annotation to the etcd CR, then druid should do the scale up of etcd replicas from 1->3.
4.  But instead of scale-up scenario, druid recreates the etcd statefulset, hence delete the already running pod, and then try to bring all 3 pods together which is not possible for obvious reasons.

**Logs of etcd-druid:**
```
2022-12-14T12:35:16.869Z	INFO	custodian-controller	Requeue item because of last error: flow "(etcd: d8c9cdb1-03ee-47e4-8da4-46127cd7181c) Deploy Flow for StatefulSet etcd-main for Namespace: shoot--i539698--test-ha" encountered task errors: [task "(recreate-sts): create sts due to peer url tls with replicas: 3" failed: retry failed with context deadline exceeded, last error: not enough ready replicas (0/3)]	{"etcd": "shoot--i539698--test-ha/etcd-main"}
2022-12-14T12:35:20.398Z	INFO	custodian-controller	Requeue item because of last error: flow "(etcd: d8c9cdb1-03ee-47e4-8da4-46127cd7181c) Deploy Flow for StatefulSet etcd-main for Namespace: shoot--i539698--test-ha" encountered task errors: [task "(recreate-sts): create sts due to peer url tls with replicas: 3" failed: retry failed with context deadline exceeded, last error: not enough ready replicas (0/3)]	{"etcd": "shoot--i539698--test-ha/etcd-main"}
```

 - Etcd version/commit ID :

ishan16696:
This PR fixes the bug of false wrong annotation `member.etcd.gardener.cloud/tls-enabled` has been added to etcd's member lease object of TLS not enabled which gives wrong information to etcd-druid and etcd-druid consider this as a case of Peer TLS not enabled, hence druid recreates the etcd statefulset. This PR will save backup-restore from this kind of scenario.

Fixes #562 

**Release note**:
```bugfix operator
Fixes bug of false wrong annotation added to etcd-member lease of TLS not enabled.
```


ishan16696:
I have now resolved the merge conflicts. PTAL

pkg/miscellaneous/miscellaneous.go
```diff
@@ -520,3 +522,31 @@ func ParsePeerURL(initialAdvertisePeerURLs, podName string) (string, error) {
 	domaiName := fmt.Sprintf("%s.%s.%s", tokens[1], tokens[2], "svc")
 	return fmt.Sprintf("%s://%s.%s:%s", tokens[0], podName, domaiName, tokens[3]), nil
 }
+
+// IsPeerURLTLSEnabled checks whether the peer address is TLS enabled or not.
+func IsPeerURLTLSEnabled() (bool, error) {
```
unmarshall, line 527: This is not sufficient the last time i checked. Since CM is mounted etcd-backup-restore will get an updated configuration which only captures the desired state as set in the etcd resource. However this does not correctly reflect the state of the actual/current peer URL TLS protocol which can only be fetched by checking the member peer URL.

pkg/member/member_control.go
```diff
@@ -46,7 +46,7 @@ type Control interface {
 	PromoteMember(context.Context) error
 
 	// UpdateMemberPeerURL updates the peer address of a specified etcd cluster member.
-	UpdateMemberPeerURL(context.Context, client.ClusterCloser) (string, error)
+	UpdateMemberPeerURL(context.Context, client.ClusterCloser) error
```
unmarshall, line 49: It is not clear on why this change has been made. If the UpdateMemberPeerURL returns an error then the status of TLS should be unknown which can also be explicitly set. Druid can then appropriately handle this state and not take any action and instead re-queue the item till a time it can determine the state.

pkg/miscellaneous/miscellaneous.go
ishan16696: I'm also not doing anything different from the current code. Let me explain:
Initially I also thought that we are getting peerURL using some api call to Etcd but later I found that's not the case.
If you look at the current code [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L208-L218), it is calling [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L246) to get peerURL, which calls [doUpdateMemberPeerAddress](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L188) which is getting the peerURL from configMap check [here](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L174) , so I felt there is no need for any api call required just to check the TLS is enabled or not.
So, I separated the functionality of updating the PeerURL and check the TLS is enabled or not.

pkg/member/member_control.go
ishan16696: same reason https://github.com/gardener/etcd-backup-restore/pull/564#discussion_r1059844968

pkg/miscellaneous/miscellaneous.go
unmarshall: There is a slight difference. In [doUpdateMemberPeerAddress](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L188-L206) the member peer URL read from the config is returned only when the same URL is successfully updated by calling `cli.MemberUpdate(memberUpdateCtx, id, []string{memberPeerURL})` else an error is returned.

ishan16696: > cli.MemberUpdate(memberUpdateCtx, id, []string{memberPeerURL})

this api call will never be successful if restoration of etcd will be going on, then [doUpdateMemberPeerAddress](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L188-L206) will return empty string which leads to this issue.
Here, I'm not using `PeerURL`  for anything apart from checking whether it is `http` or `https` .
Can you give me the scenario of this edge where this can cause issue as may be I don't have much of a insight on this whole TLS not enabled -> TLS enabled case scenario ?

pkg/server/backuprestoreserver.go
```diff
@@ -197,25 +192,26 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
 		}
 		return nil
-	})
-	if err != nil {
+	}); err != nil {
 		b.logger.Errorf("failed to update member peer url: %v", err)
 	}
 
-	peerURLTLSEnabled := b.isPeerURLTLSEnabled(memberPeerURL)
+	peerURLTLSEnabled, err := miscellaneous.IsPeerURLTLSEnabled()
```
aaronfern, line 211: Can you consider the error thrown by `m.UpdateMemberPeerURL()` before this function is called?
If `m.UpdateMemberPeerURL()` returns an error then I don't think we should call `miscellaneous.IsPeerURLTLSEnabled()`

aaronfern (changes requested):
Thanks for the PR! 
Just one comment, rest all looks good to me

pkg/server/backuprestoreserver.go
ishan16696: `IsPeerURLTLSEnabled` is now independent of `UpdateMemberPeerURL`.
>Can you consider the error thrown by m.UpdateMemberPeerURL() before this function is called?
If m.UpdateMemberPeerURL() returns an error then I don't think we should call miscellaneous.IsPeerURLTLSEnabled()

that's the whole purpose of this PR, if restoration is currently going on then definitely `UpdateMemberPeerURL` will fail but that shouldn't affect our decision of checking peerURLTLSEnabled or not.
And TBH I want to remove this whole `UpdateMemberPeerURL` logic from here as I think we don't need a backward compatibility now. right ?
Anyway `UpdateMemberPeerURL`is called after restoration.

ishan16696: ok, I got your concern. Let me explain this:
First we are not doing anything with this PeerURL which I read from config file, I'm just interested in checking whether it is `http` or `https` , which I get from the config file the same way [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L188-L193) get PeerURL to update.
Now take each scenarios:
**case 1:** now when restoration fails and [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/snapshot/restorer/restorer.go#L124-L131) didn't get execute.
then there will be no annotation added to member lease as member lease updation will get stop by backup-restore as etcd is not running.

**case 2:** replica=1, when restoration succeeds but [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/snapshot/restorer/restorer.go#L124-L131) fails
then it won't matter as it is the case of single node, no peer communication is required.

**case 3:** scale-up=0->1->3, when restoration succeeds in `0->1` but [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/snapshot/restorer/restorer.go#L124-L131) fails
then as I mentioned earlier that I'm not doing anything with this PeerURL which I get from the config file the same way [UpdateMemberPeerURL](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L188-L193) get PeerURL to update. Suppose `UpdateMemberPeerURL` succeeds then it would be the same peerURL on that basis I have calculated whether the TLS enable or not.
But `UpdateMemberPeerURL` fails, so  moving from `replica 1->replicas 3` won't be possible as this member will not be able to join the cluster, as it requires to have a peer communication. then I guess we have bigger problem which is not limited to annotation. This Problem is out-of-scope of this PR.

ishan16696: Can you give me the edge case which is handled by the current code but not handled by this PR and where this can cause issue as may be I don't have much of a insight on this whole TLS not enabled -> TLS enabled scenario ?
If you have new edge case which is currently not handled by current code as well as by this PR then I would suggest open a new issue describing the edge case. Hope that's fine.

aaronfern (approved):

pkg/miscellaneous/miscellaneous.go
unmarshall: So let us assume that the configuration has been updated and as a result the ConfigMap change reaches backup-restore. The change is that peer URL has been TLS enabled. In etcd-controller (etcd-druid), the configmap got updated but due to an error the event was re-queued. In the mean time backup-restore got the new config. In your code above this value was read and on the lease it was set as true against the annotation.

Now when etcd-controller finally services the re-queued event then it will find that this has been enabled on the member and will do nothing when it should have actually deleted the STS.

ishan16696: If that's the case then this edge is also applicable to the current code. IMO, current code is also not handling this edge case. If I'm wrong then please explain how's the current code handling this edge case.
Nevertheless, I would suggest to open new issue with whatever you have describe above as handling of this edge case is not in scope of this PR.

ishan16696: Various edge cases I have described here: https://github.com/gardener/etcd-backup-restore/pull/564#discussion_r1070760238

ishan16696: > In your code above this value was read and on the lease it was set as true against the annotation.

actually no, lease annotation will still be false as we are checking TLSEnable case only on startup (in current code as well as in this PR) and passing the boolean variable to wherever it requires.

ishan16696:
Moved the checking PeerURLTLSEnabled inside `RenewMemberLeasePeriodically`. PTAL

cc @aaronfern @unmarshall

PR #574, opened 2023-01-12
Enhances the decision to take full snapshot during startup to avoid missing of any full-snapshot.

### LINKED ISSUE #570: Enhance condition to take full snapshot during startup.
/area monitoring
/area backup
/kind enhancement

During startup, etcd-backup-restore should also consider configured FullSnapshotSchedule along with [timeStamp of last full snapshot](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/server/backuprestoreserver.go#L372-L374), so that it won't missed any full snapshot.

**Why is this needed**:
Currently, alert [KubeEtcdFullBackupFailed](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/component/etcd/monitoring.go#L189-L190) is calculated based on Last FullSnapshot Timestamp, so it checks whether `backup-restore` has taken a full snapshot within last 24h or not. But we have observed some **false** `KubeEtcdFullBackupFailed` alerts for some shoots if shoot is hibernated and wake up within 24h and [determineBackupSchedule](https://github.com/gardener/gardener/blob/387a2cca9683ec2ffdaa3cb0e6c67f46613d343a/pkg/operation/botanist/etcd.go#L194) is calculating `FullSnapshotSchedule` on the basis of maintenance window. 

Take this Scenario: 
 1.  Suppose, shoot have maintenance window `m1`.
 2. `etcd-backup-restore` took a full snapshot at timestamp `t1` .
 3. Then the cluster was hibernated at `t2`(before maintenance window i.e t2<m1) and woken up again at `t3`. (t3>t2>t1 && t3>m1)
 4. When Cluster was woken up, `t3-t1`< `24h` (24h didn’t pass yet), so no new full snapshot had taken by backup-restore as it [calculates 23.5h from timeStamp of last full snapshot](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/server/backuprestoreserver.go#L372-L374).
 4. Past `24h`, at `t4`  alert `KubeEtcdFullBackupFailed` checks the last timestamp of full snapshot and it found `t4-t1>24h`, now an alert has been raised and backup-restore is waiting to take full snapshot according to `--schedule` passed to backup-restore and it was calculated on basis of maintenance window of shoot (`m1` of next day) which is not yet reached.

/cc @timuthy @shreyas-s-rao

ishan16696:
This PR enhances the condition of checking whether to take a full snapshot or not during startup of backup-restore to avoid missing of any full-snapshot within 24hrs of window.

Fixes #570 

**Release note**:
```improvement operator
Enhances the decision to take full snapshot during startup of etcd-backup-restore to avoid missing of any full-snapshot.
```

### DIFF
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -561,5 +561,6 @@ func GetPrevDayScheduledSnapTime(nextSnapSchedule time.Time) time.Time {
 		nextSnapSchedule.Minute(),
 		nextSnapSchedule.Second(),
 		nextSnapSchedule.Nanosecond(),
-		nextSnapSchedule.Location())
+		nextSnapSchedule.Location(),
+	)
 }
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -357,7 +357,6 @@ func IsMultiNode(logger *logrus.Entry) bool {
 	}
 
 	config := map[string]interface{}{}
-	err = yaml.Unmarshal([]byte(configYML), &config)
 	if err := yaml.Unmarshal([]byte(configYML), &config); err != nil {
 		return false
 	}
@@ -551,13 +550,14 @@ func IsPeerURLTLSEnabled() (bool, error) {
 	return peerURL.Scheme == https, nil
 }
 
-// GetPrevDayScheduledSnapTime returns the previous day schedule snapshot time.
-func GetPrevDayScheduledSnapTime(nextSnapSchedule time.Time) time.Time {
+// GetPrevScheduledSnapTime returns the previous schedule snapshot time.
+// TODO: Previous full snapshot time should be calculated on basis of previous cron schedule of full snapshot.
+func GetPrevScheduledSnapTime(nextSnapSchedule time.Time, timeWindow float64) time.Time {
 	return time.Date(
 		nextSnapSchedule.Year(),
 		nextSnapSchedule.Month(),
-		nextSnapSchedule.Day()-1,
-		nextSnapSchedule.Hour(),
+		nextSnapSchedule.Day(),
+		nextSnapSchedule.Hour()-int(timeWindow),
 		nextSnapSchedule.Minute(),
 		nextSnapSchedule.Second(),
 		nextSnapSchedule.Nanosecond(),
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -550,3 +550,16 @@ func IsPeerURLTLSEnabled() (bool, error) {
 
 	return peerURL.Scheme == https, nil
 }
+
+// GetPrevDayScheduledSnapTime returns the previous day schedule snapshot time.
+func GetPrevDayScheduledSnapTime(nextSnapSchedule time.Time) time.Time {
+	return time.Date(
+		nextSnapSchedule.Year(),
+		nextSnapSchedule.Month(),
+		nextSnapSchedule.Day()-1,
+		nextSnapSchedule.Hour(),
+		nextSnapSchedule.Minute(),
+		nextSnapSchedule.Second(),
+		nextSnapSchedule.Nanosecond(),
+		nextSnapSchedule.Location())
+}
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -307,8 +307,8 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 // for the case when backup-restore becomes leading sidecar.
 func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Context, handler *HTTPHandler, ssr *snapshotter.Snapshotter, ss brtypes.SnapStore, ssrStopCh chan struct{}, ackCh chan struct{}) {
 	var (
-		err                       error
-		initialDeltaSnapshotTaken bool
+		err                      error
+		initialFullSnapshotTaken bool
 	)
 
 	for {
@@ -354,8 +354,31 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 			// is taken and the regular snapshot schedule comes into effect.
 
 			fullSnapshotMaxTimeWindowInHours := ssr.GetFullSnapshotMaxTimeWindow(b.config.SnapshotterConfig.FullSnapshotSchedule)
-			initialDeltaSnapshotTaken = false
-			if ssr.PrevFullSnapshot != nil && !ssr.PrevFullSnapshot.IsFinal && !ssr.WasScheduledFullSnapshotMissed(fullSnapshotMaxTimeWindowInHours) && !ssr.IsNextFullSnapCrossTimeWindow(fullSnapshotMaxTimeWindowInHours) {
+			initialFullSnapshotTaken = false
+			if ssr.PrevFullSnapshot == nil || ssr.PrevFullSnapshot.IsFinal || ssr.IsTakingFullSnapRequiresAtStartup(fullSnapshotMaxTimeWindowInHours) {
+				// need to take a full snapshot here
+				var snapshot *brtypes.Snapshot
+				metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindDelta}).Set(0)
+				metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull}).Set(1)
+				if snapshot, err = ssr.TakeFullSnapshotAndResetTimer(false); err != nil {
+					metrics.SnapshotterOperationFailure.With(prometheus.Labels{metrics.LabelError: err.Error()}).Inc()
+					b.logger.Errorf("Failed to take substitute first full snapshot: %v", err)
+					continue
+				}
+				initialFullSnapshotTaken = true
+				if b.config.HealthConfig.SnapshotLeaseRenewalEnabled {
+					leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
+					defer cancel()
+					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName, b.config.HealthConfig.DeltaSnapshotLeaseName); err != nil {
+						b.logger.Warnf("Snapshot lease update failed : %v", err)
+					}
+				}
+				if b.backoffConfig.Start {
+					b.backoffConfig.ResetExponentialBackoff()
+				}
+			}
+
+			if !initialFullSnapshotTaken {
 				ssrStopped, err := ssr.CollectEventsSincePrevSnapshot(ssrStopCh)
 				if ssrStopped {
 					b.logger.Info("Snapshotter stopped.")
@@ -368,7 +391,6 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 						b.logger.Warnf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
 						continue
 					}
-					initialDeltaSnapshotTaken = true
 					if b.config.HealthConfig.SnapshotLeaseRenewalEnabled {
 						leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
 						defer cancel()
@@ -384,28 +406,6 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 				}
 			}
 
-			if !initialDeltaSnapshotTaken {
-				// need to take a full snapshot here
-				var snapshot *brtypes.Snapshot
-				metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindDelta}).Set(0)
-				metrics.SnapshotRequired.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull}).Set(1)
-				if snapshot, err = ssr.TakeFullSnapshotAndResetTimer(false); err != nil {
-					metrics.SnapshotterOperationFailure.With(prometheus.Labels{metrics.LabelError: err.Error()}).Inc()
-					b.logger.Errorf("Failed to take substitute first full snapshot: %v", err)
-					continue
-				}
-				if b.config.HealthConfig.SnapshotLeaseRenewalEnabled {
-					leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
-					defer cancel()
-					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName, b.config.HealthConfig.DeltaSnapshotLeaseName); err != nil {
-						b.logger.Warnf("Snapshot lease update failed : %v", err)
-					}
-				}
-				if b.backoffConfig.Start {
-					b.backoffConfig.ResetExponentialBackoff()
-				}
-			}
-
 			// Set snapshotter state to Active
 			ssr.SetSnapshotterActive()
 
@@ -416,7 +416,7 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 
 			// Start snapshotter
 			b.logger.Infof("Starting snapshotter...")
-			startWithFullSnapshot := ssr.PrevFullSnapshot == nil || ssr.PrevFullSnapshot.IsFinal || ssr.WasScheduledFullSnapshotMissed(fullSnapshotMaxTimeWindowInHours) || ssr.IsNextFullSnapCrossTimeWindow(fullSnapshotMaxTimeWindowInHours)
+			startWithFullSnapshot := ssr.PrevFullSnapshot == nil || ssr.PrevFullSnapshot.IsFinal || ssr.IsTakingFullSnapRequiresAtStartup(fullSnapshotMaxTimeWindowInHours)
 			if err := ssr.Run(ssrStopCh, startWithFullSnapshot); err != nil {
 				if etcdErr, ok := err.(*errors.EtcdError); ok {
 					metrics.SnapshotterOperationFailure.With(prometheus.Labels{metrics.LabelError: etcdErr.Error()}).Inc()
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -353,16 +353,12 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 			// the delta snapshot memory limit), after which a full snapshot
 			// is taken and the regular snapshot schedule comes into effect.
 
-			// TODO: write code to find out if prev full snapshot is older than it is
-			// supposed to be, according to the given cron schedule, instead of the
-			// hard-coded "24 hours" full snapshot interval
-
 			// Temporary fix for missing alternate full snapshots for Gardener shoots
 			// with hibernation schedule set: change value from 24 ot 23.5 to
 			// accommodate for slight pod spin-up delays on shoot wake-up
 			const recentFullSnapshotPeriodInHours = 23.5
 			initialDeltaSnapshotTaken = false
-			if ssr.PrevFullSnapshot != nil && !ssr.PrevFullSnapshot.IsFinal && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= recentFullSnapshotPeriodInHours {
+			if ssr.PrevFullSnapshot != nil && !ssr.PrevFullSnapshot.IsFinal && !ssr.IsScheduledFullSnapshotMissed() {
 				ssrStopped, err := ssr.CollectEventsSincePrevSnapshot(ssrStopCh)
 				if ssrStopped {
 					b.logger.Info("Snapshotter stopped.")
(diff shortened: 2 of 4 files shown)


ishan16696:
This PR is ready for first review. I will add unit tests. 
Hence /do-not-merge

ishan16696:
I have added the unit tests.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -334,6 +334,94 @@ var _ = Describe("Snapshotter", func() {
...
+				// Previous full snapshot was taken 4hrs 10 mins before startup
+				ssr.PrevFullSnapshot = &brtypes.Snapshot{
+					CreatedOn: time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day(), (currentHour-4)%24, (currentMin-10)%60, 0, 0, time.Local),
+				}
+				isFullSnapMissed := ssr.IsScheduledFullSnapshotMissed()
+				Expect(isFullSnapMissed).Should(BeFalse())
+			})
+		})
+
+		Context("Previous snapshot was taken within 24hrs and gonna miss 24hrs of schedule fullsnapshot window", func() {
```
shreyas-s-rao, line 407: Test spec message is not clear.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -334,6 +334,94 @@ var _ = Describe("Snapshotter", func() {
...
+				// Previous full snapshot was taken 1 days before at exactly at scheduled time
+				ssr.PrevFullSnapshot = &brtypes.Snapshot{
+					CreatedOn: time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day()-1, (currentHour+2)%24, (currentMin+1)%60, 0, 0, time.Local),
+				}
+				isFullSnapMissed := ssr.IsScheduledFullSnapshotMissed()
+				Expect(isFullSnapMissed).Should(BeFalse())
+			})
+		})
+
+		Context("Previous snapshot was taken within 24hrs and not gonna missed schedule fullsnapshot", func() {
```
shreyas-s-rao, line 388: Same as above.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +744,21 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
 	snapstoreHash[ssr.snapstoreConfig.Provider] = newSnapstoreSecretHash
 	return true
 }
+
+// IsScheduledFullSnapshotMissed checked whether the last scheduled full-snapshot is missed or scheduled full-snapshot can be missed.
+func (ssr *Snapshotter) IsScheduledFullSnapshotMissed() bool {
```
shreyas-s-rao, line 749: Please segregate this method into two methods, one to check whether previous scheduled full snapshot was missed, and one to check whether the upcoming scheduled full snapshot might be missed.

```suggestion
// WasScheduledFullSnapshotMissed checks whether the last scheduled full-snapshot was missed or scheduled full-snapshot can be missed.
func (ssr *Snapshotter) WasScheduledFullSnapshotMissed() bool {
```

shreyas-s-rao (changes requested):
@ishan16696 thanks for your PR. I feel the code still assumes that full snapshot is supposed to be taken every 24 hours, irrespective of the cron schedule, and I feel it needs better parsing of the cron schedule (when was the last scheduled cron timestamp, when is the next scheduled cron timestamp). You can probably use something like https://github.com/aptible/supercronic/tree/master/cronexpr to achieve this (which provides `Next` and `NextN` methods, but not `Previous` method so you might have to somehow derive that from existing methods). 

Thanks for adding the tests though.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -520,3 +520,17 @@ func ParsePeerURL(initialAdvertisePeerURLs, podName string) (string, error) {
 	domaiName := fmt.Sprintf("%s.%s.%s", tokens[1], tokens[2], "svc")
 	return fmt.Sprintf("%s://%s.%s:%s", tokens[0], podName, domaiName, tokens[3]), nil
 }
+
+// GetPrevDayScheduledSnapTime returns the previous day schedule snapshot time.
+func GetPrevDayScheduledSnapTime(nextSnapSchedule time.Time) time.Time {
```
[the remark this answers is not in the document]
ishan16696: > What if user sets full snapshot schedule for once in 3 days or once a week? 

user can't set the  full snapshot schedule ... etcd-druid passed a/c to maintenance window right?
Apart from this even if that's the case when `full snapshot schedule` for once in 3 days and backup-restore get restart then we will take full snapshot as it crosses the 24 hr of limit, control won't reach `GetPrevDayScheduledSnapTime`.

ishan16696:
> I feel the code still assumes that full snapshot is supposed to be taken every 24 hours,

yes, it does ... ok I will try to make it more dynamic.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +744,21 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
...
+	now := time.Now()
+	nextSnapSchedule := ssr.schedule.Next(now)
+	timeLeftToTakeNextSnap := nextSnapSchedule.Sub(now)
+
+	if miscellaneous.GetPrevDayScheduledSnapTime(nextSnapSchedule) == ssr.PrevFullSnapshot.CreatedOn {
+		return false
+	}
+
+	// to check whether schedule full-snapshot can be missed.
+	return timeLeftToTakeNextSnap.Hours()+time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() > recentFullSnapshotPeriodInHours
```
unmarshall, line 763: Lets say the next snapshot time is in 5 mins. Will you still trigger a full snapshot? Once you trigger a full snapshot and then when the regular scheduled full snapshot time is reached then will yet another full snapshot be triggered?

pkg/snapshot/snapshotter/snapshotter.go
unmarshall: Yes or rename this method as this method now does not do what the name suggests

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -42,6 +42,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	recentFullSnapshotPeriodInHours = 23.5
```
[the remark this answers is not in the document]
unmarshall: there is already a schedule present, right. Is it possible to derive this from the schedule instead?

pkg/miscellaneous/miscellaneous.go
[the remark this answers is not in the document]
unmarshall: I think Shreyas is pointing to the name of this method which assumes that the schedule is once daily. Once you allow user to specify a schedule you cannot assume that it is once per day and then name of the method as per that assumption.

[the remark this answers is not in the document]
unmarshall: Therefore even the computation done in the method is based on that assumption which is incorrect.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -334,6 +334,94 @@ var _ = Describe("Snapshotter", func() {
...
+				ssr.PrevFullSnapshot = &brtypes.Snapshot{
+					CreatedOn: time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day(), (currentHour-4)%24, (currentMin-10)%60, 0, 0, time.Local),
+				}
+				isFullSnapMissed := ssr.IsScheduledFullSnapshotMissed()
+				Expect(isFullSnapMissed).Should(BeFalse())
+			})
+		})
+
+		Context("Previous snapshot was taken within 24hrs and gonna miss 24hrs of schedule fullsnapshot window", func() {
+			It("should return true", func() {
```
unmarshall, line 408: you should also add a case to test if there was no previous snapshot taken

pkg/server/backuprestoreserver.go
```diff
@@ -362,16 +362,12 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
...
-			// supposed to be, according to the given cron schedule, instead of the
-			// hard-coded "24 hours" full snapshot interval
-
 			// Temporary fix for missing alternate full snapshots for Gardener shoots
 			// with hibernation schedule set: change value from 24 ot 23.5 to
 			// accommodate for slight pod spin-up delays on shoot wake-up
 			const recentFullSnapshotPeriodInHours = 23.5
 			initialDeltaSnapshotTaken = false
-			if ssr.PrevFullSnapshot != nil && !ssr.PrevFullSnapshot.IsFinal && time.Since(ssr.PrevFullSnapshot.CreatedOn).Hours() <= recentFullSnapshotPeriodInHours {
+			if ssr.PrevFullSnapshot != nil && !ssr.PrevFullSnapshot.IsFinal && !ssr.IsScheduledFullSnapshotMissed() {
```
unmarshall, line 370: Using `IsScheduledFullSnapshotMissed` is not the right thing to do here. The earlier condition was sufficient. You should club the new condition with `if !initialDeltaSnapshotTaken` and OR it with this.

ishan16696:
Can you guys take a look at this func: https://github.com/gardener/etcd-backup-restore/pull/574/files#diff-5c59e1a2063c6043cf7cad27aee199b833950c80ec75202ac69699216210a2f1R776-R790.
I will further improve this func if you guys are fine with this kind of approach.

pkg/server/backuprestoreserver.go
ishan16696: > The earlier condition was sufficient

what makes you think that? for that condition this issue of alerts happened.

ishan16696: > You should club the new condition with if !initialDeltaSnapshotTaken and OR it with this.

I don't think so, else we end up taking delta snapshot first then decide whether to take a full snapshot or not.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: It's ok to have a 2 full snapshot rather then no full snapshot for >24hrs and having the alerts as well

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +750,42 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
...
+		return true
+	}
+
+	now := time.Now()
+	nextSnapSchedule := ssr.schedule.Next(now)
+
+	return miscellaneous.GetPrevDayScheduledSnapTime(nextSnapSchedule) != ssr.PrevFullSnapshot.CreatedOn
+}
+
+// IsNextFullSnapCrossTimeWindow checks whether the next schedule full-snapshot can be taken within the time window or not.
```
aaronfern, line 766:
```suggestion
// IsNextFullSnapCrossTimeWindow determines if the next scheduled full-snapshot will cross the time window or not.
```

aaronfern (changes requested):

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -42,6 +43,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
```
abdasgupta, line 46: Can we use iota here?

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -42,6 +43,15 @@ import (
...
+	Min                                 = 0 // Minutes field
+	Hour                                = 1 // Hours field
+	DayOfMonth                          = 2 // Day of month field
+	Month                               = 3 // Month field
+	DayOfWeek                           = 4 // Day of week field
+	defaultFullSnapMaxTimeWindowInHours = 23.5
```
abdasgupta, line 52: If iota is used, declare this as separate constant. Can we rename this constant as DefaultFullSnapMaxTimeWindow ? Specifiy that the value is in hours inside a comment. No need to include "InHours" in variable name

ishan16696:
I have address most of the review comments. PTAL.

pkg/server/backuprestoreserver.go
ishan16696: I have address it now, can you PTAL

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -572,6 +572,97 @@ var _ = Describe("Snapshotter", func() {
 			})
 		})
 	})
+
+	Describe("Scenarios to take full-snapshot during startup", func() {
```
aaronfern, line 576: Can you please add a test for `GetFullSnapshotMaxTimeWindow()`?

aaronfern (changes requested):
Looks okay to me in general
Just that maybe function names are a bit tedious to read

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +750,53 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
...
+	}
+
+	if !ssr.WasScheduledFullSnapshotMissed(timeWindow) {
+		return false
+	}
+	return ssr.IsNextFullSnapCanCrossTimeWindow(timeWindow)
+}
+
+// WasScheduledFullSnapshotMissed determines whether the preceding full-snapshot was missed or not.
+func (ssr *Snapshotter) WasScheduledFullSnapshotMissed(timeWindow float64) bool {
```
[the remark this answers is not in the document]
ishan16696: I like this more `WasScheduledFullSnapshotMissed ` then `MissedScheduledFullSnapshot`, may be I will try to come-up with better name.

pkg/snapshot/snapshotter/snapshotter_test.go
ishan16696: added. PTAL

pkg/snapshot/snapshotter/snapshotter_test.go
ishan16696: added. PTAL

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -572,6 +572,174 @@ var _ = Describe("Snapshotter", func() {
...
+				ssr.PrevFullSnapshot = &brtypes.Snapshot{
+					CreatedOn: time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day(), (currentHour-18)%24, (currentMin)%60, 0, 0, time.Local),
+				}
+				isFullSnapCanBeMissed := ssr.IsTakingFullSnapRequiresAtStartup(fullSnapshotTimeWindow)
+				Expect(isFullSnapCanBeMissed).Should(BeTrue())
+			})
+		})
+	})
+
+	FDescribe("Scenarios to get maximum time window for full snapshot", func() {
```
aaronfern, line 699:
```suggestion
	Describe("Scenarios to get maximum time window for full snapshot", func() {
```

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -572,6 +572,174 @@ var _ = Describe("Snapshotter", func() {
 			})
 		})
 	})
+
+	FDescribe("Scenarios to take full-snapshot during startup of backup-restore", func() {
```
aaronfern, line 576:
```suggestion
	Describe("Scenarios to take full-snapshot during startup of backup-restore", func() {
```

aaronfern (changes requested):

pkg/snapshot/snapshotter/snapshotter_test.go
ishan16696: thanks for pointing this out. I forgot to remove this

ishan16696:
Please review this PR.
/ping @gardener/etcd-druid-maintainers

abdasgupta (approved):

aaronfern (approved):
/lgtm

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +750,53 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
...
+	now := time.Now()
+	nextSnapSchedule := ssr.schedule.Next(now)
+
+	if miscellaneous.GetPrevDayScheduledSnapTime(nextSnapSchedule) == ssr.PrevFullSnapshot.CreatedOn {
```
shreyas-s-rao, line 771: Looks like we're still assuming that the previous full snapshot must have been taken exactly 1 day before the next scheduled full snapshot. Is there no way to get the actual previous cron schedule? We can take inspiration from libraries like [supercronic](https://github.com/aptible/supercronic/tree/master/cronexpr) which allows getting the `Next`  cron scheduled time, and probably enhance it to get the previous cron scheduled time.

To me it looks like that will take a while to implement, so I'm still ok with the current implementation which assumes that full snapshot is taken once every day. But please add a `TODO` comment and create an issue for `respecting all cron schedules for full snapshots`, so that we can revisit the topic later and work on it if this PR needs to go in soon for the upcoming release to fix g/g alerting issues.

@unmarshall I think you were mentioning something about enhancing the existing https://github.com/robfig/cron package to provide `Next` and `Previous` cron scheduled times right?

shreyas-s-rao (commented):
@ishan16696 thanks for making the requested changes. I feel we still need to enhance the way the cron schedule is treated within etcdbr, mainly for calculating when the previous full snapshot was taken. Please check my review comments and decide whether this can be fully fixed in this PR, or will it need to be parked for an upcoming PR if there are time constraints.

ishan16696:
Resolved the merged conflict, I will adress the review comments.

pkg/miscellaneous/miscellaneous.go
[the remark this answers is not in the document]
ishan16696: I forgot to make this dynamic, thanks for pointing this out

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: I will open a issue for this.

ishan16696:
@shreyas-s-rao I have addressed the review comments. PTAL

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +750,54 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
...
+	return ssr.IsNextFullSnapshotBeyondTimeWindow(timeWindow)
+}
+
+// WasScheduledFullSnapshotMissed determines whether the preceding full-snapshot was missed or not.
+func (ssr *Snapshotter) WasScheduledFullSnapshotMissed(timeWindow float64) bool {
+	now := time.Now()
+	nextSnapSchedule := ssr.schedule.Next(now)
+
+	ssr.logger.Infof(" previous schedule: %v", miscellaneous.GetPrevScheduledSnapTime(nextSnapSchedule, timeWindow))
+	if miscellaneous.GetPrevScheduledSnapTime(nextSnapSchedule, timeWindow) == ssr.PrevFullSnapshot.CreatedOn {
```
shreyas-s-rao, line 772: Please call the function `miscellaneous.GetPrevScheduledSnapTime(nextSnapSchedule, timeWindow)` only once and assign to variable, rather than calling the same function with same input twice.

shreyas-s-rao (changes requested):
@ishan16696 thanks for making the calculation of previous scheduled full snapshot more generic. With this approach of using a `timeWindow` to determine both previous as well as next full snapshot, etcdbr now assumes that the full snapshot schedule only generates events at fixed (regular) intervals. But cron schedule still allows end-users to define schedules such as `0 9 * * 1-3`, which translate to `At 09:00 on every day-of-week from Monday through Wednesday` - this essentially means that the timeWindow between the events `9am on Tuesday`  and `9am on Wednesday` is not the same as the timeWindow between events `9am on Wednesday and 9am on the following Monday`. Hence, etcdbr still does not support all kinds of cron schedules **yet**.

Can you please add this assumption to the end-user documentation of etcd-backup-restore, so that users are aware of this current limitation and will write their cron schedule accordingly? Thanks

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: yes, it was just for testing. I forgot to remove. thanks

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -740,3 +751,57 @@ func (ssr *Snapshotter) checkSnapstoreSecretUpdate() bool {
...
+		return defaultFullSnapMaxTimeWindow
+	} else if schedule[dayOfWeek] != "*" {
+		return defaultFullSnapMaxTimeWindow * 7
+	} else if schedule[dayOfMonth] == "*" && schedule[dayOfWeek] == "*" && strings.Contains(schedule[hour], "/") {
+		if timeWindow, err := strconv.ParseFloat(schedule[hour][strings.Index(schedule[hour], "/")+1:], 64); err == nil {
+			return timeWindow
+		}
+	}
+
+	return defaultFullSnapMaxTimeWindow
```
shreyas-s-rao, line 806:
```suggestion
	if schedule[dayOfWeek] != "*" {
		return defaultFullSnapMaxTimeWindow * 7
	}

	if schedule[dayOfMonth] == "*" && strings.Contains(schedule[hour], "/") {
		if timeWindow, err := strconv.ParseFloat(schedule[hour][strings.Index(schedule[hour], "/")+1:], 64); err == nil {
			return timeWindow
		}
	}

	return defaultFullSnapMaxTimeWindow
```

ishan16696: I have addressed but in 2nd condition I still prefer to check `schedule[dayOfWeek] == "*"` as our function is not handling all permutations of cron better to have full check.
```go
if schedule[dayOfMonth] == "*" && schedule[dayOfWeek] == "*" && strings.Contains(schedule[hour], "/") {
		if timeWindow, err := strconv.ParseFloat(schedule[hour][strings.Index(schedule[hour], "/")+1:], 64); err == nil {
			return timeWindow
		}
	}
```

shreyas-s-rao (approved):
/lgtm

PR #580, opened 2023-02-01
Added support for Application credentials to authenticate openstack.

### LINKED ISSUE #575: [Feature] Implement support for Openstack Application Credentials
[Application Credentials](https://docs.openstack.org/keystone/latest/user/application_credentials.html) can be used instead of username/password to authenticate against Openstack. Backup of etcd to Swift should be possible using application credentials.
Application credentials are useful when implementing secret rotation scenarios for Openstack.
According to [documentation](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/02-openstack-swift-storage-secret.yaml) only username/password is currently allowed.

ishan16696:
This PR adds support for passing [Application credentials](https://docs.openstack.org/keystone/latest/user/application_credentials.html) to authenticate Openstack.
This PR supports the backward compatibility, it doesn't removes the old way(username/password) of passing Openstack credentials.

Fixes #575 

**Release note**:
```noteworthy operator
Added support for Application credentials to authenticate Openstack client for Openstack backup buckets.
```

### DIFF
--- a/pkg/snapstore/swift_snapstore.go
+++ b/pkg/snapstore/swift_snapstore.go
@@ -304,7 +304,7 @@ func readSwiftCredentialDir(dirName string) (*swiftCredentials, error) {
 		}
 	}
 
-	if err := isSwiftConfigEmpty(cred); err != nil {
+	if err := isSwiftConfigCorrect(cred); err != nil {
 		return nil, err
 	}
 	return cred, nil
@@ -526,8 +526,10 @@ func getSwiftHash(config *swiftCredentials) string {
 	return getHash(data)
 }
 
-func isSwiftConfigEmpty(config *swiftCredentials) error {
-	if config.AuthType == authTypePassword {
+func isSwiftConfigCorrect(config *swiftCredentials) error {
+	if (len(config.ApplicationCredentialSecret) != 0 || len(config.ApplicationCredentialID) != 0) && (len(config.Password) != 0 || len(config.Username) != 0) {
+		return fmt.Errorf("openstack swift credentials are not passed correctly")
+	} else if config.AuthType == authTypePassword {
 		if len(config.AuthURL) != 0 && len(config.TenantName) != 0 && len(config.Password) != 0 && len(config.Username) != 0 && len(config.DomainName) != 0 {
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


ishan16696:
I have tested it out for single node etcd as well as multi-node etcd cluster that etcd cluster is not having a downtime with dynamically updating the backup-storage secrets from [username/password](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/02-openstack-swift-storage-secret.yaml#L1-L13) form of credentials to [Application](https://docs.openstack.org/keystone/latest/user/application_credentials.html) form of credentials to authenticate openstack client.

pkg/snapstore/swift_snapstore.go
```diff
@@ -181,50 +233,74 @@ func readSwiftCredentialFiles(dirname string) (*clientconfig.ClientOpts, error)
...
-func readSwiftCredentialDir(dirname string) (*swiftCredentials, error) {
+func readSwiftCredentialDir(dirName string) (*swiftCredentials, error) {
 	cred := &swiftCredentials{}
-	files, err := os.ReadDir(dirname)
+	files, err := os.ReadDir(dirName)
 	if err != nil {
 		return nil, err
 	}
 
+	if cred.AuthType, err = getSwiftCredentialAuthType(files); err != nil {
```
aaronfern, line 243: Is there a case where we have Username/Password file and ApplicationCredentialID/ApplicationCredentialName/ApplicationCredentialSecret files are both present? 
If that is the case then the flow might be a bit indeterministic

aaronfern (commented):
Thanks for the well written PR @ishan16696!
Looks good, just one question

pkg/snapstore/swift_snapstore.go
ishan16696: >  Username/Password file and ApplicationCredentialID/ApplicationCredentialName/ApplicationCredentialSecret files are both present?

I'm assuming that shouldn't be the case as I have also mentioned in the [doc examples of secrets](https://github.com/gardener/etcd-backup-restore/pull/580/files#diff-ee65c41e7448f5d9a9cb70e797bf558bb21d2bf277348daf329cf8e6261084eaR39-R52).

unmarshall: @aaronfern is right. You should not assume and have a validation which prevents this from happening.

aaronfern (approved):
/lgtm

PR #582, opened 2023-02-06
Update member_control.go

mxmxchere:
It adds an error message and catches an avoidable crash.

In case "initial-advertise-peer-urls" is not set in the etcd config, the controller crashes with a panic because it tries to call the string function of a nil-pointer. With this PR you get a quick hint what needs to be changed.

**Release note**:

```improvement user
Better error message if setting in etcd config is missing
```

### DIFF
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -178,7 +178,7 @@ func getMemberPeerURL(configFile string, podName string) (string, error) {
 	}
 	initAdPeerURL := config["initial-advertise-peer-urls"]
 	if initAdPeerURL == nil {
-		return "", errors.New("initial-advertise-peer-urls must be set in etcd config.")
+		return "", errors.New("initial-advertise-peer-urls must be set in etcd config")
 	}
 	peerURL, err := miscellaneous.ParsePeerURL(initAdPeerURL.(string), podName)
 	if err != nil {
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -177,6 +177,9 @@ func getMemberPeerURL(configFile string, podName string) (string, error) {
 		return "", err
 	}
 	initAdPeerURL := config["initial-advertise-peer-urls"]
+	if initAdPeerURL == nil {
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-1:
Thank you @mxmxchere for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/member/member_control.go
```diff
@@ -177,6 +177,9 @@ func getMemberPeerURL(configFile string, podName string) (string, error) {
 		return "", err
 	}
 	initAdPeerURL := config["initial-advertise-peer-urls"]
+	if initAdPeerURL == nil {
+		return "", fmt.Errorf("initial-advertise-peer-urls must be set in etcd config.", err)
```
ishan16696, line 181:
```suggestion
		return "", fmt.Errorf("initial-advertise-peer-urls must be set in etcd config.")
```

ishan16696 (changes requested):
Thanks @mxmxchere for PR. I made one comment.

pkg/member/member_control.go
ishan16696: there is no need to add a error as `err` is coming from previous statement and it is `nil` as well

unmarshall: Instead of using `fmt.Errorf` you can just use errors.New as the error message is static.

mxmxchere: sure, thanks for the hint :+1:

ishan16696:
@mxmxchere ,
There was one flaky tests which failed, I'm unable to rerun the pipeline tests.
Kindly asking you do a force push. so we can run the pipeline checks again.
thanks

ishan16696:
Hi @mxmxchere ,
one check failed in pipeline:
```
pkg/member/member_control.go:181:25: error strings should not be capitalized or end with punctuation or a newline
```
please make required change.

ishan16696 (approved):

PR #591, opened 2023-02-21
Copy etcd snapshots in parallel

timebertt:

Copy etcd snapshots in parallel to speed up control plane migration.

In our tests, we migrated a shoot with about 110 etcd snapshots.
Sequential copying took about 3m.
Parallel copying took about 5s.

**Release note**:

```improvement operator
Etcd snapshots are copied in parallel instead of sequentially (10 in parallel by default, configurable via `--max-parallel-copy-operations`).
```


gardener-robot-ci-2:
Thank you @timebertt for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/snapshot/copier/copier.go
```diff
@@ -138,18 +140,54 @@ func (c *Copier) copyBackups() error {
...
+		}()
+	}
+
+	go func() {
+		defer close(errors)
+		wg.Wait()
+	}()
+
+	for err := range errors {
+		allErrors = multierror.Append(allErrors, err)
```
shreyas-s-rao, line 187: Wouldn't it be more beneficial to use https://pkg.go.dev/errors@master#Join that was introduced in `go 1.20`? It will require us to upgrade the repo to use `go 1.20`, but it helps us avoid adding an extra dependencies on `github.com/hashicorp/go-multierror` and `github.com/hashicorp/errwrap` for the same functionality.

pkg/snapshot/copier/copier.go
```diff
@@ -138,18 +140,54 @@ func (c *Copier) copyBackups() error {
...
+		wg        sync.WaitGroup
+		errors    = make(chan error)
+		allErrors error
+	)
+
+	for _, s := range snapshotsToCopy {
+		snapshot := s
+		wg.Add(1)
+
+		go func() {
```
shreyas-s-rao, line 168: It is generally not a good idea to spawn one go routine for each snapshot to copy. Sure, it might be faster, but for cases where there is a large number of snapshots to be copied, the `etcdbrctl copy` command might very easily run out of memory or get cpu-throttled. It's always recommended to use bounded workers for such operations, so that we don't overload the container and still achieve parallelism for copying the snapshots. I would also recommend exposing the number of workers used for copying via a CLI flag like `--num-copiers` so that this can be configured by the user, similar to `--max-fetchers` that is used to configure the maximum number of fetcher goroutines for the restoration operation in etcd-backup-restore.
 https://github.com/gardener/etcd-backup-restore/blob/ba86fb950b139cf6edd8a79c4aff65793af34b8f/pkg/types/restorer.go#L104

shreyas-s-rao (changes requested):
@timebertt thanks for raising this PR to optimise the copying of snapshots. I have a few review comments for you. Please address them, thanks.

pkg/snapshot/copier/copier.go
timebertt: I could switch this over to `errors.Join`. But I would like to refrain from upgrading the repository to go `1.20` in the scope of this PR.
Alternatively, I can replace `multierror` with a plain `[]error` type, if you prefer.

timebertt: Done in a new commit

pkg/snapshot/copier/copier.go
timebertt: Agreed, done in a new commit.
I admit, we thought about this originally but dropped the idea because we didn't think about OOM kills. That's a good argument to limit concurrency 👍

timebertt:
@shreyas-s-rao thanks for your review. I addressed the changes, PTAL :)

shreyas-s-rao (changes requested):
@timebertt thanks a lot for making the requested changes! One small nit, but other than that LGTM

shreyas-s-rao:
@timebertt I just tested the PR, and it works like a charm. I was able to perform a copy operation on a snapstore of 975 snapshots in just 3.5 minutes using the default value for `max-parallel-copy-operations` which is 10, while the time taken on master branch is ~18 minutes for the same snapstore! That's almost 6x improvement in time taken to perform the operation. Kudos!

vendor/modules.txt
```diff
@@ -393,6 +393,10 @@ github.com/grpc-ecosystem/go-grpc-prometheus
 github.com/grpc-ecosystem/grpc-gateway/internal
 github.com/grpc-ecosystem/grpc-gateway/runtime
 github.com/grpc-ecosystem/grpc-gateway/utilities
+# github.com/hashicorp/errwrap v1.0.0
+## explicit
+# github.com/hashicorp/go-multierror v1.0.0
+## explicit
```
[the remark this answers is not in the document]
timebertt: Right, I only ran `make revendor` once.
Currently, it is not idempotent because `go mod vendor` is called before `go mod tidy`. I inverted the order to make it idempotent in a new commit :)

shreyas-s-rao (approved):
/lgtm

pkg/snapshot/copier/copier.go
```diff
@@ -138,17 +142,69 @@ func (c *Copier) copyBackups() error {
...
+			for snapshot := range queue {
+				c.logger.Infof("Copying %s snapshot %s...", snapshot.Kind, snapshot.SnapName)
+				if err := c.copySnapshot(snapshot); err != nil {
+					errors <- err
+					return
+				}
+
+				c.logger.Infof("Successfully copied %s snapshot %s...", snapshot.Kind, snapshot.SnapName)
+				wg.Done()
```
shreyas-s-rao, line 193: @timebertt I just realised that `wg.Done()` will not be called when `c.copySnapshot(snapshot)` returns error, since the function returns after pushing `err` to `errors` channel. This can cause the code to hang since if even one snapshot fails to be copied, `wg.Wait()` will not allow the channels to be closed.

Can you please move this `wg.Done()` to the beginning of the goroutine, as `defer wg.Done()`?

timebertt: Nice catch! Addressed in a new commit :)

PR #600, opened 2023-03-07
Take a full snapshot when backup-restore fails to apply watch if required revision has been compacted.

### LINKED ISSUE #599: [BUG] Snapshotter fail to apply watch when revision is already compacted.
**Describe the bug:**

When `snapshotter` get close due to some error, so it tries to restart the snapshotter by apply the watch on etcd. But due to etcd's `auto-compaction` etcd might already compacted the revision number `say X` and when backup-restore tries to apply the watch on revision which is `<=X` (means revision number is already compacted) this will leads to error in watch connection hence watch channel get close, and backup-restore never able to restart the snapshotter.

**Expected behavior:**
If `snapshotter` fail to apply watch when revision is already compacted then it should take a full snapshot to come out of this situation.

1. Start the etcd and backup-restore.
2. Let the backup-restore take one full-snapshot and apply watch from revision 2.
3. Close the backup-restore
4. Put some dummy data in etcd
5. Run compaction on etcd using `etcdctl compact <Revision no>`
6. Start the backup-restore:

```cmd
INFO[0019] Applied watch on etcd from revision: 2        actor=snapshotter
WARN[0019] Failed to collect events for first delta snapshot(s): etcdserver: mvcc: required revision has been compacted  actor=backup-restore-server
INFO[0019] Starting the garbage collector...             actor=backup-restore-server
INFO[0019] Starting snapshotter...                       actor=backup-restore-server
INFO[0019] Will take next full snapshot at time: 2023-03-07 20:16:00 +0530 IST  actor=snapshotter
INFO[0019] Starting the Snapshot EventHandler.           actor=snapshotter
INFO[0019] Closing the Snapshotter...                    actor=snapshotter
ERRO[0019] Snapshotter failed with error: watch channel closed  actor=backup-restore-server
INFO[0019] Snapshotter stopped.                          actor=backup-restore-server
```

 - Etcd version/commit ID : 
 - Etcd-backup-restore version/commit ID: `v0.22.0`
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]: All

ishan16696:
This PR fixes a bug when backup-restore fails to apply watch or fails to collect events if required revision has been compacted which will leads to snapshotter not getting restarted.
Bug is fixed when backup-restore fail take a initial delta snapshot then it is better trigger a full snapshot.

Fixes #599 

**Release note**:
```bugfix operator
Fixes a bug in snapshotter loop when backup-restore fails to collect events or fails to apply watch if required etcd revision has been compacted.
```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -307,8 +307,8 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 // for the case when backup-restore becomes leading sidecar.
 func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Context, handler *HTTPHandler, ssr *snapshotter.Snapshotter, ss brtypes.SnapStore, ssrStopCh chan struct{}, ackCh chan struct{}) {
 	var (
-		err                      error
-		initialFullSnapshotTaken bool
+		err                       error
+		initialDeltaSnapshotTaken bool
 	)
 
 	for {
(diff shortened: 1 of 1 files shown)


aaronfern (approved):

ishan16696:
I have opened a issue to add a test case later for this edge case: https://github.com/gardener/etcd-backup-restore/issues/601

pkg/server/backuprestoreserver.go
```diff
@@ -391,6 +368,7 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 						b.logger.Warnf("Failed to take first delta snapshot: snapshotter failed with error: %v", err)
```
unmarshall, line 368: If you take a delta snapshot and it fails because the revision has been compacted then it will return an error. If you get an error the you continue. This means that you are repeatedly attempting to take a delta snapshot which will never succeed, right?

ishan16696: yes, you are correct this was the bug. With this PR if delta fails we will continue with full snapshot.

PR #605, opened 2023-03-18
Added a safety check before add a learner in etcd cluster.

### LINKED ISSUE #576: [Feature] Check data-dir exist or not before adding new member as a learner in etcd cluster
Before [adding new member as a learner](https://github.com/gardener/etcd-backup-restore/blob/f69cd4deac013ed0cd5a0949a488be376fe9e87c/pkg/member/member_control.go#L90), backup-restore should check whether the data-dir exists or not, if it does exist log the proper message and clean the data-dir before adding new member as learner.

If somehow old data-dir of cluster member wasn't cleaned up, then it might lead to issue of memberID mismatch error.

ishan16696:
This PR adds a safety check of whether data-dir exist or not before adding new member as a learner(non-voting member) in etcd cluster.

Fixes #576 

**Release note**:
```improvement operator
Added a safety check before adding a learner(non-voting) member in etcd cluster.
```

### DIFF
--- a/pkg/errors/error.go
+++ b/pkg/errors/error.go
@@ -34,7 +34,7 @@ func (e *SnapstoreError) Error() string {
 	return e.Message
 }
 
-// CheckErrorForNil checks whether err is nil or not.
-func CheckErrorForNil(err error) bool {
+// IsErrNotNil checks whether err is nil or not and return boolean.
+func IsErrNotNil(err error) bool {
 	return err != nil
 }
--- a/pkg/errors/error.go
+++ b/pkg/errors/error.go
@@ -34,7 +34,7 @@ func (e *SnapstoreError) Error() string {
 	return e.Message
 }
 
-// AnyError checks whether err is nil or not.
-func AnyError(err error) bool {
+// CheckErrorForNil checks whether err is nil or not.
+func CheckErrorForNil(err error) bool {
 	return err != nil
 }
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -61,7 +61,7 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
 		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
 		if isScaleup && err == nil {
-			retry.OnError(retry.DefaultBackoff, errors.CheckErrorForNil, func() error {
+			retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
 				// Additional safety check before adding a learner
 				if _, err := os.Stat(e.Config.RestoreOptions.Config.DataDir); err == nil {
 					if err := os.RemoveAll(filepath.Join(e.Config.RestoreOptions.Config.DataDir)); err != nil {
@@ -255,7 +255,7 @@ func (e *EtcdInitializer) removeDir(dirname string) error {
 // * Add a new member as a learner(non-voting member)
 func (e *EtcdInitializer) restoreInMultiNode(ctx context.Context) error {
 	m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-	if err := retry.OnError(retry.DefaultBackoff, errors.CheckErrorForNil, func() error {
+	if err := retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
 		return m.RemoveMember(ctx)
 	}); err != nil {
 		return fmt.Errorf("unable to remove the member %v", err)
@@ -265,7 +265,7 @@ func (e *EtcdInitializer) restoreInMultiNode(ctx context.Context) error {
 		return fmt.Errorf("unable to remove the data-dir %v", err)
 	}
 
-	if err := retry.OnError(retry.DefaultBackoff, errors.CheckErrorForNil, func() error {
+	if err := retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
 		// Additional safety check before adding a learner
 		if _, err := os.Stat(e.Config.RestoreOptions.Config.DataDir); err == nil {
 			if err := os.RemoveAll(filepath.Join(e.Config.RestoreOptions.Config.DataDir)); err != nil {
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -61,7 +61,7 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
 		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
 		if isScaleup && err == nil {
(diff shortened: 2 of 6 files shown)


pkg/member/member_control.go
```diff
@@ -87,7 +89,16 @@ func NewMemberControl(etcdConnConfig *brtypes.EtcdConnectionConfig) Control {
...
 
 // AddMemberAsLearner add a member as a learner to the etcd cluster
-func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
+func (m *memberControl) AddMemberAsLearner(ctx context.Context, dataDir string) error {
+	// Additional safety check before adding a learner
+	if _, err := os.Stat(dataDir); err == nil {
+		if err := os.RemoveAll(filepath.Join(dataDir)); err != nil {
+			return fmt.Errorf("failed to remove directory %s with err: %v", dataDir, err)
+		}
+	} else if err != nil && !os.IsNotExist(err) {
```
aaronfern, line 98:
```suggestion
	} else if !os.IsNotExist(err) {
```

aaronfern (changes requested):
Thanks for the PR!
Can you please add a test case to test this functionality?

ishan16696:
> Can you please add a test case to test this functionality?

A positive test case is already there, and for -ve tests I can't mock function like `os.Stat` or `os.RemoveAll` as they are not defined by interface.

aaronfern (approved):

pkg/member/member_control.go
```diff
@@ -87,7 +89,16 @@ func NewMemberControl(etcdConnConfig *brtypes.EtcdConnectionConfig) Control {
 }
 
 // AddMemberAsLearner add a member as a learner to the etcd cluster
-func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
+func (m *memberControl) AddMemberAsLearner(ctx context.Context, dataDir string) error {
```
unmarshall, line 92: `AddMemberAsLearner` is a low level interface and should not do the check for data dir. This should be done by the caller.

pkg/member/member_control_test.go
```diff
@@ -81,7 +81,7 @@ var _ = Describe("Membercontrol", func() {
 		Context("Member is not already part of the cluster", func() {
 			It("Should add member to the cluster as a learner", func() {
 				mem := member.NewMemberControl(etcdConnectionConfig)
-				err := mem.AddMemberAsLearner(context.TODO())
+				err := mem.AddMemberAsLearner(context.TODO(), os.Getenv("ETCD_DATA_DIR"))
```
unmarshall, line 84: what is the need to use an environment variable to get the etcd-data-dir?

pkg/member/member_control.go
ishan16696: make sense, I have addressed it. PTAL

pkg/errors/error.go
```diff
@@ -34,7 +34,7 @@ func (e *SnapstoreError) Error() string {
 	return e.Message
 }
 
-// AnyError checks whether err is nil or not.
-func AnyError(err error) bool {
+// CheckErrorForNil checks whether err is nil or not.
+func CheckErrorForNil(err error) bool {
```
aaronfern, line 38: Not the best name for this function as it essentially returns `false` when error passed is `nil`
Consider renaming this func

pkg/errors/error.go
```diff
@@ -34,7 +34,7 @@ func (e *SnapstoreError) Error() string {
 	return e.Message
 }
 
-// AnyError checks whether err is nil or not.
-func AnyError(err error) bool {
+// IsErrNotNil checks whether err is nil or not and return boolean.
+func IsErrNotNil(err error) bool {
```
aaronfern, line 38:
```suggestion
func ErrNotNil(err error) bool {
```
Reads better imo

Also, why are we changing the name of this func in the first place?

ishan16696: `IsErrNotNil` making more sense to me as it is checking and returning bool
> Also, why are we changing the name of this func in the first place?

`AnyError` name was not making sense to me.

aaronfern: Okay, fair enough
No hard opinion

aaronfern (approved):

pkg/errors/error.go
unmarshall: If you look at the method in isolation then your arguments are valid. However always look from the perspective of its usage which will give you the reason this function was introduced in the first place.
```go
if err := retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
		return m.RemoveMember(ctx)
	}); err != nil {
		return fmt.Errorf("unable to remove the member %v", err)
	}
```
The second parameter that `OnError` takes is `RetriableError` which is a function which takes an error and returns a boolean indicating if a retry is required. One implementation of `RetriableError` is `AnyError` which always retry irrespective of an error. The correct way to do this would to change the implementation of this method to https://github.com/gardener/dependency-watchdog/blob/master/internal/util/retry.go#L111-L113. So irrespective of the error it will always retry.

pkg/initializer/initializer.go
```diff
@@ -56,9 +56,15 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
...
 		isEtcdMemberPresent, err = m.IsMemberInCluster(ctx)
 		if !isEtcdMemberPresent && err == nil {
-			retry.OnError(retry.DefaultBackoff, func(err error) bool {
-				return err != nil
-			}, func() error {
+			retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
+				// Additional safety check before adding a learner
+				if _, err := os.Stat(e.Config.RestoreOptions.Config.RestoreDataDir); err == nil {
+					if err := os.RemoveAll(filepath.Join(e.Config.RestoreOptions.Config.RestoreDataDir)); err != nil {
+						return fmt.Errorf("failed to remove directory %s with err: %v", e.Config.RestoreOptions.Config.RestoreDataDir, err)
```
unmarshall, line 63: If you try and remove the data directory and it returns an error then you will end up trying infinitely, right? Would unlimited number of retries in this case help?

pkg/initializer/initializer.go
```diff
@@ -252,7 +258,15 @@ func (e *EtcdInitializer) restoreInMultiNode(ctx context.Context) error {
 		return fmt.Errorf("unable to remove the data-dir %v", err)
 	}
 
-	if err := retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
+	if err := retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
```
unmarshall, line 261: This seems to be a code repetition. Can we extract this into a reusable function?

pkg/initializer/initializer.go
abdasgupta: I think, it would retry for DefaultBackoff period

abdasgupta (approved):
Thanks for the PR. Looks good to me. /lgtm

pkg/initializer/initializer.go
ishan16696: yes, it will retry for DefaultBackoff period

ishan16696:
Resolved the merged conflicts. PTAL

aaronfern (approved):

PR #608, opened 2023-03-30
Enhancing the scale-up detection conditions.

### LINKED ISSUE #606: [BUG] Scale-up feature fails when Etcd API calls get fail
**Describe the bug:**
It has been observed that while migrating cluster from non-HA to HA, one of etcd cluster members fails to join the cluster.
```
NAME            READY   STATUS    RESTARTS   AGE
etcd-events-0   2/2     Running   0          16h
etcd-events-1   1/2     Running   0          11h
etcd-events-2   2/2     Running   0          11h
```

```
k logs etcd-events-1 etcd 
2023-03-28 08:12:41.399729 E | rafthttp: request cluster ID mismatch (got 64e085a973abefff want becce9846230ffd3)
2023-03-28 08:12:41.400112 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[a8316960b8709b60]=64e085a973abefff, local=becce9846230ffd3)
2023-03-28 08:12:41.479401 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[e1c53a08cad26781]=64e085a973abefff, local=becce9846230ffd3)
2023-03-28 08:12:41.479598 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[e1c53a08cad26781]=64e085a973abefff, local=becce9846230ffd3)
2023-03-28 08:12:41.480089 E | rafthttp: request cluster ID mismatch (got 64e085a973abefff want becce9846230ffd3)
2023-03-28 08:12:41.481027 E | rafthttp: request cluster ID mismatch (got 64e085a973abefff want becce9846230ffd3)
2023-03-28 08:12:41.490713 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[a8316960b8709b60]=64e085a973abefff, local=becce9846230ffd3)
2023-03-28 08:12:41.505573 E | rafthttp: request cluster ID mismatch (got 64e085a973abefff want becce9846230ffd3)
2023-03-28 08:12:41.510403 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[a8316960b8709b60]=64e085a973abefff, local=becce9846230ffd3)
2023-03-28 08:12:41.510509 E | rafthttp: request cluster ID mismatch (got 64e085a973abefff want becce9846230ffd3)
```

**Expected behavior:**
A new etcd cluster members should join the cluster as learner and scale-up should be successful to desired replicas.

**Root cause analysis**
During Initialization backup-restore was unable to [detect scale-up case](https://github.com/gardener/etcd-backup-restore/blob/b980beecb81485314c721500b3e51efefddc8af2/pkg/initializer/initializer.go#L54-L67) because [ProbeEtcd() call](https://github.com/gardener/etcd-backup-restore/blob/b980beecb81485314c721500b3e51efefddc8af2/pkg/member/member_control.go#L135) which internally calls [Etcd client GET()](https://github.com/gardener/etcd-backup-restore/blob/b980beecb81485314c721500b3e51efefddc8af2/pkg/miscellaneous/miscellaneous.go#L321) failed, hence `IsMemberInCluster()` return `false,err` which implies `not a scale-up case` Which lead this to backup-restore initialisation assume this as a [single node restoration in multi-node](https://github.com/gardener/etcd-backup-restore/blob/b980beecb81485314c721500b3e51efefddc8af2/pkg/initializer/initializer.go#L93-L98)(if snapstore is configured) or [restoration of single node cluster](https://github.com/gardener/etcd-backup-restore/blob/b980beecb81485314c721500b3e51efefddc8af2/pkg/initializer/initializer.go#L100-L109)(if snapstore is not configured).
It took action and initialisation returns successful but configuration provided to etcd was correctly calculated(of learner), hence it will result in failure of scale-up feature.

 - Etcd version/commit ID :
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]: All

ishan16696:
This PR improves the scale-up detection conditions to avoid potential failure while migrating single node etcd to multi-node etcd cluster.

Fixes #606 

**Release note**:
```improvement operator
Enhances the scale-up detection conditions to avoid potential failure while from migrating single node etcd to multi-node etcd cluster.
```

### DIFF
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -45,21 +45,29 @@ import (
 //   - Try to perform an Etcd data restoration from the latest snapshot.
 //   - No snapshots are available, start etcd as a fresh installation.
 func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int64) error {
+	logger := e.Logger.WithField("actor", "initializer")
 	metrics.CurrentClusterSize.With(prometheus.Labels{}).Set(float64(e.Validator.OriginalClusterSize))
 	start := time.Now()
 	isEtcdMemberPresent := false
 	ctx := context.Background()
 	var err error
 
+	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+	if err != nil {
+		logger.Fatalf("failed to create clientset, %v", err)
+	}
+
 	//Etcd cluster scale-up case
-	if miscellaneous.IsMultiNode(e.Logger.WithField("actor", "initializer")) {
+	if miscellaneous.IsMultiNode(logger) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isScaleup, err := m.IsClusterScaledup(ctx)
+		isScaleup, err := m.IsClusterScaledup(ctx, clientSet)
 		if isScaleup && err == nil {
-			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
+			if err := retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
-			})
-			// return here after adding member as no restoration or validation needed
+			}); err != nil {
+				logger.Fatalf("Cluster scale-up detected but unable to add new member as learner: %v", err)
+			}
+			// return here after adding non-voting member(learner) as no restoration or validation needed
 			return nil
 		}
 	}
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -69,8 +69,6 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 			}
 			// return here after adding non-voting member(learner) as no restoration or validation needed
 			return nil
-		} else if err != nil {
-			return err
 		}
 	}
 
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -54,7 +54,7 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 	//Etcd cluster scale-up case
 	if miscellaneous.IsMultiNode(e.Logger.WithField("actor", "initializer")) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isScaleup, err := member.IsScaleup(ctx, m)
+		isScaleup, err := m.IsClusterScaledup(ctx)
 		if isScaleup && err == nil {
 			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -62,11 +62,9 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
 		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
 		if isScaleup && err == nil {
-			if err := retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
+			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
-			}); err != nil {
-				logger.Fatalf("Cluster scale-up detected but unable to add new member as learner: %v", err)
-			}
+			})
 			// return here after adding non-voting member(learner) as no restoration or validation needed
 			return nil
 		}
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -54,11 +54,9 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 	//Etcd cluster scale-up case
 	if miscellaneous.IsMultiNode(e.Logger.WithField("actor", "initializer")) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isEtcdMemberPresent, err = m.IsMemberInCluster(ctx)
-		if !isEtcdMemberPresent && err == nil {
-			retry.OnError(retry.DefaultBackoff, func(err error) bool {
-				return err != nil
-			}, func() error {
+		isScaleup, err := member.IsScaleup(ctx, m)
+		if isScaleup && err == nil {
+			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
 			})
 			// return here after adding member as no restoration or validation needed
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -52,15 +52,10 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 	ctx := context.Background()
 	var err error
 
-	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
-	if err != nil {
-		logger.Fatalf("failed to create clientset, %v", err)
-	}
-
 	//Etcd cluster scale-up case
 	if miscellaneous.IsMultiNode(logger) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
+		isScaleup, err := m.IsClusterScaledUp(ctx)
 		if isScaleup && err == nil {
 			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -54,14 +54,20 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 
 	//Etcd cluster scale-up case
 	if miscellaneous.IsMultiNode(logger) {
+		clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+		if err != nil {
+			logger.Fatalf("failed to create clientset, %v", err)
+		}
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isScaleup, err := m.IsClusterScaledUp(ctx)
+		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
 		if isScaleup && err == nil {
 			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
 			})
 			// return here after adding non-voting member(learner) as no restoration or validation needed
 			return nil
+		} else if err != nil {
+			logger.Errorf("scale-up not detected: %v", err)
 		}
 	}
 
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -60,7 +60,7 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 	//Etcd cluster scale-up case
 	if miscellaneous.IsMultiNode(logger) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isScaleup, err := m.IsClusterScaledup(ctx, clientSet)
+		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
 		if isScaleup && err == nil {
 			if err := retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
 				return m.AddMemberAsLearner(ctx)
@@ -69,6 +69,8 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 			}
 			// return here after adding non-voting member(learner) as no restoration or validation needed
 			return nil
+		} else if err != nil {
+			return err
 		}
 	}
 
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -325,8 +325,8 @@ func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.
 	}
 
 	isEtcdMemberPresent, err := m.IsMemberInCluster(ctx)
-	if err != nil || isEtcdMemberPresent {
+	if err != nil {
 		return false, err
 	}
-	return true, nil
+	return !isEtcdMemberPresent, nil
 }
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -18,7 +18,7 @@ import (
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	"k8s.io/client-go/util/retry"
-	//"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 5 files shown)


pkg/miscellaneous/miscellaneous.go
```diff
@@ -564,3 +566,12 @@ func GetPrevScheduledSnapTime(nextSnapSchedule time.Time, timeWindow float64) ti
...
 }
+
+// GetBackoff returns the backoff with Factor=2
+func GetBackoff(retryPeriod time.Duration, steps int) wait.Backoff {
+	backoff := retry.DefaultBackoff
+	backoff.Duration = retryPeriod
+	backoff.Factor = 2
+	backoff.Steps = steps
+	return backoff
+}
```
seshachalam-yv, line 577:
```suggestion
// GetBackoff returns an exponential backoff with a factor of 2. The retry period and maximum number of steps for the backoff
// are specified by the `retryPeriod` and `steps` arguments, respectively
func GetBackoff(retryPeriod time.Duration, steps int) wait.Backoff {
	return wait.Backoff{
		Steps:    steps,
		Duration: retryPeriod,
		Factor:   2,
		Jitter:   retry.DefaultBackoff.Jitter,
	}
}
```

pkg/member/member_control.go
```diff
@@ -299,3 +303,30 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
...
+		logger.Errorf("unable to check scale-up case :%v", err)
+	} else if state != nil && *state == miscellaneous.ClusterStateExisting {
+		return true, nil
+	}
+
+	isEtcdMemberPresent, err := m.IsMemberInCluster(ctx)
+	if err != nil {
+		return false, err
+	}
+	return !isEtcdMemberPresent, nil
```
seshachalam-yv, line 331:
```suggestion
	isEtcdMemberPresent, err := m.IsMemberInCluster(ctx)
	if err != nil || isEtcdMemberPresent {
		return false, err
	}
	return true, nil
```

pkg/member/member_control.go
```diff
@@ -299,3 +303,30 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
...
+// IsScaleup determines whether a etcd cluster is getting scale-up or not.
+func IsScaleup(ctx context.Context, m Control) (bool, error) {
+	logger := logrus.New().WithField("actor", "scale-up")
+	// fetch pod name from env
+	podNS := os.Getenv("POD_NAMESPACE")
+	podName := os.Getenv("POD_NAME")
+
+	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+	if err != nil {
+		logger.Fatalf("failed to create clientset, %v", err)
```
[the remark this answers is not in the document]
ishan16696: yes, it is

pkg/member/member_control.go
```diff
@@ -299,3 +303,30 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 
 	return miscellaneous.CheckIfLearnerPresent(learnerCtx, cli)
 }
+
+// IsScaleup determines whether a etcd cluster is getting scale-up or not.
+func IsScaleup(ctx context.Context, m Control) (bool, error) {
```
unmarshall, line 308: Can we do the following:

1. Rename this method to `clusterScaled`
2. Make it a method in `Control` interface instead of passing an instance of `Control` to it.
3. This method does 2 completely different things which do not belong in a single method:
    - It checks if the cluster is scaled from nonHA to HA by calling `miscellaneous.GetInitialClusterStateIfScaleup`
    - It check if the current member is already in cluster.

pkg/member/member_control.go
```diff
@@ -299,3 +303,30 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
...
+	logger := logrus.New().WithField("actor", "scale-up")
+	// fetch pod name from env
+	podNS := os.Getenv("POD_NAMESPACE")
+	podName := os.Getenv("POD_NAME")
```
unmarshall, line 312: instead use `memberControl.podName`.

pkg/member/member_control.go
```diff
@@ -299,3 +303,30 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 
 	return miscellaneous.CheckIfLearnerPresent(learnerCtx, cli)
 }
+
+// IsScaleup determines whether a etcd cluster is getting scale-up or not.
+func IsScaleup(ctx context.Context, m Control) (bool, error) {
+	logger := logrus.New().WithField("actor", "scale-up")
+	// fetch pod name from env
+	podNS := os.Getenv("POD_NAMESPACE")
```
unmarshall, line 311: make `podNamespace` as an additional field in `memberControl` struct and access that.

pkg/member/member_control.go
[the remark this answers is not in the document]
unmarshall: I agree with Sesha, that we should not cause an exit from this method. This method should return an error. The caller of this method can decide to terminate instead. It also makes writing tests for this easier.

pkg/initializer/initializer.go
```diff
@@ -54,11 +54,9 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
...
 	if miscellaneous.IsMultiNode(e.Logger.WithField("actor", "initializer")) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isEtcdMemberPresent, err = m.IsMemberInCluster(ctx)
-		if !isEtcdMemberPresent && err == nil {
-			retry.OnError(retry.DefaultBackoff, func(err error) bool {
-				return err != nil
-			}, func() error {
+		isScaleup, err := member.IsScaleup(ctx, m)
+		if isScaleup && err == nil {
+			retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
```
unmarshall, line 59: if one is sure that a cluster has been scaled up from nonHA to HA then there can be 2 approaches:
1. Infinitely attempt to add a member as learner till it succeeds
2. Try for fixed number of times and then exit causing a restart of backup-restore.
One should not proceed ahead without adding this member as a learner.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -564,3 +566,12 @@ func GetPrevScheduledSnapTime(nextSnapSchedule time.Time, timeWindow float64) ti
 		nextSnapSchedule.Location(),
 	)
 }
+
+// GetBackoff returns the backoff with Factor=2
+func GetBackoff(retryPeriod time.Duration, steps int) wait.Backoff {
```
unmarshall, line 571:
```suggestion
func CreateBackoff(retryPeriod time.Duration, steps int) wait.Backoff {
```

pkg/member/member_control.go
ishan16696: > This method does 2 completely different things which do not belong in a single method:
It checks if the cluster is scaled from nonHA to HA by calling miscellaneous.GetInitialClusterStateIfScaleup
It check if the current member is already in cluster.

What makes you think that ?
IMO it `IsScaleup` does only 1 thing which it checks cluster is getting scaled-up or not using these 2 function calls:
`miscellaneous.GetInitialClusterStateIfScaleup` and `IsMemberInCluster`.

If `miscellaneous.GetInitialClusterStateIfScaleup`  says I have the annotation in sts --> surely it is scaleup case return true no need to call `IsMemberInCluster`

If `miscellaneous.GetInitialClusterStateIfScaleup` says I don't have annotation in sts  --> we still not sure about it is not being scaleup case due to bug in adding annotation to sts: https://github.com/gardener/etcd-druid/issues/557

ishan16696: >  Rename this method to clusterScaled

what about `IsClusterScaledup` ?

unmarshall: >what about IsClusterScaledup ?

Maybe we can use `IsClusterScaledUp`

pkg/member/member_control.go
```diff
@@ -38,6 +42,9 @@ type Control interface {
 	// AddMemberAsLearner add a new member as a learner to the etcd cluster
 	AddMemberAsLearner(context.Context) error
 
+	// IsClusterScaledup determines whether a etcd cluster is getting scale-up or not.
+	IsClusterScaledup(context.Context, client.Client) (bool, error)
```
unmarshall, line 46: Scaledup are 2 words and therefore should be camel-cased. So the name should be `IsClusterScaledUp`

ishan16696:
I have addressed reviews comments @unmarshall @seshachalam-yv . PTAL
I haven’t added unit tests. I will add, currently I'm running some tests on this PR.

pkg/initializer/initializer.go
```diff
@@ -45,23 +45,29 @@ import (
...
-	if miscellaneous.IsMultiNode(e.Logger.WithField("actor", "initializer")) {
+	if miscellaneous.IsMultiNode(logger) {
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isEtcdMemberPresent, err = m.IsMemberInCluster(ctx)
-		if !isEtcdMemberPresent && err == nil {
-			retry.OnError(retry.DefaultBackoff, func(err error) bool {
-				return err != nil
-			}, func() error {
+		isScaleup, err := m.IsClusterScaledup(ctx, clientSet)
+		if isScaleup && err == nil {
```
unmarshall, line 64: if a call to `IsClusterScaledup` returns an error then you currently assume that it is not a scaleup case and completely skip `AddMemberAsLeaner`.

pkg/member/member_control.go
```diff
@@ -299,3 +314,24 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
...
+// IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not and returns a boolean
+func (m *memberControl) IsClusterScaledUp(ctx context.Context) (bool, error) {
+	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
+	if err != nil {
+		m.logger.Fatalf("failed to create clientset, %v", err)
+	}
+
+	state, err := miscellaneous.GetInitialClusterStateIfScaleup(ctx, m.logger, clientSet, m.podName, m.podNamespace)
+	if err != nil {
+		m.logger.Errorf("unable to check scale-up case :%v", err)
```
aaronfern, line 327: This log is incorrect here, as we can print this message and then return `true` further down
Maybe we can update this message or remove this log altogether

aaronfern (changes requested):

pkg/initializer/initializer.go
unmarshall: >I found that if IsClusterScaledup returns an error we have to assume that it is not a scaleup case else bootstrap case 0->3 won't be possible.

There is always a possibility that an error is returned even in the case of a scale-up when calling `IsMemberInCluster` - therefore assuming that if there is an error therefore it is not a scale up case but a bootstrap (0-3) case is IMHO incorrect.

pkg/member/member_control.go
```diff
@@ -18,12 +18,16 @@ import (
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	"k8s.io/client-go/util/retry"
+	//"sigs.k8s.io/controller-runtime/pkg/client"
```
aaronfern, line 21: Remove this commented import

pkg/member/member_control.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
-	"github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
+	etcdClient "github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
```
aaronfern, line 11:
```suggestion
	etcdclient "github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
```

pkg/member/member_control.go
```diff
@@ -299,3 +314,24 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
...
+	state, err := miscellaneous.GetInitialClusterStateIfScaleup(ctx, m.logger, clientSet, m.podName, m.podNamespace)
+	if err != nil {
+		m.logger.Errorf("annotation: %v is not present: %v", miscellaneous.ScaledToMultiNodeAnnotationKey, err)
+	} else if state != nil && *state == miscellaneous.ClusterStateExisting {
+		return true, nil
+	}
+
+	isEtcdMemberPresent, err := m.IsMemberInCluster(ctx)
+	if err != nil || isEtcdMemberPresent {
+		return false, err
```
seshachalam-yv, line 334: We are returning an error, but it is not being handled [here](https://github.com/gardener/etcd-backup-restore/pull/608/files#diff-dc0aa715ea97877db344484988dcdffda1fc0eaeda89257be6c70e9d724721d7R58).
[`IsMemberInCluster`](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/member/member_control.go#L125-L172) is returning multiple errors. 
To ensure proper error handling, we should either handle the error or log it with a reason for why we are ignoring it.

pkg/initializer/initializer.go
aaronfern: Its very difficult to distinguish between the errors here
For instance, `IsMemberInCluster` internally makes a call to `ProbeEtcd`. This could fail because we're in the bootstrapping case and etcds are not running, or due to some network issue/other issue. It's very difficult to deterministically differentiate between these two right now

pkg/member/member_control.go
ishan16696: why not follow camel-case ?

unmarshall: `etcdClient` is just fine

pkg/member/member_control.go
ishan16696: I have log the error. PTAL

aaronfern (approved):

pkg/initializer/initializer.go
```diff
@@ -45,24 +45,29 @@ import (
...
+		if err != nil {
+			logger.Fatalf("failed to create clientset, %v", err)
+		}
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
-		isEtcdMemberPresent, err = m.IsMemberInCluster(ctx)
-		if !isEtcdMemberPresent && err == nil {
-			retry.OnError(retry.DefaultBackoff, func(err error) bool {
-				return err != nil
-			}, func() error {
+		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
```
seshachalam-yv, line 62:
```go
		isScaleUp, err := m.IsClusterScaledUp(ctx, clientSet)
		if err != nil {
			logger.Warnf("Failed to verify cluster scale-up: %v. Proceeding with bootstrap operation.", err)
		}
		if isScaleup && err == nil {
			err = retry.OnError(retry.DefaultBackoff, errors.AnyError, func() error {
				return m.AddMemberAsLearner(ctx)
			})
			if err != nil {
				logger.Errorf("adding learner member failed: %v", err)
				return err
			}
			// Non-voting member (learner) has been successfully added, so no further restoration or validation is needed.
			return nil
		}
```
@ishan16696 
**Nit:** please check error first

seshachalam-yv: WDYT `logger.Warnf("Failed to verify cluster scale-up: %v. Proceeding with bootstrap operation.", err)`

seshachalam-yv (commented):
Apart from the NIT item, LGTM.
/lgtm

pkg/initializer/initializer.go
ishan16696: I will address it as a separate PR as I also need to add need unit test. I hope it make sense.

PR #614, opened 2023-04-19
Fix a bug which falsely detects scale-up scenario incase of rolling update of statefulset.

ishan16696:
As now we depends on annotation added by druid to [detect scale-up feature](https://github.com/gardener/etcd-backup-restore/blob/96e810ff6a431b74580a74932cb9736fcd7e2f67/pkg/initializer/initializer.go#L62) I noticed when a rolling update happens, this piece of code with sts status:
```go
	if *curSts.Spec.Replicas > 1 && *curSts.Spec.Replicas > curSts.Status.UpdatedReplicas {
		return pointer.StringPtr(ClusterStateExisting), nil
	}
```

etcd statefulset status:
```
status:
  availableReplicas: 2
  collisionCount: 0
  currentReplicas: 2
  currentRevision: etcd-aws-78fcdddc48
  observedGeneration: 4
  readyReplicas: 2
  replicas: 3
  updateRevision: etcd-aws-5d4d79969b
  updatedReplicas: 1
```

returns the cluster state of etcd as `existing` which gives backup-restore a false impression that this is a scale-up case.

Fixes #

**Release note**:
```bugfix operator
Fixes a bug in backup-restore which falsely detects scale-up scenario incase of rolling update of statefulset.
```

### DIFF
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -17,7 +17,6 @@ import (
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
-	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -319,17 +318,13 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 // IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not and returns a boolean
 func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.Client) (bool, error) {
 	m.logger.Info("Checking whether etcd cluster is marked for scale-up")
-	curSts := &appsv1.StatefulSet{}
-	err := clientSet.Get(ctx, client.ObjectKey{
-		Namespace: m.podNamespace,
-		Name:      m.podName,
-	}, curSts)
+	etcdsts, err := miscellaneous.GetStatefulSet(ctx, clientSet, m.podNamespace, m.podName)
 	if err != nil {
-		m.logger.Errorf("unable to fetch etcd sts: %v", err)
-	}
-
-	if err == nil && curSts != nil && miscellaneous.IsAnnotationPresent(curSts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
-		return true, nil
+		m.logger.Errorf("unable to fetch etcd statefulset: %v", err)
+	} else {
+		if etcdsts != nil && miscellaneous.IsAnnotationPresent(etcdsts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
+			return true, nil
+		}
 	}
 
 	isEtcdMemberPresent, err := m.IsMemberInCluster(ctx)
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -322,7 +322,7 @@ func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.
 	if err != nil {
 		m.logger.Errorf("unable to fetch etcd statefulset: %v", err)
 	} else {
-		if etcdsts != nil && miscellaneous.IsAnnotationPresent(etcdsts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
+		if miscellaneous.IsAnnotationPresent(etcdsts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
 			return true, nil
 		}
 	}
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -17,6 +17,7 @@ import (
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
+	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -317,10 +318,17 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 
 // IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not and returns a boolean
 func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.Client) (bool, error) {
-	state, err := miscellaneous.GetInitialClusterStateIfScaleup(ctx, m.logger, clientSet, m.podName, m.podNamespace)
+	m.logger.Info("Checking whether etcd cluster is marked for scale-up")
+	curSts := &appsv1.StatefulSet{}
+	err := clientSet.Get(ctx, client.ObjectKey{
+		Namespace: m.podNamespace,
+		Name:      m.podName,
+	}, curSts)
 	if err != nil {
-		m.logger.Errorf("annotation: %v is not present: %v", miscellaneous.ScaledToMultiNodeAnnotationKey, err)
-	} else if state != nil && *state == miscellaneous.ClusterStateExisting {
+		m.logger.Errorf("unable to fetch etcd sts: %v", err)
+	}
+
+	if err == nil && curSts != nil && miscellaneous.IsAnnotationPresent(curSts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
 		return true, nil
 	}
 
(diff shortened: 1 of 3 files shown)


aaronfern:
Now we're depending solely on the annotation to detect a scale-up case, right? That will bring in a level on indeterminism as the annotation has some issues
Also, afaiu, in the case of a rolling update, a cluster must've been running earlier, and so the etcd would used the config in it's metadata which I can assume is correct

pkg/miscellaneous/miscellaneous.go
```diff
@@ -424,7 +424,7 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
 	return true
 }
 
-// GetInitialClusterStateIfScaleup checks if it is the Scale-up scenario and returns the cluster state either `new` or `existing`.
+// GetInitialClusterStateIfScaleup returns `ClusterStateExisting` if etcd statefulset contains annotation `ScaledToMultiNodeAnnotationKey'.
 func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, clientSet client.Client, podName string, podNS string) (*string, error) {
```
unmarshall, line 428: Since the function returns an initial cluster state a `nil` return value does not make much sense. You should instead return `new`. Alternatively you could rename the method to `IsClusterMarkedForScaleup` and then return (bool, error)

pkg/miscellaneous/miscellaneous.go
```diff
@@ -441,10 +441,6 @@ func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, c
 		return pointer.StringPtr(ClusterStateExisting), nil
 	}
 
-	// fallback - preserves backward compatibility
```
unmarshall, line 444: removing this condition is fine as it never made much sense anyways to determine a scale up scenario. However in case of a rolling update a new node will be launched, this node should join an existing cluster and should not create a new cluster of its own.

ishan16696:
@aaronfern  @unmarshall PTAL

pkg/member/member_control.go
```diff
@@ -317,10 +318,17 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 
 // IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not and returns a boolean
 func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.Client) (bool, error) {
-	state, err := miscellaneous.GetInitialClusterStateIfScaleup(ctx, m.logger, clientSet, m.podName, m.podNamespace)
+	m.logger.Info("Checking whether etcd cluster is marked for scale-up")
+	curSts := &appsv1.StatefulSet{}
+	err := clientSet.Get(ctx, client.ObjectKey{
+		Namespace: m.podNamespace,
+		Name:      m.podName,
+	}, curSts)
```
unmarshall, line 326: This is getting repeated in 2 places.
Have a method:
```go
// GetStatefulSet gets the StatefulSet with the name podName in podNs namespace. It will return if there is any error or the StatefulSet is not found.
func GetStatefulSet(ctx context.Context, cli client.Client, podNamespace string, podName string) (*appsv1.StatefulSet, error) {
	curSts := &appsv1.StatefulSet{}
	if err := cli.Get(ctx, client.ObjectKey{Namespace: podNamespace, Name: podName[:strings.LastIndex(podName, "-")]}, curSts); err != nil {
		return nil, err
	}
	return curSts, nil
}
```

unmarshall: In `GetInitialClusterStateIfScaleup` you can then call:
```go
	sts, err := GetStatefulSet(ctx, clientSet, podNamespace, podName)
	if err != nil {
		logger.Errorf("unable to fetch sts {namespace: %s, name: %s} %v", podNamespace, podName, err)
		return nil, err
	}

	if metav1.HasAnnotation(sts.ObjectMeta, ScaledToMultiNodeAnnotationKey) {
		return pointer.StringPtr(ClusterStateExisting), nil
	}
```

unmarshall: And in member_control.go/IsClusterScaleup you can have:
```go
	sts, err := miscellaneous.GetStatefulSet(ctx, clientSet, m.podNamespace, m.podName)
	if err != nil {
		m.logger.Errorf("unable to fetch sts {namespace: %s, name: %s} %v", m.podNamespace, m.podName, err)
	} else {
		if metav1.HasAnnotation(sts.ObjectMeta, miscellaneous.ScaledToMultiNodeAnnotationKey) {
			return true, nil
		}
	}
	isEtcdMemberPresent, err := m.IsMemberInCluster(ctx)
	if err != nil || isEtcdMemberPresent {
		return false, err
	}
	return true, nil

```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -426,25 +426,20 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
...
+	// Read etcd statefulset to check annotation or updated replicas to toggle `initial-cluster-state`
+	etcdSts, err := GetStatefulSet(ctx, clientSet, podNS, podName)
 	if err != nil {
-		logger.Warn("error fetching etcd sts ", err)
+		logger.Errorf("unable to fetch statefulset {namespace: %s, name: %s} %v", podNS, podName[:strings.LastIndex(podName, "-")], err)
 		return nil, err
 	}
 
-	if IsAnnotationPresent(curSts, ScaledToMultiNodeAnnotationKey) {
+	if IsAnnotationPresent(etcdSts, ScaledToMultiNodeAnnotationKey) {
```
unmarshall, line 436: you do not really require a function for a single statement. You can keep it but it might be an overkill :)

pkg/member/member_control.go
```diff
@@ -319,17 +318,13 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
...
+	etcdsts, err := miscellaneous.GetStatefulSet(ctx, clientSet, m.podNamespace, m.podName)
 	if err != nil {
-		m.logger.Errorf("unable to fetch etcd sts: %v", err)
-	}
-
-	if err == nil && curSts != nil && miscellaneous.IsAnnotationPresent(curSts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
-		return true, nil
+		m.logger.Errorf("unable to fetch etcd statefulset: %v", err)
+	} else {
+		if etcdsts != nil && miscellaneous.IsAnnotationPresent(etcdsts, miscellaneous.ScaledToMultiNodeAnnotationKey) {
```
unmarshall, line 325: can `etcdsts` ever be nil?

ishan16696: It is just a safety check.

unmarshall: ok. It is not really required. If it is nil then we should let it fail as that would be quite weird and totally unexpected. If we do this in one place then we have to make sure that we do it everywhere else as well. This is called `defensive programming`

ishan16696: removed.

unmarshall (approved):
/lgtm

ishan16696:
@aaronfern do you want to take look ?

pkg/member/member_control.go
```diff
@@ -317,11 +317,14 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 
 // IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not and returns a boolean
 func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.Client) (bool, error) {
-	state, err := miscellaneous.GetInitialClusterStateIfScaleup(ctx, m.logger, clientSet, m.podName, m.podNamespace)
+	m.logger.Info("Checking whether etcd cluster is marked for scale-up")
+	etcdsts, err := miscellaneous.GetStatefulSet(ctx, clientSet, m.podNamespace, m.podName)
 	if err != nil {
-		m.logger.Errorf("annotation: %v is not present: %v", miscellaneous.ScaledToMultiNodeAnnotationKey, err)
```
aaronfern, line 322: Can you re-add this log? seems useful to me

aaronfern (commented):
One nit, otherwise /lgtm from me

pkg/member/member_control.go
ishan16696: previously we are calling `GetInitialClusterStateIfScaleup()` but now we are getting statefulset separately and `IsAnnotationPresent()` only returns boolean.

ishan16696: If annotation is not present it can be infer from these 2 logs:
`Checking whether etcd cluster is marked for scale-up`
`Checking if member %s is part of a running cluster`

aaronfern: Okay, now that I see the logs together, it seems fine enough

aaronfern (approved):

PR #616, opened 2023-04-24
Add unit tests for func IsClusterScaledUp.

ishan16696:
This PR adds unit tests for func `IsClusterScaledUp`:
https://github.com/gardener/etcd-backup-restore/blob/281a6ca98d91a22d76fae9b4ebbf429a97cf7784/pkg/member/member_control.go#L319

Fixes #

**Release note**:
```improvement operator
None
```

### DIFF
--- a/pkg/compactor/compactor_suite_test.go
+++ b/pkg/compactor/compactor_suite_test.go
@@ -68,7 +68,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	logger.Infof("Snapshot Directory is: %s", testSnapshotDir)
 
 	// Start the main ETCD process that will run untill all compaction test cases are run
-	etcd, err = utils.StartEmbeddedEtcd(testCtx, testEtcdDir, logger, embeddedEtcdPortNo)
+	etcd, err = utils.StartEmbeddedEtcd(testCtx, testEtcdDir, logger, utils.DefaultEtcdName, embeddedEtcdPortNo)
 	Expect(err).ShouldNot(HaveOccurred())
 	endpoints = []string{etcd.Clients[0].Addr().String()}
 	logger.Infof("endpoints: %s", endpoints)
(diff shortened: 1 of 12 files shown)


aaronfern (commented):
Thanks for the tests @ishan16696!

Is it possible to add two more tests?
1. When the sts is not present and the `GetStatefulSet()` call returns an error
2. When the scale-up annotation is present and the member is already part of the list

ishan16696:
> When the sts is not present and the GetStatefulSet() call returns an error

It is already there: 
https://github.com/gardener/etcd-backup-restore/blob/281a6ca98d91a22d76fae9b4ebbf429a97cf7784/pkg/miscellaneous/miscellaneous_test.go#L656


> When the scale-up annotation is present and the member is already part of the list

If annotation is present then `IsClusterScaledUp ` simply return true won't check member list.

aaronfern:
> It is already there:

Yes, but that is for `GetStatefulSet()`. Having this test for `IsClusterScaleUp()` ensures that we're catching this error and propagating it correctly to the caller

> If annotation is present then IsClusterScaledUp  simply return true won't check member list.

My Apologies. I meant when the scale-up annotation is not present and the member is already part of the list

ishan16696:
> My Apologies. I meant when the scale-up annotation is not present and the member is already part of the list

ok, member pkg tests is starting the embedded etcd, so to mimic this I have to start embedded etcd with podname.

> Yes, but that is for GetStatefulSet(). Having this test for IsClusterScaleUp() ensures that we're catching this error and propagating it correctly to the caller

but statefulset is not present then `IsClusterScaleUp` just logging the error and move to api call, I can add a test but we won't be able to detect.

aaronfern:
> so to mimic this I have to start embedded etcd with podname.

Either that or ensure that the `memberControl` object has a different pod name

> I can add a test but we won't be able to detect

Okay, yup, I get your point now. We can scrap this test

aaronfern (approved):
Thanks for making the changes @ishan16696!
/lgtm

PR #617, opened 2023-04-26
If scale-up is detected then add learner with backoff.

ishan16696:
If scale-up case is detected then better to add a learner with exponential backoff.

Fixes # https://github.com/gardener/etcd-backup-restore/pull/608#discussion_r1155803217

**Release note**:
```improvement operator
Add a learner with backoff in case of scale-up feature is triggered.
```

### DIFF
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -36,8 +36,8 @@ import (
 )
 
 const (
-	// retrySteps is the no. of steps used for exponential backoff to add a learner.
-	retrySteps = 6
+	// addLearnerAttempts are the total number of attempts that will be made to add a learner
+	addLearnerAttempts = 6
 )
 
 // Initialize has the following steps:
@@ -67,12 +67,12 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
 		if err != nil {
 			logger.Errorf("scale-up not detected: %v", err)
-		} else if isScaleup && err == nil {
+		} else if isScaleup {
 			logger.Info("Etcd cluster scale-up is detected")
 			// Add a learner(non-voting member) to a etcd cluster with retry
 			// If backup-restore is unable to add a learner in a cluster
 			// restart the `initialization` by exiting the backup-restore.
-			if err := m.AddLearnerWithRetry(ctx, retrySteps, e.Config.RestoreOptions.Config.DataDir); err != nil {
+			if err := member.AddLearnerWithRetry(ctx, m, addLearnerAttempts, e.Config.RestoreOptions.Config.DataDir); err != nil {
 				logger.Fatalf("unable to add a learner in a cluster: %v", err)
 			}
 			// return here after adding learner(non-voting member) as no restoration or validation required.
--- a/pkg/initializer/initializer.go
+++ b/pkg/initializer/initializer.go
@@ -35,6 +35,11 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+const (
+	// retrySteps is the no. of steps used for exponential backoff to add a learner.
+	retrySteps = 6
+)
+
 // Initialize has the following steps:
 //   - Check if data directory exists.
 //   - If data directory exists
@@ -60,22 +65,18 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
 		}
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
 		isScaleup, err := m.IsClusterScaledUp(ctx, clientSet)
-		if isScaleup && err == nil {
-			retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
-				// Additional safety check before adding a learner
-				if _, err := os.Stat(e.Config.RestoreOptions.Config.DataDir); err == nil {
-					if err := os.RemoveAll(filepath.Join(e.Config.RestoreOptions.Config.DataDir)); err != nil {
-						return fmt.Errorf("failed to remove directory %s with err: %v", e.Config.RestoreOptions.Config.DataDir, err)
-					}
-				} else if !os.IsNotExist(err) {
-					return err
-				}
-				return m.AddMemberAsLearner(ctx)
-			})
-			// return here after adding non-voting member(learner) as no restoration or validation needed
-			return nil
-		} else if err != nil {
+		if err != nil {
 			logger.Errorf("scale-up not detected: %v", err)
+		} else if isScaleup && err == nil {
+			logger.Info("Etcd cluster scale-up is detected")
+			// Add a learner(non-voting member) to a etcd cluster with retry
+			// If backup-restore is unable to add a learner in a cluster
+			// restart the `initialization` by exiting the backup-restore.
+			if err := m.AddLearnerWithRetry(ctx, retrySteps, e.Config.RestoreOptions.Config.DataDir); err != nil {
+				logger.Fatalf("unable to add a learner in a cluster: %v", err)
+			}
+			// return here after adding learner(non-voting member) as no restoration or validation required.
+			return nil
 		}
 	}
 
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
-	"path/filepath"
 	"strconv"
 	"time"
 
@@ -43,9 +41,6 @@ var (
 type Control interface {
 	// AddMemberAsLearner add a new member as a learner to the etcd cluster
 	AddMemberAsLearner(context.Context) error
-
-	// AddLearnerWithRetry add a new member as a learner with exponential backoff.
-	AddLearnerWithRetry(context.Context, int, string) error
 
 	// IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not.
 	IsClusterScaledUp(context.Context, client.Client) (bool, error)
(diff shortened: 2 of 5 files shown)


pkg/initializer/initializer.go
```diff
@@ -60,22 +65,18 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
...
-					}
-				} else if !os.IsNotExist(err) {
-					return err
-				}
-				return m.AddMemberAsLearner(ctx)
-			})
-			// return here after adding non-voting member(learner) as no restoration or validation needed
-			return nil
-		} else if err != nil {
+		if err != nil {
```
unmarshall, line 68: if err is nil then it cannot be guaranteed that it is not a scale-up case. I think this is the limitation with which we have to live with at the moment, right?

pkg/initializer/initializer.go
```diff
@@ -60,22 +65,18 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode, failBelowRevision int6
...
 			logger.Errorf("scale-up not detected: %v", err)
+		} else if isScaleup && err == nil {
```
unmarshall, line 70: in the else clause there is no need to check `err == nil` as it will always be true.

pkg/initializer/initializer.go
```diff
@@ -35,6 +35,11 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+const (
+	// retrySteps is the no. of steps used for exponential backoff to add a learner.
+	retrySteps = 6
```
unmarshall, line 40: this is too generic and today it is only used to add a learner. Can we rename this to `addLearnerAttempts`. The description should just say `addLearnerAttempts are the total number of attempts that will be made to add a learner` - no need to mention exponential backoff as the number of attempts has nothing to do with how you wait between 2 attempts.

seshachalam-yv, line 40: Out of curiosity, I am asking how we arrived at the number `6`?

pkg/member/member_control.go
```diff
@@ -42,6 +44,9 @@ type Control interface {
 	// AddMemberAsLearner add a new member as a learner to the etcd cluster
 	AddMemberAsLearner(context.Context) error
 
+	// AddLearnerWithRetry add a new member as a learner with exponential backoff.
+	AddLearnerWithRetry(context.Context, int, string) error
```
unmarshall, line 48: Control interface is a low level interface and should not have this method. It is the responsibility of the caller to wrap it with retry logic. Create a util function - you can also take inspiration from https://github.com/gardener/dependency-watchdog/blob/master/internal/util/retry.go

pkg/member/member_control.go
```diff
@@ -136,6 +144,34 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
 	return nil
 }
 
+// AddLearnerWithRetry add a new member as a learner with exponential backoff.
+func (m *memberControl) AddLearnerWithRetry(ctx context.Context, retrySteps int, dataDir string) error {
+	backoff := miscellaneous.CreateBackoff(RetryPeriod, retrySteps)
+
+	if err := retry.OnError(backoff, func(err error) bool {
+		return err != nil
```
unmarshall, line 152: Can you create a util function
```go
func AnyError(err error) bool {
   return err != nil
}
```

pkg/member/member_control.go
```diff
@@ -136,6 +144,34 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
...
+	}, func() error {
+		// Additional safety check before adding a learner
+		if _, err := os.Stat(dataDir); err == nil {
+			if err := os.RemoveAll(filepath.Join(dataDir)); err != nil {
+				return fmt.Errorf("failed to remove directory %s with err: %v", dataDir, err)
+			}
+		} else if !os.IsNotExist(err) {
+			return err
```
unmarshall, line 160: If the data directory does not exist then an error is returned. retry.OnError will retry 6 times and every time this error will be returned. I guess you want an early exit if the data directory is not present. In that case please create an error and then filter that error in `retriable` function argument to `retry.OnError`

pkg/initializer/initializer.go
ishan16696: > if err is nil then it cannot be guaranteed that it is not a scale-up case

yes, as you can check when member is already a part of cluster then it will return false,nil
https://github.com/gardener/etcd-backup-restore/blob/1ecfb1a55d9d1969dff6eb245e9e3a2de390d0bc/pkg/member/member_control.go#L330

pkg/initializer/initializer.go
ishan16696: we are using `Backoff` with factor `2` with Duration set to `2 sec` , so I guess no. of retry `6` should be sufficient. WDYT?
https://github.com/gardener/etcd-backup-restore/blob/1ecfb1a55d9d1969dff6eb245e9e3a2de390d0bc/pkg/miscellaneous/miscellaneous.go#L584

pkg/initializer/initializer.go
ishan16696: yes, agreed

pkg/member/member_control.go
ishan16696: > I guess you want an early exit if the data directory is not present

I don't get it, what makes you think so ?
```go
 func() error {
		if (some condition) {
                         // do something
			return err --> here it will retry
		} 
		return nil ---> here it will exit
	}
```
If err is other than `data dir not exist`, then we will retry 
else 
better to try to add a learner

pkg/member/member_control.go
ishan16696: why to create a retry func when we can reuse this `"k8s.io/client-go/util/retry"` ?

ishan16696: I can move `AddLearnerWithRetry ` out of this interface to different pkg

pkg/initializer/initializer.go
unmarshall: Yes but it is also possible that there is an error trying to make the call `m.IsMemberInCluster`. So in case of an error one ideally cannot assume that it is not part of the cluster. It only means that its membership is `unknown` as we could not determine it.

pkg/member/member_control.go
unmarshall: >why to create a retry func when we can reuse this "k8s.io/client-go/util/retry"

Sure you can use util.retry as you are using today. I only gave that as an example to have a utility function to do retries.

pkg/member/member_control.go
```diff
@@ -136,6 +144,34 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
...
 
+// AddLearnerWithRetry add a new member as a learner with exponential backoff.
+func (m *memberControl) AddLearnerWithRetry(ctx context.Context, retrySteps int, dataDir string) error {
+	backoff := miscellaneous.CreateBackoff(RetryPeriod, retrySteps)
+
+	if err := retry.OnError(backoff, func(err error) bool {
+		return err != nil
+	}, func() error {
+		// Additional safety check before adding a learner
+		if _, err := os.Stat(dataDir); err == nil {
```
unmarshall, line 155: This code has an issue.
You can try playing around with using errors in if and else [here](https://go.dev/play/p/yNP7RmoCw4G).

pkg/initializer/initializer.go
ishan16696: yes

pkg/member/member_control.go
ishan16696: Hey @unmarshall ,
this is want I want to do: 
```go
func main() {
	if _, err := os.Stat("/tmp/tringo/bingo"); err == nil {
		fmt.Println("file is found")
	} else if !os.IsNotExist(err) {
		fmt.Printf("error other than not exist: %v", err)
	} else {
		fmt.Printf("file not found %v", err)
	}
}
```
to return `err` only when `err !=data dir not exist` 
if `err==data dir not exist` then I want to continue to add a learner

unmarshall: ok, then just move out this logic into a separate function and write tests for it

pkg/member/member_control.go
```diff
@@ -136,6 +144,34 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
 	return nil
 }
 
+// AddLearnerWithRetry add a new member as a learner with exponential backoff.
+func (m *memberControl) AddLearnerWithRetry(ctx context.Context, retrySteps int, dataDir string) error {
+	backoff := miscellaneous.CreateBackoff(RetryPeriod, retrySteps)
```
aaronfern, line 149: Can you add a cap to the backoff so that we don't exceed a reasonable limit? seems to me like it can exceed to an unreasonably high value

pkg/member/member_control.go
```diff
@@ -136,6 +144,34 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
...
+
+	if err := retry.OnError(backoff, func(err error) bool {
+		return err != nil
+	}, func() error {
+		// Additional safety check before adding a learner
+		if _, err := os.Stat(dataDir); err == nil {
+			if err := os.RemoveAll(filepath.Join(dataDir)); err != nil {
+				return fmt.Errorf("failed to remove directory %s with err: %v", dataDir, err)
+			}
+		} else if !os.IsNotExist(err) {
```
aaronfern, line 159: I might've missed something, but we aren't handling the case here where `os.Stat` returns an error that is not `os.IsNotExist`. We're just continuing onto `AddMemberAsLearner`

ishan16696: please check this https://github.com/gardener/etcd-backup-restore/pull/617#discussion_r1178837127

pkg/member/member_control.go
ishan16696: > Can you add a cap to the backoff so that we don't exceed a reasonable limit? seems to me like it can exceed to an unreasonably high value

but don't we want to add a learner if control comes here then it means we are sure it is the scale-up case, so I guess better to retry with backoff. WDYT?  that's why I passed `retrySteps=6` . Do you have any other ideas/suggestions ?

pkg/member/member_control.go
ishan16696: added the unit tests. PTAL

pkg/member/member_control.go
aaronfern: Okay, thank you

pkg/member/member_control.go
ishan16696: ok, I have set the upper limit of 120sec ?

unmarshall (approved):
/lgtm

pkg/miscellaneous/miscellaneous.go
```diff
@@ -580,12 +580,25 @@ func GetPrevScheduledSnapTime(nextSnapSchedule time.Time, timeWindow float64) ti
...
 	}
 }
+
+// RemoveDir removes the directory(if exist) and do nothing if directory doesn't exist.
+func RemoveDir(dir string) error {
+	if _, err := os.Stat(dir); err == nil {
+		if err := os.RemoveAll(filepath.Join(dir)); err != nil {
+			return fmt.Errorf("failed to remove directory %s with err: %v", dir, err)
+		}
+	} else if !os.IsNotExist(err) {
```
aaronfern, line 600: Preferred way to check for `NotExists` error is `errors.Is(err, fs.ErrNotExist)`
ref: https://pkg.go.dev/os#IsNotExist

aaronfern (commented):
Just one final nit from me. Else looks good to be merged
/lgtm

pkg/miscellaneous/miscellaneous.go
ishan16696: thanks for pointing this out but provided link says `It only supports errors returned by the os package`
and we will also get error from `os pkg`. so I guess it will be fine.
I also checked inside `os pkg`:
```
func IsNotExist(err error) bool {
	return underlyingErrorIs(err, ErrNotExist)
}
```
```
ErrNotExist   = fs.ErrNotExist   // "file does not exist"
```
so, internally it is the same thing.

unmarshall: Using `os.IsNotExist` is fine.

aaronfern (approved):

PR #624, opened 2023-05-05
Improve scale-up detection.

### LINKED ISSUE #622: [Enhancement] Improve scale-up detection to provide correct config to etcd
While working on this issue: https://github.com/gardener/etcd-druid/issues/582 @unmarshall and I found sometime sts annotation might not give us always true information about scale-up. Please see this: https://github.com/gardener/etcd-druid/pull/587#discussion_r1182771744 and https://github.com/gardener/etcd-druid/pull/587#discussion_r1183231059

### LINKED ISSUE #623: [BUG] Race condition in reading scale-up annnotation
**Describe the bug:**
When cluster is marked for scale-up, etcd-druid adds an annotation `ScaledToMultiNodeAnnotationKey`  in etcd statefulset to indicate it is scale-up scenario and after scale-up becomes successful, etcd-druid might take sometime in removing this annotation but at the sametime `etcd-0` 0th pod will get restart and it has been observed that sometime it reads this scale-up `ScaledToMultiNodeAnnotationKey` annotation before etcd-druid removes it.
There is race-condition happening between scale-up annotation being removed by druid and pod restarted just after scale-up.

**Expected behavior:**
It shouldn’t detected the false positive for scale-up scenario.

**Logs:**

Although it is not a big issue as it will only leads to [single-member restoration](https://github.com/gardener/etcd-druid/blob/master/docs/single_member_restoration.md) without any downtime and that's why druid e2e tests are also passing without any issue.

ishan16696:
First, try to determine scale-up case by checking whether member is already part of cluster or not. In case of failure of api call then check for `ScaledToMultiNodeAnnotationKey` annotation in etcd statefulset. This will remove any case of false positive for scale-up scenario.

Fixes #623 and Fixes #622

**Release note**:
```improvement operator
None
```

### DIFF
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -322,6 +322,17 @@ func (m *memberControl) IsLearnerPresent(ctx context.Context) (bool, error) {
 // IsClusterScaledUp determines whether a etcd cluster is getting scale-up or not and returns a boolean
 func (m *memberControl) IsClusterScaledUp(ctx context.Context, clientSet client.Client) (bool, error) {
 	m.logger.Info("Checking whether etcd cluster is marked for scale-up")
+
+	// First, try to determine scale-up case by checking whether member is already part of cluster or not.
+	// In case of failure in checking the presence of member in a cluster then
+	// check for `ScaledToMultiNodeAnnotationKey` annotation in etcd statefulset.
+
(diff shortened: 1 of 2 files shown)


ishan16696:
druid e2e tests passed ✅

aaronfern (approved):
/lgtm

PR #628, opened 2023-05-24
wait for etcd to be ready before continuing with memberControl object…

aaronfern:
… creation

We notices that etcd backup restore does not wait for etcd to be ready before creating the `MemberControl` object as well as the `LeaderElector` object
This PR adds changes for `etcd-backup-restore` to wait for etcd to get ready before it can continue

Fixes #

**Release note**:

```improvement operator

```


ishan16696:
> We notices that etcd backup restore does not wait for etcd to be ready before creating the MemberControl object as well as the LeaderElector object
This PR adds changes for etcd-backup-restore to wait for etcd to get ready before it can continue

I kindof disagree with the PR description ... please mention about the root cause this PR is solving which is `when etcd is not up which make update PeerUrl api call fails.`

pkg/server/backuprestoreserver.go
```diff
@@ -153,6 +153,50 @@ func (b *BackupRestoreServer) startHTTPServer(initializer initializer.Initialize
...
+			return ctx.Err()
+		case <-ticker.C:
+		}
+	}
+	logger.Info("Etcd is now running. Continuing br startup")
+	return nil
+}
+
+func isEtcdRunning(ctx context.Context, timeout time.Duration, etcdConnectionConfig *brtypes.EtcdConnectionConfig, logger *logrus.Logger) bool {
+	var endPoint string
```
seshachalam-yv, line 172:
```suggestion
```
Remove var endpoint if we are agreeing with below suggest comment

pkg/server/backuprestoreserver.go
```diff
@@ -153,6 +153,50 @@ func (b *BackupRestoreServer) startHTTPServer(initializer initializer.Initialize
...
+		logger.Errorf("etcd endpoints are not passed correctly")
+		return false
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	_, err = client.Status(ctx, endPoint)
+	if err != nil {
+		logger.Errorf("failed to get status of etcd endPoint: %v with error: %v", endPoint, err)
```
seshachalam-yv, line 194:
```suggestion
	if len(etcdConnectionConfig.Endpoints) == 0 {
		logger.Errorf("etcd endpoints are not passed correctly")
		return false

	}
	endpoint := etcdConnectionConfig.Endpoints[0]

	ctx, cancel := context.WithTimeout(ctx, timeout)
	defer cancel()

	_, err = client.Status(ctx, endpoint)
	if err != nil {
		logger.Errorf("failed to get status of etcd endPoint: %v with error: %v", endpoint, err)
```

Can we follow consistent of checking error and return false if any error

WDYT?

ishan16696:
Updated the PR description and release note, @aaronfern please check. hope that’s fine.

aaronfern:
Thanks @ishan16696!
Description is very apt! I've also changed the release note slightly. Hope it still makes sense

unmarshall (approved):
/lgtm

PR #641, opened 2023-07-01
Introduce persona-based docs

vlerenc:
We would like to introduce persona-based docs focussing on the already known release notes personas:
- users
- operators
- developers

@n-boshnakov Can you please share your thoughts?
@shreyas-s-rao We agreed not to change too much (maybe anything), but I thought, maybe we can prepare some basics already, so that you have a better starting point later. Anyway, please check whether the small change is OK for you. Thanks in advance.

**Release note**:
None as ~30 PRs have/will be opened and the message would be the same everywhere.


n-boshnakov:
Just a note, but the newly created manifest will need to also be added to one of the gardener/documentation manifests (probably in [other-components.yaml](https://github.com/gardener/documentation/blob/master/.docforge/documentation/other-components.yaml)) before it can appear on the website.

vlerenc:
Thanks @n-boshnakov, I added this component there.

Are the etcd backup&restore maintainers otherwise OK to review the PR?

n-boshnakov:
> Thanks @n-boshnakov, I added this component there.
> 
> Are the etcd backup&restore maintainers otherwise OK to review the PR?

Yes, they can review the PR.

vlerenc:
Is it OK to merge this PR (cc @shreyas-s-rao)?

.docforge/manifest.yaml
```diff
@@ -0,0 +1,36 @@
+nodesSelector:
+  path: /docs
+structure:
+- name: _index.md
+  properties:
+    frontmatter:
+      title: etcd Backup & Restore
+      description: A sidecar for etcd backup & restore in Gardener
```
ishan16696, line 8: > A sidecar for etcd backup & restore in Gardener

Just I'm curious, Do we have to mention Gardener here as IMO etcd-backup-restore can be consumed independently as well like linode uses etcd-backup-restore.

I guess we should mention about etcd-druid
```
      description: A sidecar for etcd backup & restore deployed by etcd-operator aka etcd-druid
```
But I have no strong opinion.

.docforge/manifest.yaml
```diff
@@ -0,0 +1,36 @@
+nodesSelector:
+  path: /docs
+structure:
+- name: _index.md
+  properties:
+    frontmatter:
+      title: etcd Backup & Restore
```
shreyas-s-rao, line 7:
```suggestion
      title: Etcd Backup & Restore
```

.docforge/manifest.yaml
vlerenc: OK, changed (made Gardener only an example) @ishan16696.

.docforge/manifest.yaml
vlerenc: OK, changed and squashed.

shreyas-s-rao (approved):
/lgtm

PR #647, opened 2023-07-17
Upgrade Ginkgo v1 to v2 and Update Other Dependencies

seshachalam-yv:

This pull request aims to upgrade the Ginkgo from v1 to v2 and update other dependencies

In the process of these upgrades, several dependency conflicts were encountered with the `helm.sh/helm/v3` library. The updated dependencies included an upgrade to `helm.sh/helm/v3` to v3.5.0 which was found to be incompatible with `k8s.io/client-go` v0.20.6. An attempt to resolve this issue by upgrading `helm.sh/helm/v3` to v3.12.0 did not resolve the issue due to a hard dependency on `google.golang.org/grpc` v1.29.0 by `go.etcd.io/etcd` v0.5.0-alpha.5.0.20200824191128-ae9734ed278b, while `helm.sh/helm/v3` v3.12.0 required `grpc` v1.52.

As a workaround for these dependency issues, the `k8s.io/helm` library (Helm V2) was used in place of `helm.sh/helm/v3`. This change should not impact the functionality of the integration tests as the Helm library is used only for deploying charts in these tests.

It's important to note that `k8s.io/helm` represents Helm V2, which is deprecated. Hence, this change is a temporary measure to allow for the necessary dependency updates. A more permanent solution that allows for the use of a supported Helm version should be considered for future updates.

**Release note**:

```other operator
Upgraded Ginkgo v1 to v2 and updated other dependencies
```


.ci/unit_test
```diff
@@ -52,13 +52,13 @@ TEST_PACKAGES="${TEST_PACKAGES:-"cmd pkg"}"
 
 RUN_NEGATIVE="${RUN_NEGATIVE:-"true"}"
 
-GINKGO_COMMON_FLAGS=("-r" "-timeout=1h0m0s" "--progress" "-slowSpecThreshold=180" "-mod=vendor")
```
aaronfern: Why remove these two flags?
I understand why you'd like to remove `slowSpecThreshold`, but won't `progress` be beneficial since we have so many tests here?

seshachalam-yv: But I have added `--show-node-events` instead of `progress` and `--fail-on-pending` to prevent skipping the test cases

.ci/unit_test
```diff
@@ -16,64 +16,54 @@ set -e
...
+  # Run all non-negative scenarios with randomize-all parameters.
+  ginkgo -race -trace $GINKGO_COMMON_FLAGS -gcflags=all=-d=checkptr=0 --randomize-all --randomize-suites --skip="NEGATIVE\:.*" $TEST_PACKAGES
 
-  if [ "$RUN_NEGATIVE" = "true" ] ; then
+  if [[ "${RUN_NEGATIVE}" == "true" ]]; then
     echo "[INFO] Running negative tests now..."
-    #run negative scenario in a sequenced manner (removed failOnPending as one spec in restore test is marked as 'X' for excluding)
-    ginkgo -race -trace $GINKGO_COMMON_FLAGS -gcflags=all=-d=checkptr=0 --focus="NEGATIVE\:.*" ${TEST_PACKAGES}
+    # Run negative scenarios sequentially (removed failOnPending as one spec in restore test is marked as 'X' for excluding)
+    ginkgo -race -trace $GINKGO_COMMON_FLAGS  $TEST_PACKAGES
```
aaronfern, line 66: I think you need a `--focus` here else it will just run all tests again

aaronfern (changes requested):

.ci/unit_test
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-backup-restore/pull/647/commits/68b6f56000e2f3b43df2bdb9469ca1032766a287

aaronfern (approved):
Thanks for the changes!
/lgtm

PR #656, opened 2023-08-08
move ioutil func calls to different pkg as ioutil has been deprecated.

ishan16696:
This PR moves `ioutil` calls to `os pkg` as ioutil has been deprecated in Go 1.16.
Please see this: https://pkg.go.dev/io/ioutil

Fixes #

**Release note**:
```improvement operator
None
```


aaronfern (approved):
Thanks for the PR @ishan16696!
/lgtm

PR #659, opened 2023-08-18
Revendor the bbolt from `v1.3.6` -> `v1.3.7`

ishan16696:
I found out that etcd version `3.4.26` is using bbolt `v1.3.7`: https://github.com/etcd-io/etcd/blob/v3.4.26/go.mod#L35
and in backup-restore we are using bbolt `v1.3.6`: https://github.com/gardener/etcd-backup-restore/blob/master/go.mod#L33
I guess we need to revendor the bbolt in backup-restore as in this PR: https://github.com/gardener/etcd-druid/pull/656 , we are upgrading our etcd to v3.4.26

Fixes #

cc @shreyas-s-rao 

**Release note**:
```improvement operator
Revendors the bbolt from `v1.3.6` to `v1.3.7`
```


shreyas-s-rao (approved):
@ishan16696 thanks for the PR. I don't think it is necessary to use the exact same version of bbolt in both `backup-restore` and `etcd` containers, since the API used for communication is the etcd API, and as long as etcd v3 API is supported and used by both, there should be no issues.

Nevertheless, any upgrades in dependencies are welcome, since they bring improvements and/or security fixes.
/lgtm

aaronfern:
Another point of improvement would be to update the etcd version vendored in backup-restore to 3.4.26. Currently it is still at 3.4.13

ishan16696:
> Another point of improvement would be to update the etcd version vendored in backup-restore to 3.4.26. Currently it is still at 3.4.13

I remembered, you have vendored etcd version 3.4.26 in etcd-wrapper

PR #660, opened 2023-08-23
Adds sleep command at the end of compaction.

abdasgupta:
This PR adds a provision of sleep duration  after the `Compact` function done with uploading snapshot. This sleep duration is needed as prometheus is required to collect necessary metrics for network activity after a fast full snapshot upload. The sleep duration is supplied as flag to `etcdbrctl compact` command and set to 0 by default. But the flag can be configured by any agent, who is using `etcdbrctl`, to set any duration value like 60s, 2m etc. `60s` value of sleep duration is enough to allow sufficient time for Prometheus to scrape network related metrics after a fast full snapshot upload.

Fixes #
https://github.com/gardener/etcd-druid/pull/648

**Release note**:

```noteworthy user
Introduce flag `metrics-scrape-wait-duration` to `etcdbrctl compact` command, that specifies a wait duration at the end of a snapshot compaction, to allow Prometheus to scrape metrics related to compaction before the `etcdbrctl` process exits.
```

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -182,7 +182,21 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 	}
 
 	// Add a sleep command so that prometheus can collect necessary metrics related to the uploading of snapshots. see https://github.com/gardener/etcd-druid/issues/648
-	time.Sleep(opts.SleepTimeout.Duration)
+	err = SleepWithContext(ctx, opts.SleepFor.Duration)
+	if err != nil {
+		cp.logger.Warnf("Could not sleep for specified duration: %v", err)
+	}
 
 	return snapshot, nil
+}
+
+func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(sleepFor):
+			return nil
+		}
+	}
 }
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -182,7 +182,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 	}
 
 	// Add a sleep command so that prometheus can collect necessary metrics related to the uploading of snapshots. see https://github.com/gardener/etcd-druid/issues/648
-	err = SleepWithContext(ctx, opts.SleepFor.Duration)
+	err = sleepWithContext(ctx, opts.SleepForScrape.Duration)
 	if err != nil {
 		cp.logger.Warnf("Could not sleep for specified duration: %v", err)
 	}
@@ -190,7 +190,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 	return snapshot, nil
 }
 
-func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
+func sleepWithContext(ctx context.Context, sleepFor time.Duration) error {
 	for {
 		select {
 		case <-ctx.Done():
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -182,7 +182,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 	}
 
 	// Add a sleep command so that prometheus can collect necessary metrics related to the uploading of snapshots. see https://github.com/gardener/etcd-druid/issues/648
-	err = sleepWithContext(ctx, opts.SleepForScrape.Duration)
+	err = sleepWithContext(ctx, opts.MetricsScrapeWaitDuration.Duration)
 	if err != nil {
 		cp.logger.Warnf("Could not sleep for specified duration: %v", err)
 	}
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -181,5 +181,8 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 		cancel()
 	}
 
+	// Add a sleep command so that prometheus can collect necessary metrics related to the uploading of snapshots. see https://github.com/gardener/etcd-druid/issues/648
+	time.Sleep(opts.SleepTimeout.Duration)
+
 	return snapshot, nil
 }
--- a/pkg/compactor/compactor_test.go
+++ b/pkg/compactor/compactor_test.go
@@ -39,8 +39,8 @@ var _ = Describe("Running Compactor", func() {
 		embeddedEtcdQuotaBytes int64  = 8 * 1024 * 1024 * 1024
 		snapshotTimeout               = 30 * time.Second
 		defragTimeout                 = 30 * time.Second
-		// keep SleepFor low so that tests are faster
-		sleepFor            = 10 * time.Nanosecond
+		// keep SleepForScrape low so that tests are faster
+		sleepForScrape      = 10 * time.Nanosecond
 		needDefragmentation = true
 	)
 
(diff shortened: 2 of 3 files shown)


abdasgupta:
cc @istvanballok

istvanballok (commented):
/lgtm

pkg/compactor/compactor.go
```diff
@@ -181,5 +181,8 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions)
 		cancel()
 	}
 
+	// Add a sleep command so that prometheus can collect necessary metrics related to the uploading of snapshots. see https://github.com/gardener/etcd-druid/issues/648
+	time.Sleep(60 * time.Second)
```
unmarshall, line 185: Since  a context is passed to this method a better way to achieve this would be:
```go
func SleepWithContext(ctx context.Context, sleepFor time.Duration) error {
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(sleepFor):
			return nil
		}
	}
}
```

istvanballok (commented):
lgtm

pkg/compactor/compactor_test.go
```diff
@@ -39,7 +39,9 @@ var _ = Describe("Running Compactor", func() {
 		embeddedEtcdQuotaBytes int64  = 8 * 1024 * 1024 * 1024
 		snapshotTimeout               = 30 * time.Second
 		defragTimeout                 = 30 * time.Second
-		needDefragmentation           = true
+		// keep SleepFor low so that tests are faster
+		sleepFor            = 10 * time.Nanosecond
```
shreyas-s-rao, line 43: Please change this value to 0 seconds, since we don't really test the scraping of metrics in unit tests and rightly so. Unit tests should be very quick, and should not include any code that simply sleeps, unless the sleep is actually used for checking some functionality. Even if so, such tests should ideally be part of e2e tests.

pkg/types/compactor.go
```diff
@@ -27,6 +27,8 @@ const (
 	defaultDefragTimeout time.Duration = 8 * time.Minute
 	// defaultSnapshotTimeout defines default timeout duration for taking compacted FullSnapshot.
 	defaultSnapshotTimeout time.Duration = 30 * time.Minute
+	//defaultSleepForInSeconds defines default timeout for sleep command in compaction
+	defaultSleepFor time.Duration = 60 * time.Second
```
shreyas-s-rao, line 31:
```suggestion
	defaultSleepFor time.Duration = 0
```

The default behavior should ideally not break any previous behavior of the code. Since there was never a sleep previously, the default for this value should be zero, and should be explicitly set to `60 seconds` by druid when deploying backup-restore, for the gardener usecase.

pkg/types/compactor.go
```diff
@@ -65,6 +70,7 @@ func (c *CompactorConfig) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.FullSnapshotLeaseName, "full-snapshot-lease-name", c.FullSnapshotLeaseName, "full snapshot lease name")
 	fs.StringVar(&c.DeltaSnapshotLeaseName, "delta-snapshot-lease-name", c.DeltaSnapshotLeaseName, "delta snapshot lease name")
 	fs.BoolVar(&c.EnabledLeaseRenewal, "enable-snapshot-lease-renewal", c.EnabledLeaseRenewal, "Allows compactor to renew the full snapshot lease when successfully compacted snapshot is uploaded")
+	fs.DurationVar(&c.SleepFor.Duration, "sleep-timeout-in-seconds", c.SleepFor.Duration, "Timeout for sleep command after the snapshot upload is over")
```
shreyas-s-rao, line 73:
```suggestion
	fs.DurationVar(&c.SleepFor.Duration, "sleep-for-seconds", c.SleepFor.Duration, "Sleep duration (in seconds) after compaction is completed")
```

The term `timeout` signifies that some action is to be performed before the specified timeout is elapsed. But here this isn't the case. Rather, this is a `sleep` after the compaction operation is completed, only so that prometheus can have enough time to scrape metrics. hence removed the term `timeout`.

shreyas-s-rao (changes requested):
@abdasgupta thanks for the PR. I have a few suggestions/comments. PTAL, thanks!

pkg/types/compactor.go
unmarshall: Please also ensure that you have addressed the [review comment](https://github.com/gardener/etcd-backup-restore/pull/660#discussion_r1303776914)

unmarshall: The field name `SleepFor` is too generic and does not tell anything about the intended use case. It is only there to allow prometheus to scrape metrics. We should then call it `MetricsScrapeWaitDuration` instead or something similar.

pkg/compactor/compactor.go
abdasgupta: Done

pkg/types/compactor.go
```diff
@@ -46,7 +46,7 @@ type CompactorConfig struct {
 	DeltaSnapshotLeaseName string            `json:"deltaSnapshotLeaseName,omitempty"`
 	EnabledLeaseRenewal    bool              `json:"enabledLeaseRenewal"`
 	// see https://github.com/gardener/etcd-druid/issues/648
-	SleepFor wrappers.Duration `json:"sleepFor,omitempty"`
+	SleepForScrape wrappers.Duration `json:"sleepForScrape,omitempty"`
```
shreyas-s-rao, line 49: There is no mention of "time" or "duration" in this field name. It simply says "sleep for scrape", which could be a boolean value for all we know, which could mean "if set to true, compactor will sleep for some scrape". There is no mention of what is being scraped, or what kind of value this field can take.

As mentioned in https://github.com/gardener/etcd-backup-restore/pull/660#discussion_r1308493167, please change this field name to make it more meaningful, for both developers and users of etcdbr.

pkg/compactor/compactor_test.go
```diff
@@ -39,7 +39,9 @@ var _ = Describe("Running Compactor", func() {
 		embeddedEtcdQuotaBytes int64  = 8 * 1024 * 1024 * 1024
 		snapshotTimeout               = 30 * time.Second
 		defragTimeout                 = 30 * time.Second
-		needDefragmentation           = true
+		// keep SleepForScrape 0 so that tests are faster
```
shreyas-s-rao, line 42: This value is set to 0 primarily because there is no necessity for scraping metrics. Tests being faster is a side benefit of this. Please mention this.

Or better yet, simply remove this variable and directly set the field to `0` when creating `compactorConfig` in the test.

shreyas-s-rao (changes requested):
@abdasgupta please check the provided suggestions and address them. Thanks.

pkg/compactor/compactor_test.go
abdasgupta: Done

pkg/types/compactor.go
abdasgupta: Done

shreyas-s-rao (changes requested):
@abdasgupta thanks for making the changes. I have a few comments regarding the changes made. PTAL, thanks

shreyas-s-rao:
@abdasgupta thanks for making the changes. Please adapt the PR description and release note to reflect the new flag name and behavior.

Also, I would ask you to not force-push to PRs in the future, since it becomes very difficult to review the new changes that were made. Instead, you may choose to add a squash-merge label to the PR, so that whoever is merging the PR will squash-merge the PR and leave out the `addressed review comments` commits.

shreyas-s-rao:
The release note needs to be more precise and correctly formatted. See below example:

```
Introduce flag `metrics-scrape-wait-duration` to `etcdbrctl compact` command, that specifies a wait duration at the end of a snapshot compaction, to allow Prometheus to scrape metrics related to compaction before the `etcdbrctl` process exits.
```

shreyas-s-rao (approved):
/lgtm

Issue #556, opened 2022-11-11
[Feature] Defragmentor of backup-restore should also consider the etcd db size along with scheduled defrag


ishan16696:
**Feature (What you would like to be added):**
[Defragmentor logic](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/defragmentor/defrag.go#L106-L107) of backup-restore should also consider the `etcd db` size along with schedule time of defragmentation and if Defragmentor observed that `etcd db` size has crossed the certain limit(say 7Gi) then it should trigger the out-of-schedule defragmentation on etcd so that `etcd db` size won't crosses the practical limit of 8Gi.

**Motivation (Why is this needed?):**
It has been observed that sometimes etcd's db size abruptly increases and cross the practical limit of 8Gi which leads to unavailability of etcd cluster. To resolve this we require the manual intervention to trigger the [defragmentation on etcd](https://etcd.io/docs/v3.2/op-guide/maintenance/#defragmentation)

**Approach/Hint to the implement solution (optional):**

ishan16696:
This feature can also mitigate `NOSPACE` alarm https://github.com/gardener/etcd-backup-restore/issues/531

ishan16696:
There are 2 kind of **bad cases** possible :

1. One is defrag is helping i.e etcd-db size getting reduce but etcd db is getting full very fast again and again, in such cases triggering defrag again and again is overall kill.
3. In other cases, defragmentation has no effect, likely because the database consists of unique keys, leaving no space to be reclaimed.

Issue #531, opened 2022-09-06
[Feature] Backup-restore should detects alarm and remove the  `alarm:NOSPACE` from etcd.


ishan16696:
**Feature (What you would like to be added):**
Backup-restore should be able to detect the alarm(if any cluster member have) and backup-restore should be able to take action accordingly and remove the alarm from etcd.

**Motivation (Why is this needed?):**
Currently etcd `--quota-backend-bytes` is to `8Gi` and when etcd backend database size exceeds the given quota, etcd raises a cluster-wide alarm that puts the cluster into a maintenance mode which only accepts reads and deletes operations.
So, to fix this we have to trigger the manual defragmentation on etcd which free-up the etcd space and we have to manually remove the alarm `alarm:NOSPACE` as backup-restore don't have any functionality to remove the alarm.

```
# disarm alarm
$ ETCDCTL_API=3 etcdctl alarm disarm
memberID:13803658152347727308 alarm:NOSPACE
```

**Approach/Hint to the implement solution (optional):**

PR #668, opened 2023-10-10
Make embedded etcd lean during restoration to avoid any restoration failure

### LINKED ISSUE #636: Periodically run defragmentation on the embedded etcd during restoration
Periodically run (compaction and) defragmentation on the embedded etcd during restoration, either after every X deltas applied or after every Y GB of etcd DB size reached

Restoration of large backups whose events add up to more than the etcd quota (set to default of 8GB) will fail. This especially hurts during snapshot compactions, where restoration is run on backups containing large number of events (defaulting to 1 million events), and restoration is quite likely to fail. It now becomes necessary to defragment the etcd DB during restoration, ie, while applying delta snapshots.

The criteria to run these defragmentations can either:
1. be based on DB size (which is a large topic by itself, covered in https://github.com/gardener/etcd-backup-restore/issues/556), or
2. restorer can simply run (compaction and) defragmentation after applying every `X` number of delta snapshots
  **i.** Since the maximum size of a delta snapshot is determined by `delta-snapshot-memory-limit`, and the embedded etcd quota size is determined by `embedded-etcd-quota-bytes`, we can calculate the maximum number of delta snapshots that can be restored before a defragmentation is required, using the formula `(embedded-etcd-quota-bytes / delta-snapshot-memory-limit) / 2` (additionally diving by `2` to defragment after approximately half the etcd DB size quota is reached), giving us the maximum allowed number of deltas for safely restoring the etcd.
  **ii.** restorer can keep track of the total size of delta snapshots it has applied, and defragment once half the etcd DB size quota is reached.

ishan16696:
It has been observed that sometimes during big restoration, `Embedded Etcd database space exceeds` causing the restoration failure.
This PR makes an embedded etcd lean during restoration to avoid any failure due to `etcd database space exceeds`.
To make embedded etcd lean:
- backup-restore will periodically run compaction on embedded etcd.
- backup-restore will keep a check on database size, and if it crosses certain threshold, backup-restore trigger defragmentation.

Fixes #636 

**Release note**:

```noteworthy operator
Fix a restoration failure which can occurs due to an etcd database space exceeds during restoration.
```

### DIFF
--- a/pkg/snapshot/restorer/restorer.go
+++ b/pkg/snapshot/restorer/restorer.go
@@ -57,7 +57,11 @@ import (
 )
 
 const (
-	etcdDialTimeout = time.Second * 30
+	etcdConnectionTimeout                     = 30 * time.Second
+	etcdCompactTimeout                        = 2 * time.Minute
+	etcdDefragTimeout                         = 5 * time.Minute
+	periodicallyMakeEtcdLean                  = 10
+	thresholdPercentageForDBSizeAlarm float64 = .8
 )
 
 // Restorer is a struct for etcd data directory restorer
@@ -113,33 +117,26 @@ func (r *Restorer) Restore(ro brtypes.RestoreOptions, m member.Control) (*embed.
 
 	defer func() {
 		if err := os.RemoveAll(ro.Config.TempSnapshotsDir); err != nil {
-			r.logger.Errorf("Failed to remove restoration temp directory %s: %v", ro.Config.TempSnapshotsDir, err)
+			r.logger.Errorf("failed to remove restoration temp directory %s: %v", ro.Config.TempSnapshotsDir, err)
 		}
 	}()
 
-	r.logger.Infof("Starting embedded etcd server...")
+	r.logger.Infof("Starting an embedded etcd server...")
 	e, err := miscellaneous.StartEmbeddedEtcd(r.logger, &ro)
 	if err != nil {
 		return e, err
 	}
 
+	embeddedEtcdEndpoints := []string{e.Clients[0].Addr().String()}
+
 	clientFactory := etcdutil.NewClientFactory(ro.NewClientFactory, brtypes.EtcdConnectionConfig{
 		MaxCallSendMsgSize: ro.Config.MaxCallSendMsgSize,
-		Endpoints:          []string{e.Clients[0].Addr().String()},
+		Endpoints:          embeddedEtcdEndpoints,
 		InsecureTransport:  true,
 	})
-	clientKV, err := clientFactory.NewKV()
-	if err != nil {
-		return e, err
-	}
-	defer func() {
-		if err := clientKV.Close(); err != nil {
-			r.logger.Errorf("Failed to close etcd KV client: %v", err)
-		}
-	}()
 
 	r.logger.Infof("Applying delta snapshots...")
-	if err := r.applyDeltaSnapshots(clientKV, ro); err != nil {
+	if err := r.applyDeltaSnapshots(clientFactory, embeddedEtcdEndpoints, ro); err != nil {
 		return e, err
 	}
 
@@ -148,7 +145,11 @@ func (r *Restorer) Restore(ro brtypes.RestoreOptions, m member.Control) (*embed.
 		if err != nil {
 			return e, err
 		}
-		defer clientCluster.Close()
+		defer func() {
+			if err := clientCluster.Close(); err != nil {
+				r.logger.Errorf("failed to close etcd cluster client: %v", err)
+			}
+		}()
 		m.UpdateMemberPeerURL(context.TODO(), clientCluster)
 	}
 	return e, nil
@@ -411,7 +412,28 @@ func makeWALAndSnap(logger *zap.Logger, walDir, snapDir string, cl *membership.R
 }
 
 // applyDeltaSnapshots fetches the events from delta snapshots in parallel and applies them to the embedded etcd sequentially.
-func (r *Restorer) applyDeltaSnapshots(clientKV client.KVCloser, ro brtypes.RestoreOptions) error {
+func (r *Restorer) applyDeltaSnapshots(clientFactory client.Factory, endPoints []string, ro brtypes.RestoreOptions) error {
+
+	clientKV, err := clientFactory.NewKV()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := clientKV.Close(); err != nil {
+			r.logger.Errorf("failed to close etcd KV client: %v", err)
+		}
+	}()
+
+	clientMaintenance, err := clientFactory.NewMaintenance()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := clientMaintenance.Close(); err != nil {
+			r.logger.Errorf("failed to close etcd maintenance client: %v", err)
+		}
+	}()
+
 	snapList := ro.DeltaSnapList
 	numMaxFetchers := ro.Config.MaxFetchers
 
@@ -420,6 +442,9 @@ func (r *Restorer) applyDeltaSnapshots(clientKV client.KVCloser, ro brtypes.Rest
 	if err := r.applyFirstDeltaSnapshot(clientKV, firstDeltaSnap); err != nil {
 		return err
 	}
+
+	embeddedEtcdQuotaBytes := float64(ro.Config.EmbeddedEtcdQuotaBytes)
+
 	if err := verifySnapshotRevision(clientKV, snapList[0]); err != nil {
 		return err
 	}
@@ -430,22 +455,28 @@ func (r *Restorer) applyDeltaSnapshots(clientKV client.KVCloser, ro brtypes.Rest
 	}
 
 	var (
-		remainingSnaps  = snapList[1:]
-		numSnaps        = len(remainingSnaps)
-		numFetchers     = int(math.Min(float64(numMaxFetchers), float64(numSnaps)))
-		snapLocationsCh = make(chan string, numSnaps)
-		errCh           = make(chan error, numFetchers+1)
-		fetcherInfoCh   = make(chan brtypes.FetcherInfo, numSnaps)
-		applierInfoCh   = make(chan brtypes.ApplierInfo, numSnaps)
-		stopCh          = make(chan bool)
-		wg              sync.WaitGroup
+		remainingSnaps    = snapList[1:]
+		numSnaps          = len(remainingSnaps)
+		numFetchers       = int(math.Min(float64(numMaxFetchers), float64(numSnaps)))
+		snapLocationsCh   = make(chan string, numSnaps)
+		errCh             = make(chan error, numFetchers+1)
+		fetcherInfoCh     = make(chan brtypes.FetcherInfo, numSnaps)
+		applierInfoCh     = make(chan brtypes.ApplierInfo, numSnaps)
+		wg                sync.WaitGroup
+		stopCh            = make(chan bool)
+		stopHandleAlarmCh = make(chan bool)
+		dbSizeAlarmCh     = make(chan string)
+		dbSizeDisAlarmCh  = make(chan bool)
 	)
 
-	go r.applySnaps(clientKV, remainingSnaps, applierInfoCh, errCh, stopCh, &wg)
+	go r.applySnaps(clientKV, clientMaintenance, remainingSnaps, dbSizeAlarmCh, dbSizeDisAlarmCh, applierInfoCh, errCh, stopCh, &wg, endPoints, embeddedEtcdQuotaBytes)
 
 	for f := 0; f < numFetchers; f++ {
 		go r.fetchSnaps(f, fetcherInfoCh, applierInfoCh, snapLocationsCh, errCh, stopCh, &wg, ro.Config.TempSnapshotsDir)
 	}
+
+	go r.handleAlarm(stopHandleAlarmCh, dbSizeAlarmCh, dbSizeDisAlarmCh, clientMaintenance)
+	defer close(stopHandleAlarmCh)
 
 	for i, remainingSnap := range remainingSnaps {
 		fetcherInfo := brtypes.FetcherInfo{
@@ -456,7 +487,7 @@ func (r *Restorer) applyDeltaSnapshots(clientKV client.KVCloser, ro brtypes.Rest
 	}
 	close(fetcherInfoCh)
 
-	err := <-errCh
+	err = <-errCh
 
 	if cleanupErr := r.cleanup(snapLocationsCh, stopCh, &wg); cleanupErr != nil {
 		r.logger.Errorf("Cleanup of temporary snapshots failed: %v", cleanupErr)
@@ -540,13 +571,22 @@ func (r *Restorer) fetchSnaps(fetcherIndex int, fetcherInfoCh <-chan brtypes.Fet
 }
 
 // applySnaps applies delta snapshot events to the embedded etcd sequentially, in the right order of snapshots, regardless of the order in which they were fetched.
-func (r *Restorer) applySnaps(clientKV client.KVCloser, remainingSnaps brtypes.SnapList, applierInfoCh <-chan brtypes.ApplierInfo, errCh chan<- error, stopCh <-chan bool, wg *sync.WaitGroup) {
+func (r *Restorer) applySnaps(clientKV client.KVCloser, clientMaintenance client.MaintenanceCloser, remainingSnaps brtypes.SnapList, dbSizeAlarmCh chan string, dbSizeDisAlarmCh <-chan bool, applierInfoCh <-chan brtypes.ApplierInfo, errCh chan<- error, stopCh <-chan bool, wg *sync.WaitGroup, endPoints []string, embeddedEtcdQuotaBytes float64) {
 	defer wg.Done()
 	wg.Add(1)
 
+	// To reduce or to stop the growing size of embedded etcd database during restoration
+	// it's important to track number of delta snapshots applied to an embedded etcd
+	// to raise an alarm(if required).
+	// it is initialize with "1" as backup-restore has already applied first delta snapshot.
+	numberOfDeltaSnapApplied := 1
+
+	// A flag to track whether a previous attempt to make embedded etcd lean failed or succeeds.
+	// If failed then backup-restore should re-try after applying next delta snapshot.
+	prevAttemptToMakeEtcdLeanFailed := false
+
 	pathList := make([]string, len(remainingSnaps))
 	nextSnapIndexToApply := 0
-
 	for {
 		select {
 		case _, more := <-stopCh:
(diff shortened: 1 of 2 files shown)


aaronfern (commented):
One point from me

pkg/snapshot/restorer/restorer.go
```diff
@@ -887,3 +945,84 @@ func ErrorArrayToError(errs []error) error {
...
+	for {
+		select {
+		case <-stopHandleAlarmCh:
+			r.logger.Info("Closing handle alarm...")
+			return
+		case endPoint := <-dbSizeAlarmCh:
+			r.logger.Info("Received a dbsize alarm")
+			r.logger.Infof("Calling defrag on endpoint: [%v]", endPoint)
+			if err := func() error {
+				ctx, cancel := context.WithTimeout(context.TODO(), etcdDefragTimeout)
```
aaronfern, line 961:
```suggestion
				ctx, cancel := context.WithTimeout(context.Background(), etcdDefragTimeout)
```
It's better to use `context.Background()` if no parent context is passed to the function. Similar for other places where `context.TODO()` is used

aaronfern (commented):
One point from me

pkg/snapshot/restorer/restorer.go
```diff
@@ -887,3 +944,86 @@ func ErrorArrayToError(errs []error) error {
...
+	defer cancel()
+
+	// check database size of embedded etcdServer.
+	status, err := clientMaintenance.Status(ctx, endPoints[0])
+	if err != nil {
+		return fmt.Errorf("unable to check embedded etcd status: %v", err)
+	}
+
+	if float64(status.DbSizeInUse) > thresholdPercentageForDBSizeAlarm*embeddedEtcdQuotaBytes ||
+		float64(status.DbSize) > thresholdPercentageForDBSizeAlarm*embeddedEtcdQuotaBytes {
```
seshachalam-yv, line 1002: After compaction, trigger defragmentation based on the difference between `DbSizeInUse` and `DbSize`, instead of using a threshold size for `DbSizeInUse`, `DbSize`
WDYT @ishan16696  ?

seshachalam-yv (changes requested):

pkg/snapshot/restorer/restorer.go
ishan16696: > based on the difference between DbSizeInUse and DbSize

lets assume `DbSize=7Gi` and `DbSizeInUse=6Gi`
then difference b/w `DbSize-DbSizeInUse=1Gi` then you still need some kind of benchmark(or threshold) to decide which way to go , isn't ?
And IMO difference b/w `DbSize` and `DbSizeInUse` can leads to some edge cases like when difference b/w 2 dbsize turned out to be `0` and so on, so why not just keep everything simple and use just a threshold with dbsize rather then some kind of difference which may leads to edge cases. WDYT ?

pkg/snapshot/restorer/restorer.go
```diff
@@ -430,23 +455,29 @@ func (r *Restorer) applyDeltaSnapshots(clientKV client.KVCloser, ro brtypes.Rest
...
+		numFetchers       = int(math.Min(float64(numMaxFetchers), float64(numSnaps)))
+		snapLocationsCh   = make(chan string, numSnaps)
+		errCh             = make(chan error, numFetchers+1)
+		fetcherInfoCh     = make(chan brtypes.FetcherInfo, numSnaps)
+		applierInfoCh     = make(chan brtypes.ApplierInfo, numSnaps)
+		wg                sync.WaitGroup
+		stopCh            = make(chan bool)
+		stopHandleAlarmCh = make(chan bool)
+		dbSizeAlarmCh     = make(chan string)
+		dbSizeDisAlarmCh  = make(chan bool)
```
shreyas-s-rao, line 469: I feel we are increasing complexity by adding these many channels. I understand that the purpose of `HandleAlarm()` and these channels is to help catch an alarm from etcd and then disarm it. But do we need to do this so proactively is my question. Let's say we are applying a delta snapshot and suddenly the threshold is breached, and an etcd alarm is raised. By the time the alarm is handled and disarmed asynchronously via these channels, many delta events would have failed to apply, right?

Instead, we anyway already have `periodicallyMakeEtcdLean`, which ensures that the etcd is compacted+defragmented after applying every X delta snapshots. I think this should suffice for now, rather than trying to handle alarms that are raised when the DB size breaches the quota.

The general solution for this problem is to use DB-size based defragmentation, which will be coming in steward, and can be switched on for restoration as well as for regular etcd functioning. Once we have this, then we don't need any special handling for restoration case. WDYT?

shreyas-s-rao (commented):
@ishan16696 thanks for the PR. I feel that this PR increases complexity of an already complex piece of code. We are now introducing 3 more channels, which will only make coordination between the `HandleAlarm()` and other methods that much more complex. Instead, I would suggest we focus on solving a very specific problem in this PR by simply performing defrag after every X delta snapshots or events, and take care of alarm handling as part of https://github.com/gardener/etcd-steward/issues/6. WDYT?

shreyas-s-rao:
@ishan16696 another suggestion that @seshachalam-yv and I had, was to do this - before applying every delta snapshot, simply fetch the `dbSize`, and check if the sum of `dbSize` and `DeltaSnapshotMemoryLimit` exceeds the `etcdDBQuota`, then trigger compaction+defragmentation, and then continue with applying the delta snapshot.

This way, we know that restoration will never run into quota limit alarm, so we don't even need to worry about handling the alarm, atleast for restoration case.

ishan16696:
> Instead, I would suggest we focus on solving a very specific problem in this PR

I think you have misunderstood the terminology which I have used in this PR i.e `Alarm`. Let me explain this:
- There are etcd alarm like `NOSPACE alarm`,  which I have described in this issue: https://github.com/gardener/etcd-backup-restore/issues/531. This PR **DOES NOT** address/tackle this issue of `etcd Alarm` in any way. I would suggest please go through the code once again. let me know if you have any doubt.
- In this PR I have introduced a `backup-restore Alarm` concept (don't get confuse with `etcd alarm`). In this `backup-restore alarm` func which will run handle the case like `DBSizeAlarm` which means when etcd DB size is about to crosses the threshold, hence it will take action like `defrag` 
- now, you may ask why not simply called a `defrag` ?
   Ans to this is right now we are/will be calling `defrag` over 3 times in different places
    1. A schedule defrag
    2. defrag during restoration this is done in this PR
    3. defrag `out-of-schdule`: Please refer to this: https://github.com/gardener/etcd-backup-restore/issues/556 which I think we need to increase the priority as etcd-steward will take time and issue of `tcdserver: mvcc: database space exceeded` occurs more frequently and it has been asked also asked by g/g
    So, the logic behind is .. start a func in backup-restore which can handle any kind of ops (I called them `backup-restore alarm`), to take necessary action.
  We can just copy this same functionality to etcd-steward as well.


>  I would suggest we focus on solving a very specific problem in this PR by simply performing defrag after every X delta snapshots or events

The PR is only handle the what it meant to be. It only fixes the big restoration of etcd by calling compaction and defragmentaion(if required). Please check the PR description, I have mentioned it clearly.

> I feel that this PR increases complexity of an already complex piece of code

no, It doesn't. It takes away the complexity of other issue which anyway some body have to write. 
It also make very easy for me to write unit tests, Please check the tests [here](https://github.com/gardener/etcd-backup-restore/pull/668/files#diff-230e8e1897f9e0bcbbb2062bb32fd5ee94d1fdc8fa444efe6dd120f7281103b2), you will understand

pkg/snapshot/restorer/restorer.go
ishan16696: > these channels is to help catch an alarm from etcd and then disarm it.

nopes, you have misunderstood the code. Please refer to this comment: https://github.com/gardener/etcd-backup-restore/pull/668#issuecomment-1818283375
Let me know if you have doubts.

ishan16696:
> before applying every delta snapshot, simply fetch the dbSize, and check if the sum of dbSize and DeltaSnapshotMemoryLimit exceeds the etcdDBQuota, then trigger compaction+defragmentation, and then continue with applying the delta snapshot.

there is no need to check `dbsize` on every deltaSnapshot. There are 2 reasons to that:
1. Checking the dbsize, you can make unncesary API call to embedded etcd, let's say you are apply 5k delta snapshots, you end up making 5k API calls just to check dbsize.
2. We alredy know the maximum size of delta snapshot is `100MB`, then why not use that . I did the same thing. 

I'm calling compaction on every `10 delta-snapshots` applied i.e 10*100Mi = 1Gi then we apply compaction and check the threshold.

ishan16696:
@shreyas-s-rao and @seshachalam-yv I would suggest you guys, please go through the unit test which I have wrote in this PR, then you will have better understanding of each scenario. Thanks

pkg/snapshot/restorer/restorer.go
```diff
@@ -57,7 +57,11 @@ import (
 )
 
 const (
-	etcdDialTimeout = time.Second * 30
+	etcdConnectionTimeout                     = 30 * time.Second
+	etcdCompactTimeout                        = 2 * time.Minute
+	etcdDefragTimeout                         = 5 * time.Minute
+	periodicallyMakeEtcdLean                  = 10
```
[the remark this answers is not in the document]
ishan16696: > The current name does not specify delta snapshots anywhere. Why not call it etcdCompactDefragIntervalNumDeltaSnapshots instead, just as a suggestion?

IMO, `etcdCompactDefragIntervalNumDeltaSnapshots` it is very large variable name and as it is embedded etcd, if you embedded as well then variable name will become more large i.e `embeddedEtcdCompactDefragIntervalNumDeltaSnapshots` 

>  It indicates that a compact+defrag is performed, rather than a generic make lean term, 
yeah, I deliberately choose generic term, `makeEtcdLean` as symbol or name for abstraction, How it should be done or functionality that shouldn't be part of variable name.

> and also talks about the interval in terms of num delta snapshots intervals. WDYT?

yeah, this I can try to include.

shreyas-s-rao:
> We alredy know the maximum size of delta snapshot is 100MB, then why not use that . I did the same thing.
I'm calling compaction on every 10 delta-snapshots applied i.e 10*100Mi = 1Gi then we apply compaction and check the threshold.

We cannot assume that this is set to 100Mi by default. It could be set to any value. If for instance, it's set to 1Gi, then by the time we reach the 10th delta snapshot we would have run into a space quota alarm from etcd. Instead, I would recommend to either make this configurable, or to calculate the number of deltas based on the configured delta snapshot memory limit. That's the safer option.

ishan16696:
> We cannot assume that this is set to 100Mi by default. It could be set to any value. If for instance, it's set to 1Gi, then by the time we reach the 10th delta snapshot we would have run into a space quota alarm from etcd. 

`DeltaSnapshotMemoryLimit` default value is `10Mi`, in production it is used as `100Mi`, so I don't think anyone will set DeltaSnapshotMemoryLimit to `1Gi` and called it as a delta snapshot .... seems no longer delta to me 😅 . Even if somebody has set `DeltaSnapshotMemoryLimit` to some large value like `1Gi`, delta snapshot will be taken with this accordingly with this: `min(DeltaSnapshotMemoryLimit, DeltaSnapshotPeriod)` calculation, so `DeltaSnapshotPeriod` also determines and limit the size of deltaSnapshot. Assume `DeltaSnapshotPeriod` is set 10mins or 15mins then in a normal scenario, 1Gi of data can't come within 15mins of interval.

> Instead, I would recommend to either make this configurable, or to calculate the number of deltas based on the configured delta snapshot memory limit. That's the safer option.

I don't want to make it configurable as we already have so many flags to configure and this is an abstraction which is only going to use in our restoration case, end-user shouldn't bother or shouldn't know about it.

ishan16696:
From the gardener use case `periodicallyMakeEtcdLean=10` is fine as we know `DeltaSnapshotMemoryLimit` is set to `100Mi`.
And for community users, I tried to calculate `periodicallyMakeEtcdLean` using `DeltaSnapshotMemoryLimit` but I faced many challenges like edge cases:
- it is also difficult to fetch `deltaSnapshotMemoryLimit` it from `snapshotterConfig`  as end-user might not use `server`  subcommand, they may be used only `restore`  sub-command. In that case, it will be difficult to calculate `deltaSnapshotMemoryLimit`.
- may be end-user is using `snapshot` sub-command only to take snapshot but when they will try to run `restore` sub-command, how do you make sure they have passed the same `deltaSnapshotMemoryLimit` in `restore` sub-command as they have passed in `snapshot` sub-command.

IMO, we can keep things simple and keep `periodicallyMakeEtcdLean` to value `10`.

pkg/snapshot/restorer/restorer.go
```diff
@@ -887,3 +944,86 @@ func ErrorArrayToError(errs []error) error {
...
+func (r *Restorer) MakeEtcdLeanAndCheckAlarm(revision int64, endPoints []string, embeddedEtcdQuotaBytes float64, dbSizeAlarmCh chan string, dbSizeDisAlarmCh <-chan bool, clientKV client.KVCloser, clientMaintenance client.MaintenanceCloser) error {
+	if err := func() error {
+		ctx, cancel := context.WithTimeout(context.Background(), etcdCompactTimeout)
+		defer cancel()
+		if _, err := clientKV.Compact(ctx, revision, clientv3.WithCompactPhysical()); err != nil {
+			return fmt.Errorf("Compact API call failed: %v", err)
+		}
+		return nil
+	}(); err != nil {
+		return err
```
seshachalam-yv, line 988: Why is there an anonymous function used here?
It appears that within this function, an error is being handled. Are you expecting the anonymous function to return an error, and then, if an error is present, to return that error from the main function

pkg/snapshot/restorer/restorer.go
```diff
@@ -887,3 +944,86 @@ func ErrorArrayToError(errs []error) error {
...
+	}
+	r.logger.Infof("Successfully compacted embedded etcd till revision: %v", revision)
+
+	ctx, cancel := context.WithTimeout(context.Background(), etcdConnectionTimeout)
```
seshachalam-yv, line 992: Multiple contexts are created in the `MakeEtcdLeanAndCheckAlarm` function. Could we simplify this by using a single context with a deadline for the entire process?

seshachalam-yv (changes requested):

pkg/snapshot/restorer/restorer.go
seshachalam-yv: 😕

seshachalam-yv:
> IMO, we can keep things simple and keep `periodicallyMakeEtcdLean` to value `10`.

I see that you've set `periodicallyMakeEtcdLean` to `10`, presumably based on the idea that each delta snapshot might be up to `100 MB`, adding up to a total of `1 GB`. However, I suggest a different approach. Instead of relying on a fixed count of delta snapshots, it's more effective to monitor their cumulative size. We should initiate compaction and defragmentation when they collectively reach the `1 GB` threshold. This is when the `MakeEtcdLean` function would be most effective. 

@ishan16696 Could you please reconsider this logic?

seshachalam-yv:
> no, It doesn't. It takes away the complexity of other issue which anyway some body have to write.
> It also make very easy for me to write unit tests, Please check the tests [here](https://github.com/gardener/etcd-backup-restore/pull/668/files#diff-230e8e1897f9e0bcbbb2062bb32fd5ee94d1fdc8fa444efe6dd120f7281103b2), you will understand

I've reviewed your code and noticed you're using `gomock` to mock functions for testing. Could you explain how the added complexity in the code is actually facilitating easier test writing? 

Additionally, I'm curious about the rationale behind using a separate goroutine, `go r.HandleAlarm(...)`, specifically for triggering defragmentation. It's not immediately clear why this is necessary, and understanding your reasoning here would be helpful.

pkg/snapshot/restorer/restorer.go
ishan16696: > anonymous function used here

anonymous function is used to cancel the ctx using `defer cancel()` , defer is execute at the end of function, hence there was no choice but to go with anonymous function.

> It appears that within this function, an error is being handled. Are you expecting the anonymous function to return an error, and then, if an error is present, to return that error from the main function

please check the whole code, when anonymous function returns error we log that error send the `dbSizeDisAlarmCh <- false` to channel to indicate it fails

pkg/snapshot/restorer/restorer.go
ishan16696: parent context was there as I need to create context.
And multiple context are necessary according to the API calls like calling compaction might take time, hence context with bigger timeout.
Calling status API call shouldn't take more time, hence context with less timeout.

ishan16696:
> Additionally, I'm curious about the rationale behind using a separate goroutine, go r.HandleAlarm(...), specifically for triggering defragmentation. It's not immediately clear why this is necessary, and understanding your reasoning here would be helpful.

I think I have already explain this in this comment: https://github.com/gardener/etcd-backup-restore/pull/668#issuecomment-1818283375 , can you please go through this comment and let me know, if you get it

ishan16696:
> However, I suggest a different approach. Instead of relying on a fixed count of delta snapshots, it's more effective to monitor their cumulative size. We should initiate compaction and defragmentation when they collectively reach the 1 GB threshold. This is when the MakeEtcdLean function would be most effective.

I'm not getting your approach, what is cumulative size ? Is it deltaSnapshotSize + dbSize ? How does it make a difference ?
There is no need to check dbsize on every deltaSnapshot. There are 2 reasons to that:

Checking the `dbsize`, you can make unncessary API call to embedded etcd, let's say you are apply `5k` delta snapshots, you end up making `5k` API calls just to check dbsize.
With my approach you will only make 5k/10~=500 API calls. Isn't this advantageous

pkg/snapshot/restorer/restorer.go
seshachalam-yv: >please check the whole code, when anonymous function returns error we log that error send the dbSizeDisAlarmCh <- false to channel to indicate it fails

You are referring `HandleAlarm` but here it's not necessary to anonymous function.

if you feel, immediately you want to cancel context in above pointed code, then do it consistently for all the context in this function.

pkg/snapshot/restorer/restorer.go
```diff
@@ -57,11 +57,11 @@ import (
...
 const (
-	etcdConnectionTimeout                     = 30 * time.Second
-	etcdCompactTimeout                        = 2 * time.Minute
-	etcdDefragTimeout                         = 5 * time.Minute
-	periodicallyMakeEtcdLean                  = 10
-	thresholdPercentageForDBSizeAlarm float64 = 80.0 / 100.0
+	etcdConnectionTimeout                                 = 30 * time.Second
+	etcdCompactTimeout                                    = 2 * time.Minute
+	etcdDefragTimeout                                     = 5 * time.Minute
+	periodicallyMakeEtcdLeanAfterDeltaSnapApplied         = 10
```
shreyas-s-rao, line 63:
```suggestion
	periodicallyMakeEtcdLeanDeltaSnapshotInterval         = 10
```

shreyas-s-rao (changes requested):

pkg/snapshot/restorer/restorer.go
ishan16696: done

shreyas-s-rao (approved):
/lgtm

pkg/snapshot/restorer/restorer.go
ishan16696: > if you feel, immediately you want to cancel context in above pointed code, then do it consistently for all the context in this function.

ideally it should be but it depends on context timeout like status call have context.Timeout every small as compared to compaction and defragmentation.

ishan16696: If you want then I can make these changes.

ishan16696: ok, If you aren't satisfied, then I will make the changes as you requested. I hope that's fine.
But that shouldn't be bigger issue as we discussed both ways are doing similar thing.

ishan16696: I have removed the anonymous function from compaction. PTAL

seshachalam-yv (approved):
/lgtm

Issue #354, opened 2021-07-05
[Feature]To take out of sync full/delta snapshot,  delegate the incoming http request from backup Follower to the backup Leader.


ishan16696:
**Feature (What you would like to be added):**
If backup-restore `Follower` receives the HTTP/HTTPS  requests to take [out-of-sync full/delta snapshot](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L103-L105) then delegate that incoming request to the backup-restore `Leader` and Let the backup-restore leader take/upload the out-of-sync snapshot(full as well as incremental).

**Motivation (Why is this needed?):**
While adding support to bring the [snapshotters](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#backup) (both full and incremental) go routines under the control of the leader-election as described [here](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#work-flows-in-etcd-backup-restore), We observed that all backup-restore Follower's [http handler](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L93-L105) don't have updated [SnapshotterState](https://github.com/gardener/etcd-backup-restore/blob/13ddcaa7ec25be4b23f204124ac3f718fe9920c9/pkg/snapshot/snapshotter/snapshotter.go#L68-L92), if the request comes to the REST endpoints fo `follower` to take out of sync full/delta snapshot then the snapshot taken will not be the latest, and it is not good to let backup-restore `Follower` take the snapshot, it should delegate the request to backup-restore `Leader`.

**Approach/Hint to the implement solution (optional):**

PR #675, opened 2023-11-01
Queries the actual store to fetch latest snapshot while serving the URL endpoint.

abdasgupta:
Currently, if someone queries `/snapshot/latest` endpoint in backup-restore server,  latest snapshots are served from whatever is stored in the already running snapshotter. But, ideally, the snapshots should be fetched from the actual backend store and then serve to the end users. This PR fetches the latest snapshots from actual store when an end user executes `/snapshot/latest` endpoint.

One use case of this PR is e2e tests for compaction. Full snapshots are stored in same storage by both the snapshotter and compactor. Currently, `/snapshot/latest` endpoint serves the latest snapshots from the details stored in snapshotter. While snapshotter has the information about what latest snapshots are pushed by itself, it does not have the information about the full snapshots pushed by compactor which runs separately. So, in e2e test for compaction, if we want to check the latest snapshots through this endpoint after a compaction job, this endpoint provides stale data about full snapshot as it only has the info about full snapshots pushed by the snapshotter.

Fixes partial https://github.com/gardener/etcd-druid/issues/704

**Release note**:

```improvement user
The snapshots are fetched from the actual backend store when queried for latest snapshots on `/snapshot/latest` endpoint.
```

### DIFF
--- a/pkg/server/httpAPI.go
+++ b/pkg/server/httpAPI.go
@@ -38,6 +38,7 @@ import (
 	"github.com/gardener/etcd-backup-restore/pkg/member"
 	"github.com/gardener/etcd-backup-restore/pkg/miscellaneous"
 	"github.com/gardener/etcd-backup-restore/pkg/snapshot/snapshotter"
+	"github.com/gardener/etcd-backup-restore/pkg/snapstore"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
 	"github.com/ghodss/yaml"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -372,9 +373,23 @@ func (h *HTTPHandler) serveLatestSnapshotMetadata(rw http.ResponseWriter, req *h
 		return
 	}
 
+	store, err := snapstore.GetSnapstore(h.SnapstoreConfig)
+	if err != nil {
+		h.Logger.Warnf("Unable to create snapstore from configured storage provider: %v", err)
+		rw.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	fullSnap, deltaSnaps, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(store)
+	if err != nil {
+		h.Logger.Warnf("Unable to fetch latest snapshots from snapstore: %v", err)
+		rw.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
 	resp := latestSnapshotMetadataResponse{
-		FullSnapshot:   h.Snapshotter.PrevFullSnapshot,
-		DeltaSnapshots: h.Snapshotter.PrevDeltaSnapshots,
+		FullSnapshot:   fullSnap,
+		DeltaSnapshots: deltaSnaps,
 	}
 
 	json, err := json.Marshal(resp)


ishan16696:
Hi @abdasgupta ,
Can you please provide the context for this PR, where you want to use this endpoint: `/snapshot/latest` ? 
As I have never seen any usecases of this `/snapshot/latest` endpoint.
I'm asking this because in multi-node context only backup-restore should respond to this REST request, please refer to this issue: https://github.com/gardener/etcd-backup-restore/issues/354 for details. I worked on this issue long back. I solved it for endpoint `/snapshot/delta` and `/snapshot/full`, as `/snapshot/latest` endpoint nobody used, may be I have left this endpoint.

ishan16696:
> I solved it for endpoint `/snapshot/delta` and `/snapshot/full`, as `/snapshot/latest` endpoint nobody used, may be I have left this endpoint.

ok, I found out this functionality is already there for endpoint: `/snapshot/latest` endpoint, sorry for the confusion
https://github.com/gardener/etcd-backup-restore/blob/ae1c95c8ba59d7134a3ac0b473b8f983716cfc3d/pkg/server/httpAPI.go#L364-L372

ishan16696:
>  But, ideally, the snapshots should be fetched from the actual backend store and then serve to the end users. This PR fetches the latest snapshots from actual store when an end user executes /snapshot/latest endpoint.

but why it should be fetched from snapstore as backup-restore leader already have this information as it is taking the snapshot, so why use that rather then go to snapstore and list the snapshots ? Have you seen any weird behaviour like when `/latest` doesn't give latest snapshot information ?

abdasgupta:
> > But, ideally, the snapshots should be fetched from the actual backend store and then serve to the end users. This PR fetches the latest snapshots from actual store when an end user executes /snapshot/latest endpoint.
> 
> but why it should be fetched from snapstore as backup-restore leader already have this information as it is taking the snapshot, so why not use that rather then go to snapstore and list the snapshots ? Have you seen any weird behaviour like when `/latest` doesn't give latest snapshot information ?

Yes. I have seen weird behaviour while developing the e2e tests for compaction. Ideally, a URL endpoint should always fetch data from actual backend, not from any cache or something. Full snapshots are stored in same storage by both the snapshotter and compactor. While snapshotter has the information about what latest snapshots are pushed by itself, it does not have the information about the full snapshots pushed by compactor which runs separately. So, in e2e test for compaction, if we want to check the latest snapshots  through this endpoint after a compaction job, this endpoint provides stale data about full snapshot as it only have the info about full snapshots pushed by the snapshotter.

ishan16696:
> While snapshotter has the information about what latest snapshots are pushed by itself, it does not have the information about the full snapshots pushed by compactor which runs separately. So, in e2e test for compaction, if we want to check the latest snapshots through this endpoint after a compaction job, this endpoint provides stale data about full snapshot as it only have the info about full snapshots pushed by the snapshotter.

thanks for clarifying, now it make sense. I think it should be mentioned in PR description as it seems important point to me.

ishan16696 (approved):
LGTM!!

abdasgupta:
I updated the PR description @ishan16696

PR #688, opened 2023-12-06
switch GCR -> Artifact-Registry

ccwienk:
GCR has been deprecated [0] in favour of Artifact-Registry.

Thus, change push-targets for OCI-Images:

- europe-docker.pkg.dev/gardener-project/snapshots for snapshots
- europe-docker.pkg.dev/gardener-project/releases for releases
- europe-docker.pkg.dev/gardener-project/public for combined view of snapshots + releases

[0]
https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr

**Release note**:

```breaking operator
Change OCI Image Registry from GCR (`eu.gcr.io/gardener-project`) to Artifact-Registry (`europe-docker.pkg.dev/gardener-project/releases`). Users should update their references.

```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -17,14 +17,13 @@ etcd-backup-restore:
       - --testrun-prefix=playground
       - --set=projectNamespace=garden-it
 
-  template: 'default'
   base_definition:
-    repo: ~
     traits:
       version:
         preprocess:
           'inject-commit-hash'
         inject_effective_version: true
+      component_descriptor: ~
       publish:
         oci-builder: docker-buildx
         platforms:
@@ -32,7 +31,6 @@ etcd-backup-restore:
         - linux/arm64
         dockerimages:
           etcdbrctl:
-            registry: 'gcr-readwrite'
             image: 'eu.gcr.io/gardener-project/gardener/etcdbrctl'
             dockerfile: 'build/Dockerfile'
             inputs:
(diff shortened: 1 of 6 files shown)


.ci/pipeline_definitions
```diff
@@ -55,23 +54,31 @@ etcd-backup-restore:
...
     release:
       traits:
         version:
           preprocess: 'finalize'
+        component_descriptor:
+          ocm_repository: europe-docker.pkg.dev/gardener-project/releases
+        publish:
+          dockerimages:
+            etcdbrctl:
+              image: europe-docker.pkg.dev/gardener-project/releases/gardener/etcdbrctl
```
shreyas-s-rao, line 81: Shouldn't this be `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl`? If not, then [how] are these images copied over? Or is the `public` registry some sort of mirror/proxy for the `releases` + `snapshots` registries?

shreyas-s-rao (commented):
@ccwienk thanks for adapting the image registry paths. I had one question though, about where the image actually gets pushed to. PTAL

.ci/pipeline_definitions
ccwienk: @shreyas-s-rao : `public` is a "virtual" (as GCP calls it) registry that offers a combined view for the union set of `releases` and `snapshots`. so every image in either registry will be visible in `public`, as well.

shreyas-s-rao (approved):
/lgtm

PR #691, opened 2023-12-13
Add support for overriding storage API endpoint for GCS

### LINKED ISSUE #690: [Feature] Add support for using GCS regional endpoints
Add support for using [GCS regional endpoints](https://cloud.google.com/storage/docs/regional-endpoints).

This is required to allow etcd-backup-restore to securely access backup buckets created within the same region, such as `me-central2` (Saudi Arabia), so that the etcd-backup-restore can make network calls to GCP within the same region, keeping the etcd data within the same region in-transit.

/area backup
/platform gcp

/cc @dguendisch

shreyas-s-rao:
/area backup
/platform gcp
/kind enhancement

Add support for overriding storage API endpoint for Google GCS provider. This allows users to specify their own storage API [endpoint override](https://pkg.go.dev/cloud.google.com/go#hdr-Endpoint_Override), such as a [regional endpoint](https://cloud.google.com/storage/docs/regional-endpoints), or even a local GCS emulator endpoint. Users need to set environment variable `GOOGLE_STORAGE_API_ENDPOINT`, with the value in the format `http[s]://host[:port]/storage/v1/`.

Fixes #690 

/cc @dguendisch @anveshreddy18 
/invite @ishan16696 

**Release note**:

```improvement user
Add support for overriding storage API endpoint for provider GCS, by setting environment variable `GOOGLE_STORAGE_API_ENDPOINT`, with the value in the format `http[s]://host[:port]/storage/v1/`. ⚠️ Note: GC storage API endpoint will not be overridden for `copy` subcommand, since backup buckets may reside in different regions.
```

### DIFF
--- a/pkg/snapstore/gcs_snapstore.go
+++ b/pkg/snapstore/gcs_snapstore.go
@@ -36,8 +36,10 @@ import (
 )
 
 const (
-	storeCredentials       = "GOOGLE_APPLICATION_CREDENTIALS"
-	sourceStoreCredentials = "SOURCE_GOOGLE_APPLICATION_CREDENTIALS"
+	storeCredentials         = "GOOGLE_APPLICATION_CREDENTIALS"
+	storageAPIEndpoint       = "GOOGLE_STORAGE_API_ENDPOINT"
+	sourceStoreCredentials   = "SOURCE_GOOGLE_APPLICATION_CREDENTIALS"
+	sourceStorageAPIEndpoint = "SOURCE_GOOGLE_STORAGE_API_ENDPOINT"
 )
 
 // GCSSnapStore is snapstore with GCS object store as backend.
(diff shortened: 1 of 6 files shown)


ishan16696:
Thanks for PR @shreyas-s-rao ,
Please update [this](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/04-google-cloud-storage-secret.yaml) secrets example file as well

ishan16696:
@shreyas-s-rao why we are updating the helm charts in this PR as we haven't maintain the charts since long time, you can refer this https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/values.yaml#L10C1-L10C1

shreyas-s-rao:
> why we are updating the helm charts in this PR as we haven't maintain the charts since long time, you can refer this https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/values.yaml#L10,
as it became hard to deploy multi-node through helm charts as charts don't involves etcd-druid.

@ishan16696 Agreed, it is difficult to maintain the helm charts here, but we can't fully ignore it, since there are many community users who use etcd-backup-restore without etcd-druid, unless we enforce it. We should spend some effort in updating these charts, and keep them updated when every new PR comes in, independent of druid.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -58,14 +59,23 @@ const (
...
 	ctx := context.TODO()
-	var opts []option.ClientOption
+	var opts []option.ClientOption // no need to explicitly set store credentials here since the Google SDK picks it up from the standard environment variable
+
+	if _, ok := os.LookupEnv(envSourceStoreCredentials); !ok { // do not set endpoint override when copying backups between buckets, since the buckets may reside on different regions
+		endpoint := strings.TrimSpace(os.Getenv(envStorageAPIEndpoint))
+		if endpoint != "" {
+			opts = append(opts, option.WithEndpoint(endpoint))
+		}
+	}
```
ishan16696, line 69:
```suggestion
                 if endpoint, isSet := os.LookupEnv(envStorageAPIEndpoint); isSet {
			opts = append(opts, option.WithEndpoint(strings.TrimSpace(endpoint)))
		}
	}
```

shreyas-s-rao: Does `isSet` guarantee that endpoint will not be empty or a bunch of whitespaces? Probably not. That's the reason I've explicitly checked for endpoint being empty after trimming whitespaces, so that we don't configure the google client with an invalid endpoint by accident.

shreyas-s-rao: os.LookupEnv():
```
// LookupEnv retrieves the value of the environment variable named
// by the key. If the variable is present in the environment the
// value (which may be empty) is returned and the boolean is true.
// Otherwise the returned value will be empty and the boolean will
// be false.
```

The env var can be present but still be empty, and `isSet` will be returned as `true`.

ishan16696: ok, I don't have strong opinion on this.

ishan16696:
Can you re-trigger pipeline build as I think pipeline has been fixed now ?

ishan16696 (approved):
LGTM!!

ishan16696:
Please rebase this PR on latest master @shreyas-s-rao

anveshreddy18 (approved):
Thanks @shreyas-s-rao for the PR. 
/lgtm

PR #697, opened 2024-01-01
[FakeGCS] Add a new field `emulatorEnabled` in gcp secret & configure Gcloud client for GCS Emulator

### LINKED ISSUE #696: [Enhancement] Add a new field `enableGCSEmulator` in gcp secret & configure Gcloud client for GCS Emulator

* Add a new field `enableGCSEmulator` in gcp secret
* configure Google cloud client for GCS Emulator when `enableGCSEmulator` is `True` ( case insensitive )

For [Etcd-druid](https://github.com/gardener/etcd-druid) to use GCS Emulator [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) in running e2e-tests locally, Google cloud client in `backup-restore` has to be configured to connect to the Emulator when `enableGCSEmulator` field in the gcp secret is set to `True`. 

anveshreddy18:

* Add a new field `emulatorEnabled` in gcp secret
* configure Google cloud client for GCS Emulator when `emulatorEnabled` is `true`

This is helpful to run `etcdbr` tests locally and on pipelines without actually using the real infra. This can also be used by [Etcd-druid](https://github.com/gardener/etcd-druid) to use GCS Emulator [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) for running e2e-tests.

For these, Google cloud client in `backup-restore` has to be configured to connect to the Emulator. 

This PR enables that in either of the two ways 
1) `GOOGLE_EMULATOR_ENABLED` env var is set when running `etcdbr` as a process using `etcdbrctl` 
2) `emulatorEnabled` field in the gcp secret is set to `true` via helm chart

Fixes #696 

**Release note**:

```improvement developer
Added support for using fake-gcs-server for all etcdbr functionalities. To enable: Either 
1) Set `GOOGLE_EMULATOR_ENABLED` environment variable when running `etcdbrctl` command OR 
2) Set `emulatorEnabled: true` in GCP backup secret when deploying via Helm chart.
```

### DIFF
--- a/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
@@ -365,3 +365,4 @@ spec:
       resources:
         requests:
           storage: {{ .Values.storageCapacity }}
+          
--- a/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
@@ -218,6 +218,13 @@ spec:
               key: "storageAPIEndpoint"
               optional: true
   {{- end }}
+  {{- if .Values.backup.gcs.emulatorEnabled }}
+        - name: "GOOGLE_EMULATOR_ENABLED"
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Release.Name }}-etcd-backup
+              key: "emulatorEnabled"
+  {{- end }}
 {{- else if eq .Values.backup.storageProvider "Swift" }}
         - name: "OS_AUTH_URL"
           valueFrom:
--- a/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
@@ -192,7 +192,7 @@ spec:
               name: {{ .Release.Name }}-etcd-backup
               key: "storageKey"
   {{- if .Values.backup.abs.emulatorEnabled }}
-        - name: "EMULATOR_ENABLED"
+        - name: "AZURE_EMULATOR_ENABLED"
           valueFrom:
             secretKeyRef:
               name: {{ .Release.Name }}-etcd-backup
@@ -224,6 +224,7 @@ spec:
             secretKeyRef:
               name: {{ .Release.Name }}-etcd-backup
               key: "emulatorEnabled"
+              optional: true
   {{- end }}
 {{- else if eq .Values.backup.storageProvider "Swift" }}
         - name: "OS_AUTH_URL"
--- a/example/storage-provider-secrets/04-google-cloud-storage-secret.yaml
+++ b/example/storage-provider-secrets/04-google-cloud-storage-secret.yaml
@@ -6,4 +6,5 @@ metadata:
 type: Opaque
 data:
   serviceaccount.json: ...
-#  storageAPIEndpoint: # http[s]://host[:port]/storage/v1/
+#  storageAPIEndpoint: aHR0cFtzXTovL2hvc3RbOnBvcnRdL3N0b3JhZ2UvdjEv # http[s]://host[:port]/storage/v1/
+#  emulatorEnabled: dHJ1ZQ== # true (optional)
--- a/pkg/snapstore/gcs_snapstore.go
+++ b/pkg/snapstore/gcs_snapstore.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -29,6 +30,7 @@ const (
 	envStoreCredentials       = "GOOGLE_APPLICATION_CREDENTIALS"
 	envStorageAPIEndpoint     = "GOOGLE_STORAGE_API_ENDPOINT"
 	envSourceStoreCredentials = "SOURCE_GOOGLE_APPLICATION_CREDENTIALS"
+	envEmulatorEnabled        = "GOOGLE_EMULATOR_ENABLED"
 )
 
 // GCSSnapStore is snapstore with GCS object store as backend.
@@ -40,6 +42,13 @@ type GCSSnapStore struct {
 	maxParallelChunkUploads uint
 	minChunkSize            int64
 	tempDir                 string
+	chunkDirSuffix          string
+}
+
+// gcsEmulatorConfig holds the configuration for the fake GCS emulator
+type gcsEmulatorConfig struct {
+	enabled  bool   // whether the fake GCS emulator is enabled
+	endpoint string // the endpoint of the fake GCS emulator
 }
 
 const (
@@ -50,16 +59,28 @@ const (
 // NewGCSSnapStore create new GCSSnapStore from shared configuration with specified bucket.
 func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
 	ctx := context.TODO()
+	var emulatorConfig gcsEmulatorConfig
+	emulatorConfig.enabled = isEmulatorEnabled()
 	var opts []option.ClientOption // no need to explicitly set store credentials here since the Google SDK picks it up from the standard environment variable
 
-	if _, ok := os.LookupEnv(envSourceStoreCredentials); !ok { // do not set endpoint override when copying backups between buckets, since the buckets may reside on different regions
+	if _, ok := os.LookupEnv(envSourceStoreCredentials); !ok || emulatorConfig.enabled { // do not set endpoint override when copying backups between buckets, since the buckets may reside on different regions
 		endpoint := strings.TrimSpace(os.Getenv(envStorageAPIEndpoint))
 		if endpoint != "" {
 			opts = append(opts, option.WithEndpoint(endpoint))
+			if emulatorConfig.enabled {
+				emulatorConfig.endpoint = endpoint
+			}
 		}
 	}
-
-	if config.IsSource {
+	var chunkDirSuffix string
+	if emulatorConfig.enabled {
+		err := emulatorConfig.configureClient(opts)
+		if err != nil {
+			return nil, err
+		}
+		chunkDirSuffix = brtypes.ChunkDirSuffix
+	}
+	if config.IsSource && !emulatorConfig.enabled {
 		filename := os.Getenv(envSourceStoreCredentials)
 		if filename == "" {
 			return nil, fmt.Errorf("environment variable %s is not set", envSourceStoreCredentials)
@@ -73,11 +94,11 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
 	}
 	gcsClient := stiface.AdaptClient(cli)
 
-	return NewGCSSnapStoreFromClient(config.Container, config.Prefix, config.TempDir, config.MaxParallelChunkUploads, config.MinChunkSize, gcsClient), nil
+	return NewGCSSnapStoreFromClient(config.Container, config.Prefix, config.TempDir, config.MaxParallelChunkUploads, config.MinChunkSize, chunkDirSuffix, gcsClient), nil
 }
 
 // NewGCSSnapStoreFromClient create new GCSSnapStore from shared configuration with specified bucket.
-func NewGCSSnapStoreFromClient(bucket, prefix, tempDir string, maxParallelChunkUploads uint, minChunkSize int64, cli stiface.Client) *GCSSnapStore {
+func NewGCSSnapStoreFromClient(bucket, prefix, tempDir string, maxParallelChunkUploads uint, minChunkSize int64, chunkDirSuffix string, cli stiface.Client) *GCSSnapStore {
 	return &GCSSnapStore{
 		prefix:                  prefix,
 		client:                  cli,
@@ -85,7 +106,31 @@ func NewGCSSnapStoreFromClient(bucket, prefix, tempDir string, maxParallelChunkU
 		maxParallelChunkUploads: maxParallelChunkUploads,
 		minChunkSize:            minChunkSize,
 		tempDir:                 tempDir,
+		chunkDirSuffix:          chunkDirSuffix,
 	}
+}
+
+// isEmulatorEnabled checks if the fake GCS emulator is enabled
+func isEmulatorEnabled() bool {
+	isFakeGCSEnabled, ok := os.LookupEnv(envEmulatorEnabled)
+	if !ok {
+		return false
+	}
+	emulatorEnabled, err := strconv.ParseBool(isFakeGCSEnabled)
+	if err != nil {
+		return false
+	}
+	return emulatorEnabled
+}
+
+// configureClient configures the fake gcs emulator
+func (e *gcsEmulatorConfig) configureClient(opts []option.ClientOption) error {
+	err := os.Setenv("STORAGE_EMULATOR_HOST", strings.TrimPrefix(e.endpoint, "http://"))
+	if err != nil {
+		return fmt.Errorf("failed to set the environment variable for the fake GCS emulator: %v", err)
+	}
+	opts = append(opts, option.WithoutAuthentication())
+	return nil
 }
 
 // Fetch should open reader for the snapshot file from store.
@@ -155,7 +200,7 @@ func (s *GCSSnapStore) Save(snap brtypes.Snapshot, rc io.ReadCloser) error {
 	var subObjects []stiface.ObjectHandle
 	prefix := adaptPrefix(&snap, s.prefix)
 	for partNumber := int64(1); partNumber <= noOfChunks; partNumber++ {
-		name := path.Join(prefix, snap.SnapDir, snap.SnapName, fmt.Sprintf("%010d", partNumber))
+		name := path.Join(prefix, snap.SnapDir, fmt.Sprintf("%s%s", snap.SnapName, s.chunkDirSuffix), fmt.Sprintf("%010d", partNumber))
 		obj := bh.Object(name)
 		subObjects = append(subObjects, obj)
 	}
@@ -184,7 +229,7 @@ func (s *GCSSnapStore) uploadComponent(snap *brtypes.Snapshot, file *os.File, of
 	sr := io.NewSectionReader(file, offset, size)
 	bh := s.client.Bucket(s.bucket)
 	partNumber := ((offset / chunkSize) + 1)
-	name := path.Join(adaptPrefix(snap, s.prefix), snap.SnapDir, snap.SnapName, fmt.Sprintf("%010d", partNumber))
+	name := path.Join(adaptPrefix(snap, s.prefix), snap.SnapDir, fmt.Sprintf("%s%s", snap.SnapName, s.chunkDirSuffix), fmt.Sprintf("%010d", partNumber))
 	obj := bh.Object(name)
 	ctx, cancel := context.WithTimeout(context.TODO(), chunkUploadTimeout)
 	defer cancel()
--- a/pkg/snapstore/gcs_snapstore.go
+++ b/pkg/snapstore/gcs_snapstore.go
@@ -111,7 +111,7 @@ func NewGCSSnapStoreFromClient(bucket, prefix, tempDir string, maxParallelChunkU
 
 // isEmulatorEnabled checks if the fake GCS emulator is enabled
 func isEmulatorEnabled() bool {
-	isFakeGCSEnabled, ok := os.LookupEnv(EnvGcsEmulatorEnabled)
+	isFakeGCSEnabled, ok := os.LookupEnv(EnvGCSEmulatorEnabled)
 	if !ok {
 		return false
 	}
--- a/pkg/snapstore/gcs_snapstore.go
+++ b/pkg/snapstore/gcs_snapstore.go
@@ -30,7 +30,6 @@ const (
 	envStoreCredentials       = "GOOGLE_APPLICATION_CREDENTIALS"
 	envStorageAPIEndpoint     = "GOOGLE_STORAGE_API_ENDPOINT"
 	envSourceStoreCredentials = "SOURCE_GOOGLE_APPLICATION_CREDENTIALS"
-	envEmulatorEnabled        = "GOOGLE_EMULATOR_ENABLED"
 )
 
 // GCSSnapStore is snapstore with GCS object store as backend.
@@ -112,7 +111,7 @@ func NewGCSSnapStoreFromClient(bucket, prefix, tempDir string, maxParallelChunkU
 
 // isEmulatorEnabled checks if the fake GCS emulator is enabled
 func isEmulatorEnabled() bool {
-	isFakeGCSEnabled, ok := os.LookupEnv(envEmulatorEnabled)
+	isFakeGCSEnabled, ok := os.LookupEnv(EnvGcsEmulatorEnabled)
 	if !ok {
 		return false
 	}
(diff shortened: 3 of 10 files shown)


pkg/snapstore/gcs_snapstore.go
```diff
@@ -39,6 +40,7 @@ const (
 	envStoreCredentials       = "GOOGLE_APPLICATION_CREDENTIALS"
 	envStorageAPIEndpoint     = "GOOGLE_STORAGE_API_ENDPOINT"
 	envSourceStoreCredentials = "SOURCE_GOOGLE_APPLICATION_CREDENTIALS"
+	envEnableGCSEmulator      = "GOOGLE_ENABLE_GCS_EMULATOR"
```
seshachalam-yv, line 43: Can we rename `GOOGLE_ENABLE_GCS_EMULATOR` to `EMULATOR_ENABLED`? It is more clear and effectively conveys whether the emulator is enabled or not.

example/storage-provider-secrets/04-google-cloud-storage-secret.yaml
```diff
@@ -7,3 +7,4 @@ type: Opaque
 data:
   serviceaccount.json: ...
 #  storageAPIEndpoint: # http[s]://host[:port]/storage/v1/
+#  enableGCSEmulator: # True (case-insensitive)
```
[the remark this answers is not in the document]
seshachalam-yv: Why case-insensitive ?

pkg/snapstore/gcs_snapstore.go
```diff
@@ -68,6 +71,24 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
 			opts = append(opts, option.WithEndpoint(endpoint))
 		}
 	}
+	chunkSuffix := "" // add '@chunk' suffix to chunks folder when using fake gcs storage
```
seshachalam-yv, line 74: Consider adding a detailed comment here explaining the purpose of chunkSuffix. For instance, you could reference the related GitHub issue or provide reasoning from the fake GCS.
```suggestion
	var chunkSuffix string
```

seshachalam-yv (changes requested):

pkg/snapstore/gcs_snapstore.go
anveshreddy18: Thanks a lot for the review @seshachalam-yv. I was using that because I use the same variable in druid [fakegcs PR](https://github.com/gardener/etcd-druid/pull/741/files#:~:text=EnvGoogleEnableGCSEmulator%20%3D%20%22GOOGLE_ENABLE_GCS_EMULATOR%22). And the reason why I used `GOOGLE_ENABLE_GCS_EMULATOR` in druid is because all other variables used for different providers use the similar convention of prefixing with provider-name. [Reference](https://github.com/gardener/etcd-druid/blob/7bd94c48b8f39346c79d45f0546c850be72c12ab/pkg/common/constants.go#L53). Due to these reasons, I think it makes sense to keep this. Also `EMULATOR_ENABLED` may sound like it's enabled by default as it uses the word `ENABLED`.  Wdyt?

example/storage-provider-secrets/04-google-cloud-storage-secret.yaml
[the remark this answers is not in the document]
anveshreddy18: > Why case-insensitive ?

As it gives the user more flexibility of writing it in any form he/she chooses and not necessarily restricting them to just use "TRUE" only. They can use "True", "true", "TRUE", etc.. And similarly for the "faLsE" also 

> To improve clarity, we can use emulatorEnabled, which clearly indicates whether the emulator is enabled or not.

I think this is similar to the other comment, I stated in the [reply](https://github.com/gardener/etcd-backup-restore/pull/697#discussion_r1448708918) why we use that, since these two are inter linked, changing one requires changing others. I'm open to change, this can be discussed in the other comment [here](https://github.com/gardener/etcd-backup-restore/pull/697#discussion_r1448325839)

pkg/snapstore/gcs_snapstore.go
anveshreddy18: Thanks. Updated this in the latest commit.

chart/etcd-backup-restore/templates/etcd-statefulset.yaml
```diff
@@ -200,6 +200,12 @@ spec:
             secretKeyRef:
               name: {{ .Release.Name }}-etcd-backup
               key: "storageAPIEndpoint"
```
renormalize, line 202: An `{{- end }}` statement which is supposed to end the `{{- if .Values.backup.gcs.storageAPIEndpoint }}` block is missing after line 202 - right after `key: "storageAPIEndpoint"` which is cause for the lint error.

Please add this to fix the lint error and thereby fix the build.

renormalize (changes requested):

chart/etcd-backup-restore/templates/etcd-statefulset.yaml
```diff
@@ -200,6 +200,12 @@ spec:
...
+  {{- if .Values.backup.gcs.enableGCSEmulator }}
```
renormalize, line 203: An `{{- end }}` statement which is supposed to end the `{{- if .Values.backup.gcs.storageAPIEndpoint }}` block is missing after line 202 - right after `key: "storageAPIEndpoint"` which is cause for the lint error.

Please add this to fix the lint error and thereby fix the build.

renormalize (changes requested):

chart/etcd-backup-restore/templates/etcd-statefulset.yaml
anveshreddy18: My bad. The build was showing the error after line 203 and I had no clue what went wrong there. 
Thanks a lot @renormalize for finding this.

example/storage-provider-secrets/04-google-cloud-storage-secret.yaml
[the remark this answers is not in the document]
anveshreddy18: For future context: The variable has been renamed to `emulatorEnabled` instead of GCS specific naming to make it uniform across multiple fake providers i.e Azure, etc.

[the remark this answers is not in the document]
renormalize: The codebase is split in naming convention for `bool` types: some variables use `XEnabled` and some variables use `EnableX`, where the latter is more common from what I could find. Would it not be clearer to name this common flag to `enableEmulator`? What do you think @seshachalam-yv @anveshreddy18?

pkg/snapstore/snapshot.go
```diff
@@ -134,7 +134,8 @@ func ParseSnapshot(snapPath string) (*brtypes.Snapshot, error) {
 			s.IsFinal = true
 		}
 	}
-	unixTime, err := strconv.ParseInt(timeWithSnapSuffix[0], 10, 64)
+	creationTimeAndChunkSuffix := strings.Split(timeWithSnapSuffix[0], "@")
```
seshachalam-yv, line 137: [Here](https://github.com/gardener/etcd-backup-restore/pull/697/files#diff-fdebef784a0c29069a2bcd6a93ca1602a51c9bc83ec50df693da585e52eed31cR74-R92), we're appending the `@chunk` suffix to address issue encountered with the emulator `fakeGCS`.

@anveshreddy18, I'm curious about the rationale behind searching for `@` to parse the creationTime. It seems more straightforward to utilize the entire suffix instead of just `@`. 
```go
creationTimeAndChunkSuffix := strings.Split(timeWithSnapSuffix[0], "@")
```
---

"@shreyas-s-rao, @ishan16696, what do you think about adding a `chunkSuffix` option to the snapstore interface, just like we do with `prefix`? This way, not just `gcssnapstore`, but all storage providers could use it if needed."

pkg/snapstore/gcs_snapstore.go
```diff
@@ -68,6 +71,27 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
+			if endpoint, ok = os.LookupEnv(envStorageAPIEndpoint); !ok {
+				return nil, fmt.Errorf("GCS Emulator host is not set")
+			}
+			endpoint = strings.TrimSpace(endpoint)
+			os.Setenv("STORAGE_EMULATOR_HOST", strings.TrimPrefix(endpoint, "http://"))
+			opts = append(opts, option.WithEndpoint(endpoint))
+			opts = append(opts, option.WithoutAuthentication())
+			chunkSuffix = "@chunk"
+		}
+	}
```
shreyas-s-rao, line 94: Please pull this code of checking whether a GCS emulator is being used, into a separate function, to keep the code flow as simple as possible. Additionally,

pkg/snapstore/snapshot.go
shreyas-s-rao: Rather, I would ask why do we need to use `@chunk` at all? Can't we simply use a `/` at the end to indicate that it's a "folder"? Although, I'm not aware of how fake-gcs forms URIs of the objects within the folder - whether it would end up having two `/`s, like `a/b//c`. Maybe this is something that you can explore @anveshreddy18

shreyas-s-rao (changes requested):
@anveshreddy18 thanks for the PR. I have couple of suggestions. PTAL

pkg/snapstore/snapshot.go
anveshreddy18: Thanks for the suggestion @shreyas-s-rao , but I'm not particular about any suffix, but we need some sort of suffix definitely, pls refer to this [issue](https://github.com/fsouza/fake-gcs-server/issues/1434) on fake-gcs repo that I created. 

So in the event of adding a suffix (it could be anything for now), the question by @seshachalam-yv was instead of using a pre-set value to differentiate the time and snapshot suffix with the `@` char, can we pass a suffix from the snapstore struct itself so that it can be used both while creating the snapshot with the new name & also while parsing it in the snapshot.go file. 

So for eg: we use some `x` as suffix, then both while creating it and parsing it, we should use the same whole suffix we wanted. Instead of using `@chunk` while creating and just `@` while parsing the snapshot name, he's suggesting that we use the whole word we used to identify the chunk while parsing also. And that whole word should come from a snapstore rather than having the word in individual files. Wdyt?

shreyas-s-rao: > can we pass a suffix from the snapstore struct itself so that it can be used both while creating the snapshot with the new name & also while parsing it in the snapshot.go file.

Ideally yes, it has to be done from the snapstore itself. If you plan to add suffix to the chunk files, then I am not in favour of using a new delimiter like `@`, when we can use a standard delimiter like `.`, which we already use for suffixing certain snapshot information. So the suffix would simply be `.chunk.<part-number>` possibly.

So in `ParseSnapshot()` (line 128), `timeWithSnapSuffix` can now have more to the suffix, which is `.chunk.<part-number>`, which indicates that the snapshot is a chunk object with chunk-part-number too. You will need to handle this along with the existing handling of compression suffix and final suffix.

Of course, you will also have to now form the snapshot name too according to this logic, only if fake GCS is being used.

@anveshreddy18 @seshachalam-yv @ishan16696 thoughts?

pkg/snapstore/gcs_snapstore.go
```diff
@@ -68,6 +71,27 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
+		}
+		if isEmulator {
+			var endpoint string
+			if endpoint, ok = os.LookupEnv(envStorageAPIEndpoint); !ok {
+				return nil, fmt.Errorf("GCS Emulator host is not set")
+			}
+			endpoint = strings.TrimSpace(endpoint)
+			os.Setenv("STORAGE_EMULATOR_HOST", strings.TrimPrefix(endpoint, "http://"))
+			opts = append(opts, option.WithEndpoint(endpoint))
+			opts = append(opts, option.WithoutAuthentication())
```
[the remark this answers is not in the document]
anveshreddy18: Addressed in the latest commit. PTAL @seshachalam-yv

pkg/snapstore/snapshot.go
anveshreddy18: Updated in the latest commit. I'm getting the suffix from `brtypes` instead of hardcoding it. PTAL

pkg/snapstore/gcs_snapstore.go
anveshreddy18: Done.

chart/etcd-backup-restore/templates/etcd-statefulset.yaml
```diff
@@ -337,4 +344,4 @@ spec:
       - "ReadWriteOnce"
       resources:
         requests:
-          storage: {{ .Values.storageCapacity }}
+          storage: {{ .Values.storageCapacity }}
```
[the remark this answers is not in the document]
anveshreddy18: That was not deleted intentionally, It got auto removed. I'll add this in the next commit.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -50,6 +52,12 @@ type GCSSnapStore struct {
 	maxParallelChunkUploads uint
 	minChunkSize            int64
 	tempDir                 string
+	chunkDirSuffix          string
+}
+
+type EmulatorConfig struct {
```
renormalize, line 58: Since `EmulatorConfig` is only used in `gcs_snapstore.go`, is there any reason to make this struct exported in `package snapstore`?

renormalize (commented):
Nice PR @anveshreddy18! Just a few nits which are Go related from my side.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -83,21 +97,51 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
 		tempDir:                 tempDir,
+		chunkDirSuffix:          chunkDirSuffix,
+	}
+}
+
+// NewEmulatorConfig returns a new EmulatorConfig instance
+func NewEmulatorConfig() *EmulatorConfig {
+	return &EmulatorConfig{
+		Enabled:  false,
+		endpoint: "",
```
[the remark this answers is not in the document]
anveshreddy18: Yeah true. Updated in the latest commit. PTAL

pkg/snapstore/gcs_snapstore.go
```diff
@@ -83,21 +97,51 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
+	fakegcsenabled, ok := os.LookupEnv(envEmulatorEnabled)
+	if !ok {
+		return false
+	}
+	emulatorEnabled, err := strconv.ParseBool(fakegcsenabled)
+	if err != nil {
+		return false
+	}
+	return emulatorEnabled
+}
```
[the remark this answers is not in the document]
anveshreddy18: Thanks for the suggestion on naming. 

But I would like to have this function return it and then assign it to `emulatorConfig.Enabled` as people glancing at the code know what to expect from this func and they see that the return value getting assigned to `emulatorConfig.Enabled`. It makes more sense as they don't have to get into the func to know that.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -50,6 +52,12 @@ type GCSSnapStore struct {
 	maxParallelChunkUploads uint
 	minChunkSize            int64
 	tempDir                 string
+	chunkDirSuffix          string
+}
+
+type EmulatorConfig struct {
+	Enabled  bool   // whether the fake GCS emulator is enabled
+	endpoint string // the endpoint of the fake GCS emulator
```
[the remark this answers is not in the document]
anveshreddy18: Thanks, I didn't realise that. Updated in the latest commit.

anveshreddy18:
@shreyas-s-rao, I've addressed your review comments. Can you PTAL. Thanks!

pkg/snapstore/gcs_snapstore.go
```diff
@@ -60,16 +69,22 @@ const (
...
 		if endpoint != "" {
 			opts = append(opts, option.WithEndpoint(endpoint))
+			emulatorConfig.endpoint = endpoint
 		}
 	}
-
-	if config.IsSource {
+	var chunkDirSuffix string
+	if emulatorConfig.enabled {
+		ConfigureGCSEmulator(&chunkDirSuffix, opts, emulatorConfig)
```
shreyas-s-rao, line 85: I would suggest to not mutate chunkDirSuffix in this function. Instead, the function can simply be a method on the `emulatorConfig` struct, and can be called `configureClient`. This method should simply take the `opts` and add the new `option.WithoutAuthentication()` and also call `os.SetEnv()`. Additionally, the chunkSuffixDir should be done separately within the `if emulatorConfig.enabled{}` block, because this is etcdbr-snapstore specific configuration.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -83,21 +98,43 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
+	}
+	emulatorEnabled, err := strconv.ParseBool(isFakeGCSEnabled)
+	if err != nil {
+		return false
+	}
+	return emulatorEnabled
+}
+
+// ConfigureGCSEmulator configures the fake GCS emulator
+func ConfigureGCSEmulator(chunkDirSuffix *string, opts []option.ClientOption, emulatorConfig gcsEmulatorConfig) {
```
shreyas-s-rao, line 131:
```suggestion
func (e *gcsEmulatorConfig) configureClient(opts []option.ClientOption) {
```
Avoid unnecessarily exposing helper functions.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -83,21 +98,43 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
+	endpoint := emulatorConfig.endpoint
+	os.Setenv("STORAGE_EMULATOR_HOST", strings.TrimPrefix(endpoint, "http://"))
+	opts = append(opts, option.WithoutAuthentication())
+	*chunkDirSuffix = brtypes.ChunkDirSuffix
```
shreyas-s-rao, line 135: As mentioned in previous comment, move this out of this function/method.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -83,21 +98,43 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
+	if err != nil {
+		return false
+	}
+	return emulatorEnabled
+}
+
+// ConfigureGCSEmulator configures the fake GCS emulator
+func ConfigureGCSEmulator(chunkDirSuffix *string, opts []option.ClientOption, emulatorConfig gcsEmulatorConfig) {
+	endpoint := emulatorConfig.endpoint
+	os.Setenv("STORAGE_EMULATOR_HOST", strings.TrimPrefix(endpoint, "http://"))
```
shreyas-s-rao, line 133:
```suggestion
	err := os.Setenv("STORAGE_EMULATOR_HOST", strings.TrimPrefix(emulatorConfig.endpoint, "http://"))
```
You can avoid creating an unnecessary variable here. Also, please handle the error returned.

pkg/snapstore/snapshot.go
```diff
@@ -126,6 +126,10 @@ func ParseSnapshot(snapPath string) (*brtypes.Snapshot, error) {
 	//parse creation time as well as parse the Snapshot compression suffix
 	lastNameToken := strings.Split(tokens[3], "/")
 	timeWithSnapSuffix := strings.Split(lastNameToken[0], ".")
+	// Check & remove if the last token is a chunk directory. The chunkDirSuffix is set by only GCS snapstore when using a emulator for testing.
+	if "."+timeWithSnapSuffix[len(timeWithSnapSuffix)-1] == brtypes.ChunkDirSuffix {
```
shreyas-s-rao, line 130: Please use fmt.Sprintf() to construct the LHS string, since that will be more readable.

shreyas-s-rao (changes requested):
@anveshreddy18 thanks for the PR. I have a few comments/suggestions. Please address them, thanks.

chart/etcd-backup-restore/templates/etcd-statefulset.yaml
```diff
@@ -201,6 +201,13 @@ spec:
               name: {{ .Release.Name }}-etcd-backup
               key: "storageAPIEndpoint"
   {{- end }}
+  {{- if .Values.backup.gcs.emulatorEnabled }}
+        - name: "EMULATOR_ENABLED"
```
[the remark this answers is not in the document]
anveshreddy18: Originally I prefixed it with `GOOGLE` like other params you mentioned. But I modified it because of the following [comments](https://github.com/gardener/etcd-backup-restore/pull/697#discussion_r1448325839). And later with this suggestion, Azurite PR got merged following the [same convention](https://github.com/gardener/etcd-backup-restore/blob/dc0f973b9db963b5a0042371ae3a125436f5801d/chart/etcd-backup-restore/templates/etcd-statefulset.yaml#L194-L199). That's why I have changed it as well to keep it same across the providers. WDYT?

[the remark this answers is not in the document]
shreyas-s-rao: Ouch. It would have been better to use a provider-specific environment variable name to be very precise about which provider the emulator has been enabled for.
I would request you to discuss with @renormalize and see whether you can use the provider-specific prefix for these, as is done with the other provider environment variables.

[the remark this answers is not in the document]
renormalize: The initial convention we both had followed was similar to the one you suggested, but we changed the convention after @seshachalam-yv's [comment](https://github.com/gardener/etcd-backup-restore/pull/697#discussion_r1448325839). 

I have no strong opinions on this. I'll change the convention followed once all reach an agreement.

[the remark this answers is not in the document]
anveshreddy18: Me & @shreyas-s-rao have come to agree upon the use of `PROVIDER_EMULATOR_ENABLED` format so that it is clear which emulator is being used for the job when we run etcdbr against multiple providers sequentially in pipelines, having a common env var causes ambiguities between deciding fake provider or actual provider across providers. 

All of this can be avoided by naming them with their prefix. I've made my changes in this respect. I would request @renormalize also to make similar change for the Azurite

shreyas-s-rao (approved):
@anveshreddy18 thanks for making the changes. 
/lgtm

Please rebase on latest master before merging.

shreyas-s-rao:
@anveshreddy18 please also modify the release note to talk about how to enable emulator for etcdbrctl command, rather than from the provider secret perspective. Something like 
```
Added support for using (fake-gcs-server)[https://github.com/fsouza/fake-gcs-server] for taking backups. Please set environment variable `GOOGLE_EMULATOR_ENABLED` when running `etcdbrctl` command to enable this support. Alternatively, you may also set the field `emulatorEnabled: true` in the GCP backup secret when deploying etcd-backup-restore via the helm chart to enable this.
```

This mentions to users of both `etcdbrctl` process as well as the helm charts, as to how to enable this feature.

seshachalam-yv (approved):
/lgtm

renormalize:
Thanks for also reflecting the changes in Azure with regards to naming convention, @anveshreddy18.
Testing will be significantly easier for GCS and Azure; now that emulators are supported for the three major providers.

Remove the unnecessary commit messages from the squash commit before merging for a cleaner commit message and history.

Thanks!

renormalize (approved):
Thanks for also reflecting the changes in Azure with regards to naming convention, @anveshreddy18.
Testing will be significantly easier for GCS and Azure; now that emulators are supported for the three major providers.

Remove the unnecessary commit messages from the squash commit before merging for a cleaner commit message and history.

Thanks!

PR #703, opened 2024-01-06
Fix regression in chunk deletion for openstack provider

### LINKED ISSUE #702: [BUG] Chunk objects get wrongly garbage collected when using openstack swift
**Describe the bug:**
Etcd-backup-restore uses multipart uploads for snapshots for all object storage providers, to improve snapshot upload efficiency as well as resilience to network issues. This process involves uploading individual parts, or "chunks", and finally finalizing/composing/completing/committing these individual parts into a single snapshot object on the storage bucket. While most providers recreate the final object from the individual objects on the server side, openstack swift simply expects a 0-byte "manifest object" to be uploaded, which stores the list of names of the individual "segment objects" (chunks) using the same object name as their prefix. Please refer to the [documentation](https://docs.openstack.org/swift/latest/overview_large_objects.html) for more details on this. Please note that in etcd-backup-restore, we use a DLO (dynamic large object), and not an SLO (static large object). The bucket should look something like this:

not-GCed

As a contrasting example, GCP's GCS object storage provider creates a new object from the individual chunks, by concatenating those chunks into a single object ([ref](https://cloud.google.com/storage/docs/composite-objects#composing)). This allows for the chunks to be garbage-collected once the object has been "composed", aka finalized.

Hence, etcdbr garbage collector needs to take special care not to delete the chunks for openstack swift provider. Recently a PR #673 was merged, which implemented chunk garbage collection for both GCP and openstack. While it was necessary for GCP, it was not required for openstack, due to the reasons mentioned above. This has caused a regression in copy subcommand, where the manifest object existed for certain snapshots, but the segments were gone due to GC, as can be seen from the below screenshot.

GCed

If chunks should not be garbage collected anymore, then who will clean up those chunks later? Regular deletion of the snapshot object would simply delete the manifest object, but not the segments/chunks. Since the object is a DLO, we cannot use the query parameter `multipart-manifest=delete` for the DELETE API call to swift, since that is allowed only for an SLO type object. Thus, we need to additionally take care of removing the segment objects as well when deleting the snapshot object.

**Expected behavior:**
In the case of openstack swift, chunks should be deleted only during deletion of the main snapshot object, and should never be garbage-collected by themselves.

- create openstack swift bucket
- run etcdbr (with garbage collection mode `LimitBased`) to create the following snapshots in this order: full, delta, delta, full
- wait for GC to run (use small GC period to let it run quicker): you should observe the segment/chunks for the 2 delta snapshots being deleted, but the manifest objects remain

/cc @plkokanov 

 - Etcd version/commit ID : 
 - Etcd-backup-restore version/commit ID: v0.27.0 and v0.24.7
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]: Swift

shreyas-s-rao:
This PR fixes a recent regression in chunk deletion behavior for Openstack Swift snapshots. Additionally, remove backward compatibility tests between `v1` and `v2` for garbage collection, since `v1` is no longer supported. Support for `v1` backup directory structure in other parts of the code should be gradually removed as well, but not in this PR, to avoid scope spill.

Fixes #702

/cc @plkokanov 
/invite @ishan16696 @anveshreddy18 @seshachalam-yv 

**Release note**:

```improvement operator
A regression in chunk deletion behavior for openstack provider has now been fixed.
```

### DIFF
--- a/pkg/snapstore/swift_snapstore.go
+++ b/pkg/snapstore/swift_snapstore.go
@@ -523,14 +523,13 @@ func (s *SwiftSnapStore) Delete(snap brtypes.Snapshot) error {
 			chunkObjectNames = append(chunkObjectNames, path.Join(chunk.Prefix, chunk.SnapDir, chunk.SnapName))
 		}
 
-		chunkObjectsDeleteResult := objects.BulkDelete(s.client, s.bucket, chunkObjectNames)
-		if chunkObjectsDeleteResult.Err != nil {
+		if chunkObjectsDeleteResult := objects.BulkDelete(s.client, s.bucket, chunkObjectNames); chunkObjectsDeleteResult.Err != nil {
 			return chunkObjectsDeleteResult.Err
 		}
 	}
 
-	manifestObjectDeleteResult := objects.Delete(s.client, s.bucket, path.Join(snap.Prefix, snap.SnapDir, snap.SnapName), nil)
-	return manifestObjectDeleteResult.Err
+	// delete manifest object
+	return objects.Delete(s.client, s.bucket, path.Join(snap.Prefix, snap.SnapDir, snap.SnapName), nil).Err
 }
 
 // GetSwiftCredentialsLastModifiedTime returns the latest modification timestamp of the Swift credential file(s)
--- a/pkg/snapstore/swift_snapstore.go
+++ b/pkg/snapstore/swift_snapstore.go
@@ -56,6 +56,10 @@ type SwiftSnapStore struct {
 	maxParallelChunkUploads uint
 	minChunkSize            int64
 	tempDir                 string
+	// deleteChunksForObject decides whether deletion of a (manifest) object should also delete the
+	// associated chunks (segment objects).
+	// Default: true
+	deleteChunksForObject bool
 }
 
 type applicationCredential struct {
@@ -110,10 +114,9 @@ func NewSwiftSnapStore(config *brtypes.SnapstoreConfig) (*SwiftSnapStore, error)
 	})
 	if err != nil {
 		return nil, err
-
 	}
 
-	return NewSwiftSnapstoreFromClient(config.Container, config.Prefix, config.TempDir, config.MaxParallelChunkUploads, config.MinChunkSize, client), nil
+	return NewSwiftSnapstoreFromClient(config.Container, config.Prefix, config.TempDir, config.MaxParallelChunkUploads, config.MinChunkSize, client, true), nil
 
 }
 
@@ -313,7 +316,7 @@ func readSwiftCredentialDir(dirName string) (*swiftCredentials, error) {
 }
 
 // NewSwiftSnapstoreFromClient will create the new Swift snapstore object from Swift client
-func NewSwiftSnapstoreFromClient(bucket, prefix, tempDir string, maxParallelChunkUploads uint, minChunkSize int64, cli *gophercloud.ServiceClient) *SwiftSnapStore {
+func NewSwiftSnapstoreFromClient(bucket, prefix, tempDir string, maxParallelChunkUploads uint, minChunkSize int64, cli *gophercloud.ServiceClient, deleteChunksForObject bool) *SwiftSnapStore {
 	return &SwiftSnapStore{
 		bucket:                  bucket,
 		prefix:                  prefix,
@@ -321,6 +324,7 @@ func NewSwiftSnapstoreFromClient(bucket, prefix, tempDir string, maxParallelChun
 		maxParallelChunkUploads: maxParallelChunkUploads,
 		minChunkSize:            minChunkSize,
 		tempDir:                 tempDir,
+		deleteChunksForObject:   deleteChunksForObject,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


ishan16696:
unit tests failed:
```
Summarizing 3 Failures:
  [FAIL] Save, List, Fetch, Delete from mock snapstore When both v1 and v2 are present [It] When both v1 and v2 are present
  /tmp/build/bc4dbee3/pull-request-gardener.etcd-backup-restore-master-pr/pkg/snapstore/snapstore_test.go:217
  [FAIL] Save, List, Fetch, Delete from mock snapstore When Only v2 is present [It] When Only v2 is present
  /tmp/build/bc4dbee3/pull-request-gardener.etcd-backup-restore-master-pr/pkg/snapstore/snapstore_test.go:267
  [FAIL] Save, List, Fetch, Delete from mock snapstore When Only v1 is present [It] When Only v1 is present
  /tmp/build/bc4dbee3/pull-request-gardener.etcd-backup-restore-master-pr/pkg/snapstore/snapstore_test.go:165
```

> Additionally, remove backward compatibility tests between v1 and v2 for garbage collection, since v1 is no longer supported. Support for v1 backup directory structure in other parts of the code should be gradually removed as well, but not in this PR, to avoid scope spill.

I would suggest let's address all of these in separate PR, let's keep this PR for bug fix, so that it become easy and quick to review. WDYT?

ishan16696:
unit tests also failed locally
```console
[AfterSuite]
/Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapstore/snapstore_suite_test.go:43
time="2024-01-08T14:53:39+05:30" level=info msg="Shutting down test server..."
[AfterSuite] PASSED [0.000 seconds]
------------------------------

Summarizing 3 Failures:
  [FAIL] Save, List, Fetch, Delete from mock snapstore When Only v1 is present [It] When Only v1 is present
  /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapstore/snapstore_test.go:165
  [FAIL] Save, List, Fetch, Delete from mock snapstore When both v1 and v2 are present [It] When both v1 and v2 are present
  /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapstore/snapstore_test.go:217
  [FAIL] Save, List, Fetch, Delete from mock snapstore When Only v2 is present [It] When Only v2 is present
  /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapstore/snapstore_test.go:267

Ran 115 of 115 Specs in 3.519 seconds
FAIL! -- 112 Passed | 3 Failed | 0 Pending | 0 Skipped
--- FAIL: TestSnapstore (3.52s)
FAIL

Ginkgo ran 1 suite in 8.47315375s

Test Suite Failed
```
I would suggest you can fix them in a separate PR.

pkg/snapstore/swift_snapstore.go
```diff
@@ -483,13 +484,47 @@ func (s *SwiftSnapStore) List() (brtypes.SnapList, error) {
...
+
+	var chunkObjectNames []string
+	for _, chunk := range chunks {
+		chunkObjectNames = append(chunkObjectNames, path.Join(chunk.Prefix, chunk.SnapDir, chunk.SnapName))
+	}
+
+	manifestObjectDeleteResult := objects.Delete(s.client, s.bucket, path.Join(snap.Prefix, snap.SnapDir, snap.SnapName), nil)
+	if manifestObjectDeleteResult.Err != nil {
+		return manifestObjectDeleteResult.Err
+	}
```
ishan16696, line 524:
```suggestion
	if manifestObjectDeleteResult := objects.Delete(s.client, s.bucket, path.Join(snap.Prefix, snap.SnapDir, snap.SnapName), nil); manifestObjectDeleteResult.Err != nil { 
	return manifestObjectDeleteResult.Err
	}

```

pkg/snapstore/swift_snapstore.go
```diff
@@ -483,13 +484,47 @@ func (s *SwiftSnapStore) List() (brtypes.SnapList, error) {
...
+
+	chunkObjectsDeleteResult := objects.BulkDelete(s.client, s.bucket, chunkObjectNames)
+	return chunkObjectsDeleteResult.Err
```
ishan16696, line 527:
```suggestion
return objects.BulkDelete(s.client, s.bucket, chunkObjectNames).Err
```

ishan16696 (changes requested):
few nits

pkg/snapstore/swift_snapstore.go
anveshreddy18, line 524: What if after the manifest object is deleted without any problem but then it gets the error deleting the chunk object in bulk in rare network issues. These chunks will be in the store forever because their manifest is deleted, so we won't get to this place ever again to delete those orphan chunks.

So, for deleting chunks like these, can we allow them to be deleted in the GC.

Or instead of dealing with the chunks here, can we delete all the chunks from GC with existing logic if they are out of our threshold holding time i.e 4 weeks. Then we can reduce the workload on snapshotter and make all the work through GC. Just a suggestion because when I raised the PR, I was trying to minimise workload on snapshotter.

shreyas-s-rao:
@ishan16696 thanks for your review. Addressed them. PTAL

pkg/snapstore/swift_snapstore.go
shreyas-s-rao: @anveshreddy18 this is a good point. Based on our out-of-band discussion yesterday, I have now changed the order of deletion of the objects to solve this problem. The chunks will be deleted first, followed by the manifest object. This way, if there is an error in deletion of the chunks, then the `Delete()` operation can be retried in the next run of garbage collector, since the manifest object still exists (the `List()` call simply lists all non-chunk objects, aka the manifest objects, which are considered to be the snapshot objects by any user of snapstore). This will allow for re-attempting of deletion of the remaining chunk objects, as returned by `getSnapshotChunks()` function. 

If chunk object deletion succeeds, but manifest object deletion fails, then the same logic holds - the Delete() operation will be retried in the next iteration of GC, which will then re-attempt to delete the manifest object.

This way, we don't need to have any additional handling of openstack chunk objects in garbage collector. This way the chunks are deleted along with the manifest object, and not retained for 4 weeks. This helps with cost savings, since buckets are charged as per the amount of data stored in it doe a given duration, like GB/month.

anveshreddy18: Thanks Shreyas for addressing my concern
> I have now changed the order of deletion of the objects to solve this problem.

Great! This will work. But I just have a small doubt.
Consider the scenario: 

- After all the chunks are deleted ( could be in multiple cycles of GC )
- Manifest object failed to delete for some reason in the latest cycle of GC
- In the next GC cycle, when it attempts to delete the chunks again, it gets empty list and hence `chunkObjectNames` is an empty slice of strings. 

Now I'm not sure what happens when the line `chunkObjectsDeleteResult := objects.BulkDelete(s.client, s.bucket, chunkObjectNames)` is executed as `chunkObjectNames` list is empty, I'm hoping `chunkObjectsDeleteResult.Err` would be nil. I tried searching for it, but couldn't find much documentation on that.  PTAL about this once.  If you are confident that there won't be any error from your knowledge of dealing with cloud providers, then all good.

shreyas-s-rao: Good point @anveshreddy18 . I'll add a check for the length of `chunks` returned, and call BulkDelete() based on that.

shreyas-s-rao: Done

anveshreddy18 (approved):
Thanks for the PR @shreyas-s-rao.
/lgtm

ishan16696 (approved):
Thanks @shreyas-s-rao for quickly fixing this issue.
LGTM!!

ishan16696:
Please cherry-pick to branch `rel-v0.24`

PR #710, opened 2024-01-30
Improve `etcd-backup-restore`'s  behavior with OpenStack Swift's Dynamic Large Objects for better error handling, and improved unit tests

### LINKED ISSUE #705: [Enhancement] Unit tests for OpenStack Swift `snapstore` should mimic the behavior of Dynamic Large Objects
Unit tests for openstack swift snapstore should consider Dynamic Large Objects

As mentioned in #702 , etcd-backup-restore uploads snapshots to openstack swift buckets as [DLOs (Dynamic Large Objects)](https://docs.openstack.org/swift/latest/overview_large_objects.html#dynamic-large-objects). The handling of chunks (segments) for such objects is different from multi-part uploaded objects in other storage providers, since chunks can be deleted without affecting the snapshot object, and vice versa. Whereas, in openstack swift, without the chunks, the manifest object is rendered useless, and restoration/copy operations would fail if a snapshot is missing its chunks.

While #703 fixes a regression in the handling of these chunks, the unit tests need to be adapted to consider such DLOs. The PR simply introduces a `deleteChunksForObject` bool field in the swift snapstore struct to circumvent the handling of DLOs in the unit tests, but this is not enough. The unit tests need to mock the actual behavior of the real snapstore, as closely as possible.

The unit tests use a mock swift client, which responds to different requests like GET, PUT, DELETE, etc. These handlers for these operations operate on a simple map called `objectsMap`, which stores the object names and their contents. The following changes need to be made to the handlers:
- A `snapstore.Save()` (PUT op) operation would simply write one single object to this `objectsMap`, whereas it is supposed to write to objects:
  1. the segment object, suffixed with `/00000001`, whose contents is the actual snapshot contents
  2. the manifest object, with the actual snapshot name, with contents being empty, since the manifest object is a 0-byte object

- The `snapstore.Delete()` (DELETE op) should now delete both the segment object as well as the manifest object from the `objectsMap`.

- The `snapstore.Fetch()` (GET op) on a snapshot should return a snapshot object that contains the content from the associated segment object, but named after the snapshot (manifest) object.

renormalize:

OpenStack Swift's Dynamic Large Objects' behavior was not considered properly till now, and this pull request improves `etcd-backup-restore`'ss behavior, by improving error handling when objects are deleted from the object store, and better unit tests which consider this behavior - to avoid potential future regressions in behavior like in #702.

******

**Improving error handling**

As stated in [bulk-delete](https://docs.openstack.org/swift/latest/api/bulk-delete.html) docs for Swift, the errors that occur during deletion server side in Swift, are a part of the HTTP response `Body`, and this content has to be checked for the errors that have occurred. The error handling that involves reading the data from the response `Body` is added.

******

**Improving unit tests**

The mock Swift client is a server that is defined in `swift_snapstore_test.go`, where each relevant `SwiftSnapStore` method like `List()`, `Fetch()`, `Save()`, `Delete()`'s invocation's response handler for the unit tests is defined. These handlers act directly on the `objectMap` `map` which is used to mock the behavior of an object store.

The following changes were made to the handlers to replicate the behavior of Dynamic Large Objects as closely as possible:

- `Save()` (`PUT` operation): is handled by `handleCreateTextObject()`. In `Save()`, each segment object is uploaded by `chunkUploader()` and on the successful upload of all the segments, the manifest object is uploaded. The handler `handleCreateTextObject()` simply adds the uploaded object to the `objectMap` which holds the contents of the snapshot. The handler behaves correctly when `chunkUploader()` is invoked to upload the segment objects, but it does not add the manifest object to the `objectMap` when the manifest object is uploaded in `Save()`. This is corrected by adding a key to the `objectMap` when the manifest object is uploaded.

- `Fetch()` (`GET` operation): is handled by `handleDownloadObject()`, and it originally just iterated over the keys of the `objectMap`, and combined all contents of keys which had the queried object as the prefix. Since the `range` operation of a `map` is not in sorted order, the keys which have the queried object as the prefix (segment objects) are first added to a list, and the contents are appended in sorted order to reconstruct the stored object.

- `Delete()` (`PUT` and `DELETE` operations):
  - `PUT` operation: To delete all the segment objects that correspond to a single manifest object, `Delete()` invokes `objects.BulkDelete()`. A new handler `handleBulkDeleteObject()` is registered with `initializeMockSwiftServer()` which responds to `PUT` requests with the `bulk-delete=true` query parameter, to delete multiple objects at once. `handleBulkDeleteObject()` deletes all segment objects that correspond to the queried object. See [bulk-delete](https://docs.openstack.org/swift/latest/api/bulk-delete.html) for more information.
  - `DELETE` operation: To delete the manifest object, `Delete()` invokes `objects.Delete()`. This call is handled by `handleDeleteObject()` which deletes the manifest object, and no changes are made as its behavior is as intended.

- `List()` (`GET` operation): no changes were necessary in `handleListObjectNames()` since it already behaved as intended.

Since the unit tests now function as intended, the field `deleteChunksForObject` in the `SwiftSnapStore` `struct` which was introduced in a patch release to circumvent the tests, is now removed.

******

Fixes #705 

**Release note**:

```improvement developer|operator
Improved error handling for OpenStack Swift during deletion of objects.
```

### DIFF
--- a/pkg/snapstore/snapshot.go
+++ b/pkg/snapstore/snapshot.go
@@ -65,7 +65,7 @@ func ParseSnapshot(snapPath string) (*brtypes.Snapshot, error) {
 	}
 
 	lastIndex = lastIndex + 3
-	prefix := snapPath[:lastIndex-1]
+	prefix := snapPath[:lastIndex]
 	snapPath = snapPath[lastIndex:]
 
 	tok := strings.Split(snapPath, "/")
--- a/pkg/snapstore/snapshot.go
+++ b/pkg/snapstore/snapshot.go
@@ -65,7 +65,7 @@ func ParseSnapshot(snapPath string) (*brtypes.Snapshot, error) {
 	}
 
 	lastIndex = lastIndex + 3
-	prefix := snapPath[:lastIndex]
+	prefix := snapPath[:lastIndex-1]
 	snapPath = snapPath[lastIndex:]
 
 	tok := strings.Split(snapPath, "/")
--- a/pkg/snapstore/snapstore_test.go
+++ b/pkg/snapstore/snapstore_test.go
@@ -42,20 +42,25 @@ var (
 
 var _ = Describe("Save, List, Fetch, Delete from mock snapstore", func() {
 	var (
-		snap1           brtypes.Snapshot
-		snap2           brtypes.Snapshot
-		snap3           brtypes.Snapshot
-		snap4           brtypes.Snapshot
-		snap5           brtypes.Snapshot
-		expectedVal1    []byte
-		expectedVal1_02 []byte
-		expectedVal2    []byte
-		expectedVal2_02 []byte
-		expectedVal4    []byte
-		expectedVal5    []byte
-		expectedVal4_02 []byte
-		expectedVal5_02 []byte
-		snapstores      map[string]brtypes.SnapStore
+		snap1        brtypes.Snapshot
+		snap2        brtypes.Snapshot
+		snap3        brtypes.Snapshot
+		snap4        brtypes.Snapshot
+		snap5        brtypes.Snapshot
+		expectedVal1 []byte
+		expectedVal2 []byte
+		expectedVal4 []byte
+		expectedVal5 []byte
+		// segments for Swift
+		expectedVal1Chunk01 []byte
+		expectedVal1Chunk02 []byte
+		expectedVal2Chunk01 []byte
+		expectedVal2Chunk02 []byte
+		expectedVal4Chunk01 []byte
+		expectedVal4Chunk02 []byte
+		expectedVal5Chunk01 []byte
+		expectedVal5Chunk02 []byte
+		snapstores          map[string]brtypes.SnapStore
 	)
 
 	BeforeEach(func() {
@@ -107,15 +112,19 @@ var _ = Describe("Save, List, Fetch, Delete from mock snapstore", func() {
 		expectedVal1 = []byte("value1")
 		expectedVal2 = []byte("value2")
 		// segments for Swift
-		expectedVal1_02 = []byte("value1_02")
-		expectedVal2_02 = []byte("value2_02")
+		expectedVal1Chunk01 = []byte("value1Chunk01")
+		expectedVal1Chunk02 = []byte("value1Chunk02")
+		expectedVal2Chunk01 = []byte("value2Chunk01")
+		expectedVal2Chunk02 = []byte("value2Chunk02")
 
 		// prefixv2
 		expectedVal4 = []byte("value4")
 		expectedVal5 = []byte("value5")
 		// segments for Swift
-		expectedVal4_02 = []byte("value4_02")
-		expectedVal5_02 = []byte("value5_02")
+		expectedVal4Chunk01 = []byte("value4Chunk01")
+		expectedVal4Chunk02 = []byte("value4Chunk02")
+		expectedVal5Chunk01 = []byte("value5Chunk01")
+		expectedVal5Chunk02 = []byte("value5Chunk02")
 
 		snapstores = map[string]brtypes.SnapStore{
 			"s3": NewS3FromClient(bucket, prefixV2, "/tmp", 5, brtypes.MinChunkSize, &mockS3Client{
@@ -156,17 +165,18 @@ var _ = Describe("Save, List, Fetch, Delete from mock snapstore", func() {
 			for provider, snapStore := range snapstores {
 				// Create store for mock tests
 				resetObjectMap()
-				objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName)] = &expectedVal1
-				objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName)] = &expectedVal2
 				if provider == "swift" {
 					// segment objects
-					objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName, "0000000001")] = &expectedVal1
-					objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName, "0000000002")] = &expectedVal1_02
-					objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName, "0000000001")] = &expectedVal2
-					objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName, "0000000002")] = &expectedVal2_02
+					objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName, "0000000001")] = &expectedVal1Chunk01
+					objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName, "0000000002")] = &expectedVal1Chunk02
+					objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName, "0000000001")] = &expectedVal2Chunk01
+					objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName, "0000000002")] = &expectedVal2Chunk02
 					// manifest objects
 					objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName)] = &[]byte{}
 					objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName)] = &[]byte{}
+				} else {
+					objectMap[path.Join(prefixV1, snap1.SnapDir, snap1.SnapName)] = &expectedVal1
+					objectMap[path.Join(prefixV1, snap2.SnapDir, snap2.SnapName)] = &expectedVal2
 				}
 
 				logrus.Infof("Running mock tests for %s when only v1 is present", provider)
@@ -174,7 +184,6 @@ var _ = Describe("Save, List, Fetch, Delete from mock snapstore", func() {
 				// List snap1 and snap2
 				snapList, err := snapStore.List()
 				Expect(err).ShouldNot(HaveOccurred())
-
 				if provider == "swift" {
 					Expect(snapList.Len()).To(Equal(6))
 					Expect(snapList[0].SnapName).To(Equal(snap2.SnapName))
@@ -191,7 +200,7 @@ var _ = Describe("Save, List, Fetch, Delete from mock snapstore", func() {
 					buf := new(bytes.Buffer)
 					_, err = io.Copy(buf, rc)
 					Expect(err).ShouldNot(HaveOccurred())
-					combinedExpectedValue := append(expectedVal1, expectedVal1_02...)
+					combinedExpectedValue := append(expectedVal1Chunk01, expectedVal1Chunk02...)
 					Expect(buf.Bytes()).To(Equal(combinedExpectedValue))
 				} else {
 					rc, err := snapStore.Fetch(*snapList[1])
@@ -213,6 +222,7 @@ var _ = Describe("Save, List, Fetch, Delete from mock snapstore", func() {
 				// reset the objectMap
 				resetObjectMap()
 				dummyData := make([]byte, 6*1024*1024)
+
 				// Save a new snapshot 'snap3'
 				err = snapStore.Save(snap3, io.NopCloser(bytes.NewReader(dummyData)))
 				Expect(err).ShouldNot(HaveOccurred())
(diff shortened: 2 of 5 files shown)


pkg/snapstore/swift_snapstore_test.go
```diff
@@ -150,3 +178,50 @@ func handleDeleteObject(w http.ResponseWriter, r *http.Request) {
...
+	segmentObjects := strings.Split(string(buf.Bytes()), "\n")
+	// last string is an empty string, artefact of '\n' at the end
+	segmentObjects = segmentObjects[:len(segmentObjects)-1]
+	for _, segmentObject := range segmentObjects {
+		segmentObject = "/" + segmentObject
+		// objects.BulkDelete() internally calls url.QueryEscape
+		unescapedQueryForObject, err := url.QueryUnescape(segmentObject)
+		if err != nil {
+			errorMessage := fmt.Sprintf("failed to unescape url %v", err)
+			logrus.Errorf(errorMessage)
```
shreyas-s-rao, line 208:
```suggestion
			logrus.Errorf(fmt.Sprintf("failed to unescape url %v", err))
```
Saves a line. Please follow this convention for other occurrences as well.

pkg/snapstore/swift_snapstore_test.go
```diff
@@ -30,6 +32,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type bulkDeleteJSONResponse struct {
+	Deleted      int      `json:"deleted"`
+	NotFound     int      `json:"not_found"`
+	ErrorStrings []string `json:"errors,omitempty"`
+}
+
```
shreyas-s-rao, line 40: Is this something you have come with by yourself? The standard [Result struct](https://github.com/gardener/etcd-backup-restore/blob/4f1f191f886124d9c45e8606255c0e00f7b220d9/vendor/github.com/gophercloud/gophercloud/results.go#L14-L39) used by openstack should be more than sufficient, right?

It gives you a `Body`, `Header` and `Err`. `Err` can be formed using errors.Join(), which takes an array/slice of error objects. And if we really need to record notFound and Deleted counts, we can directly write it to the response Body. Unless this struct is absolutely necessary, please remove it, to avoid reinventing the wheel.

pkg/snapstore/swift_snapstore_test.go
```diff
@@ -150,3 +178,50 @@ func handleDeleteObject(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			errorMessage := fmt.Sprintf("Resource not found: %s", segmentObject)
+			logrus.Errorf(errorMessage)
+			jsonResponse.NotFound++
+			jsonResponse.ErrorStrings = append(jsonResponse.ErrorStrings, errorMessage)
+		}
+	}
+
+	marshalledResponse, _ := json.Marshal(jsonResponse)
+	w.WriteHeader(http.StatusOK)
```
shreyas-s-rao, line 225: Use an array of errors called `errs` and append each collected error to that. At the end, if there are any errors present, then write header with an error response code, based on what kind of error (notFound, internal error, etc). Additionally, set response `Err` to errors.Join(errs), which allows you to join all the collected errors into a single error object.

The BulkDelete we use does not consume the response Body, so it may not make much sense to write anything to it, but rather just log the results like `total segments`, `found`, `not found`, `deleted`, etc

pkg/snapstore/snapstore_test.go
```diff
@@ -487,14 +546,42 @@ func modifyCredentialFileInDirectory(credentialDirectory, credentialFile string)
 	return os.WriteFile(credentialFilePath, []byte("MODIFIED CONTENT"), os.ModePerm)
 }
 
+// generates a unique string that would be the contents of a snapshot
+func generateContentsForSnapshot(snapshot *brtypes.Snapshot) string {
```
shreyas-s-rao, line 550: Why do we need this, when we can write the actual snapshot contents to the snapshot object, like was done previously?

shreyas-s-rao (changes requested):
@renormalize thanks for this much needed PR. I have a few suggestions and questions. PTAL, thanks.

pkg/snapstore/swift_snapstore_test.go
shreyas-s-rao: Oops, I didn't check https://docs.openstack.org/swift/latest/api/bulk-delete.html, apologies for that. I have seen that now, and wonder whether we need to correctly handle the errors from the response `Body` [here](https://github.com/gardener/etcd-backup-restore/blob/1a0dcf14220c452d86d2a8c6b65c26946d7475b3/pkg/snapstore/swift_snapstore.go#L521-L523), rather than response `Err`.

pkg/snapstore/swift_snapstore_test.go
```diff
@@ -102,13 +123,20 @@ func handleDownloadObject(w http.ResponseWriter, r *http.Request) {
...
+	for key := range objectMap {
 		if strings.HasPrefix(key, prefix) {
-			data := *val
-			contents = append(contents, data...)
+			sortedKeys = append(sortedKeys, key)
 		}
 	}
 
+	// append segment objects in order
+	sort.Strings(sortedKeys)
```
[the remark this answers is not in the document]
renormalize: Initially, I had used `slices.Sort()` as suggested. However, `slices.Sort()` was introduced in Go 1.21, while we are still on Go 1.20 - which caused the build to fail  ([3c0a58e](https://github.com/gardener/etcd-backup-restore/pull/710/commits/3c0a58e1a7db20df985d3a11d1ff2fd64f90b19b)). I reverted to `sort.Strings()` after this.

pkg/snapstore/swift_snapstore_test.go
renormalize: Understood, will changes all occurrences to follow this.

pkg/snapstore/snapstore_test.go
renormalize: The number of `expectedValX` variables would be a lot more since we need to create and hold the data that would be stored in the segment objects for Swift as well, and for each additional snapshot that is added to the tests in the future, an extra 2 `expectedValXSegmentObject` would need to be added.

Instead of just hardcoding a different value for each snapshot, since all the snapshots are unique, a string is just created on the fly and stored in the `objectMap` before the tests are started. When the `Fetch` is called, this string can be generated again to check the validity of the snapshots.

Doing this eliminates the burden of matching `snapX` to `expectedValX` both while reading the code and adding code.

pkg/snapstore/swift_snapstore_test.go
```diff
@@ -150,3 +178,50 @@ func handleDeleteObject(w http.ResponseWriter, r *http.Request) {
...
+		errorMessage := fmt.Sprintf("failed to read content %v", err)
+		logrus.Errorf(errorMessage)
+		jsonResponse.ErrorStrings = append(jsonResponse.ErrorStrings, errorMessage)
+		w.WriteHeader(http.StatusOK)
+		return
+	}
+
+	segmentObjects := strings.Split(string(buf.Bytes()), "\n")
+	// last string is an empty string, artefact of '\n' at the end
+	segmentObjects = segmentObjects[:len(segmentObjects)-1]
```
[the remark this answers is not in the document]
renormalize: Thanks for the suggestion, makes the segment object list parsing from the buffer cleaner!

From the docs of [DLOs](https://docs.openstack.org/swift/latest/overview_large_objects.html#direct-api):

> The manifest file is simply a zero-byte (not enforced) file with the extra X-Object-Manifest: <container>/<prefix> header, where <container> is the container the object segments are in and <prefix> is the common prefix for all the segments.

> The manifest file should have no content. However, this is not enforced. If the manifest path itself conforms to container/prefix specified in X-Object-Manifest, and if manifest has some content/data in it, it would also be considered as segment and manifest’s content will be part of the concatenated GET response. The order of concatenation follows the usual DLO logic which is - the order of concatenation adheres to order returned when segment names are sorted.

The segment object names are not added to the manifest object contents. The segment object files that are to be deleted are inferred from the name of the manifest object file.
Any contents that are written to the manifest object will make it be treated as a segment object too, which is not intended behavior.

pkg/snapstore/swift_snapstore_test.go
```diff
@@ -150,3 +178,50 @@ func handleDeleteObject(w http.ResponseWriter, r *http.Request) {
...
+
+	var jsonResponse bulkDeleteJSONResponse
+
+	buf := new(bytes.Buffer)
+	if _, err := io.Copy(buf, r.Body); err != nil {
+		errorMessage := fmt.Sprintf("failed to read content %v", err)
+		logrus.Errorf(errorMessage)
+		jsonResponse.ErrorStrings = append(jsonResponse.ErrorStrings, errorMessage)
+		w.WriteHeader(http.StatusOK)
+		return
```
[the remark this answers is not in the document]
renormalize: The Bulk Delete Response [docs](https://docs.openstack.org/swift/latest/api/bulk-delete.html#bulk-delete-response):

> When Object Storage processes the request, it performs multiple sub-operations. Even if all sub-operations fail, the operation returns a 200 status. The bulk operation returns a response body that contains details that indicate which sub-operations have succeeded and failed. Some sub-operations might succeed while others fail. Examine the response body to determine the results of each delete sub-operation.

Since the response from Swift would be the 200 code always, I've emulated that in the tests too, and information regarding failure and success regarding the sub-operations is sent in the body of the response as done here.
I've not emulated the response body content structure of Swift since it would be unnecessary in unit tests.

pkg/snapstore/swift_snapstore_test.go
renormalize: Will just log the errors that occur, since the body of the response is not consumed anyway.

Should the response code be something other than 200, since Swift always returns 200 as explained in https://github.com/gardener/etcd-backup-restore/pull/710#discussion_r1512209890

pkg/snapstore/swift_snapstore_test.go
renormalize: The `Result` struct would not be sufficient since we need to extract information about the operation. Instead we should be using the following struct to obtain the relevant information:

https://github.com/gardener/etcd-backup-restore/blob/4f1f191f886124d9c45e8606255c0e00f7b220d9/vendor/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects/results.go#L458-L464

Once we `BulkDeleteResult.Extract()` to obtain the `BulkDeleteResponse`, we can check for errors that could have possibly occurred in the `Errors` field.

The error handling as linked by you in `swift_snapstore.go` is not sufficient since the error while performing the `POST` operation is checked for, and returned. None of the errors that could potentially occur server side are checked for.

I will remove the `bulkDeleteJSONResponse` struct and make use of `BulkDeleteResponse` to populate the response's `Body`.

pkg/snapstore/snapstore_test.go
```diff
@@ -33,26 +33,31 @@ import (
...
 	objectMap        = map[string]*[]byte{}
-	prefixV1  string = "v1"
-	prefixV2  string = "v2"
 )
 
+// testSnapStore embedds brtypes.Snapstore and contains the number of
+// objects that are stored per snapshot for a given storage provider
+type testSnapStore struct {
+	brtypes.SnapStore
+	numberObjectsPerSnapshot int
```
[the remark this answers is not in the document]
renormalize: Addressed in [ff3b90d](https://github.com/gardener/etcd-backup-restore/pull/710/commits/ff3b90dc69c5d8d3d02c782d17f58da3cdd3776e).

pkg/snapstore/snapshot.go
```diff
@@ -65,7 +65,7 @@ func ParseSnapshot(snapPath string) (*brtypes.Snapshot, error) {
 	}
 
 	lastIndex = lastIndex + 3
-	prefix := snapPath[:lastIndex]
+	prefix := snapPath[:lastIndex-1]
```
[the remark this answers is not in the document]
renormalize: It was decided to not touch legacy code in this PR, since it might have unintended consequences.
Addressed in [ff3b90d](https://github.com/gardener/etcd-backup-restore/pull/710/commits/ff3b90dc69c5d8d3d02c782d17f58da3cdd3776e).

renormalize:
@shreyas-s-rao, I've addressed all your review comments, please take a look and resolve the pending comments.

shreyas-s-rao (approved):
@renormalize thank you so much for addressing my comments. 
/lgtm

PR #711, opened 2024-01-30
Full snapshot lease update retry on failure

### LINKED ISSUE #678: [BUG] Full Snapshot lease update should be retried and should not wait for the next full snapshot trigger
**Describe the bug:**
It is possible that after taking a full snapshot, the full-snapshot-lease update fails. Reasons for failure can be many (KAPI not available, KAPI not reachable due to network issues, connections are timing out etc.). 
Currently we try and update the snapshot lease just once. See [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/snapshotter/snapshotter.go#L639-L643). 

This is not an issue with the delta snapshot, as delta snapshots get taken every 5 mins, so a re-attempt is automatically done to update the delta-snapshot-lease. Since taking full-snapshots is expensive, today we do this once in 24 hrs. So the next attempt to update the full-snapshot-lease will be after 24 hrs.

This has a side effect that the snapshot ready condition maintained by etcd-druid on the etcd resource status continues to state that the status is unknown:

Example:
```yaml
- lastTransitionTime: "2023-11-08T21:09:17Z"
    lastUpdateTime: "2023-11-09T03:44:23Z"
    message: Cannot determine etcd backup status
    reason: Unknown
    status: Unknown
    type: BackupReady
```
Consequence of this that gardener reports that not all control plane components are healthy. This was recently observed again in Live#3989.

**Expected behavior:**
Re-attempts should be made to update the full-snapshot-lease.

One way to reproduce this to scale down KAPI and trigger a full snapshot.

anveshreddy18:

Currently a Full Snapshot Lease update is triggered when a Full Snapshot ( either scheduled or out-of-schedule ) is taken. And as the Full Snapshot is taken every 24 hrs ( configurable ), if there ever is a failure in updating the Full Snapshot lease, it has to wait for the next full snapshot for it to get updated, which is a long time to wait in this case. This creates a problem which is well documented in this [issue](https://github.com/gardener/etcd-backup-restore/issues/678) by @unmarshall, thanks for that!.

This PR attempts to update the full snapshot lease by periodically trying to update it with an interval defined by `FullSnapshotLeaseUpdateInterval` in the snapshotter.healthConfig. The retry stops once the lease is upto date, so as to not make unnecessary calls to API server. Basically ensuring that full snapshot lease is upto date for most of the time. 

Fixes #678

**NOTE:** With the etcd-druid [PR#764](https://github.com/gardener/etcd-druid/pull/764) getting merged, it allows to configure the `fullSnapshotLeaseUpdateInterval` from Etcd yaml.

To test this with kind setup, remove the `get` option under lease from the role used by `etcd-test` serviceaccount and trigger a full snapshot, the snapshotter won't be able to fetch the lease hence failure, check the backup-restore logs, now insert the `get` back and see the lease getting updated in the next call to lease update, and the periodic retry stops. Tip: decrease the `FullSnapshotLeaseUpdateInterval` time to 1 minute to make this process faster.

**Release note**:

```improvement operator
Introduces periodic updates to the Full Snapshot Lease, addressing delays in lease updates during failures
```
```
Introduced a new flag `full-snapshot-lease-update-interval` that can be used to set the periodic interval for full snapshot lease update. If the flag is not set, default interval of 3 minutes is considered. 
```

### DIFF
--- a/.ci/unit_test
+++ b/.ci/unit_test
@@ -38,7 +38,7 @@ test_with_coverage() {
 ################################################################################
 
 # To run a specific package, run TEST_PACKAGES=<PATH_TO_PACKAGE> make test
-TEST_PACKAGES="${TEST_PACKAGES:-"pkg/snapshot/snapshotter"}"
+TEST_PACKAGES="${TEST_PACKAGES:-"cmd pkg"}"
 
 RUN_NEGATIVE="${RUN_NEGATIVE:-"true"}"
 
--- a/.ci/unit_test
+++ b/.ci/unit_test
@@ -38,7 +38,7 @@ test_with_coverage() {
 ################################################################################
 
 # To run a specific package, run TEST_PACKAGES=<PATH_TO_PACKAGE> make test
-TEST_PACKAGES="${TEST_PACKAGES:-"cmd pkg"}"
+TEST_PACKAGES="${TEST_PACKAGES:-"pkg/snapshot/snapshotter"}"
 
 RUN_NEGATIVE="${RUN_NEGATIVE:-"true"}"
 
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -143,27 +143,28 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 	return nil
 }
 
-// UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot
-func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
+// UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot.
+// The boolean return value indicates whether the lease was updated or not
+func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) (bool, error) {
 	if k8sClientset == nil {
-		return &errors.EtcdError{
+		return false, &errors.EtcdError{
 			Message: "nil clientset passed",
 		}
 	}
 
 	if fullSnapshot == nil {
-		return &errors.EtcdError{
+		return false, &errors.EtcdError{
 			Message: "can not update full snapshot lease, nil snapshot passed",
 		}
 	}
 
 	namespace, err := utils.GetEnvVarOrError(podNamespace)
 	if err != nil {
-		return &errors.EtcdError{
+		return false, &errors.EtcdError{
 			Message: fmt.Sprintf("Pod namespace env var not present: %v", err),
 		}
 	}
-
+	fullSnapshotLeaseUpdated := false
 	// Retry on conflict is necessary because multiple actors update the full snapshot lease.
 	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		fullSnapLease := &v1.Lease{}
@@ -185,7 +186,7 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
 			if err != nil {
 				return err
 			}
-			if rev > fullSnapshot.LastRevision {
+			if rev >= fullSnapshot.LastRevision {
 				return nil
 			}
 		}
@@ -205,14 +206,14 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
 		}
 
 		logger.Info(logString, " at time ", renewedTime)
+		fullSnapshotLeaseUpdated = true
 		return nil
 	}); err != nil {
-		return &errors.EtcdError{
+		return false, &errors.EtcdError{
 			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
 		}
 	}
-
-	return nil
+	return fullSnapshotLeaseUpdated, nil
 }
 
 // UpdateDeltaSnapshotLease renews delta snapshot lease and updates the holderIdentity field with the total number or revisions stored in delta snapshots since the last full snapshot was taken
@@ -279,14 +280,15 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 }
 
 // FullSnapshotCaseLeaseUpdate Updates the fullsnapshot lease and the deltasnapshot lease as needed when a full snapshot is taken
-func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, deltaSnapshotLeaseName string) error {
-	if err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName); err != nil {
-		return &errors.EtcdError{
+func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, deltaSnapshotLeaseName string) (bool, error) {
+	fullSnapshotLeaseUpdated, err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName)
+	if err != nil {
+		return fullSnapshotLeaseUpdated, &errors.EtcdError{
 			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
 		}
+	} else {
+		return fullSnapshotLeaseUpdated, nil
 	}
-
-	return nil
 }
 
 // DeltaSnapshotCaseLeaseUpdate Updates the fullsnapshot lease and the deltasnapshot lease as needed when a delta snapshot is taken
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -279,8 +279,8 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 	return nil
 }
 
-// FullSnapshotCaseLeaseUpdate Updates the fullsnapshot lease and the deltasnapshot lease as needed when a full snapshot is taken
-func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, deltaSnapshotLeaseName string) (bool, error) {
+// FullSnapshotCaseLeaseUpdate Updates the fullsnapshot lease as needed when a full snapshot is taken
+func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) (bool, error) {
 	fullSnapshotLeaseUpdated, err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName)
 	if err != nil {
 		return fullSnapshotLeaseUpdated, &errors.EtcdError{
@@ -290,7 +290,7 @@ func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, full
 	return fullSnapshotLeaseUpdated, nil
 }
 
-// DeltaSnapshotCaseLeaseUpdate Updates the fullsnapshot lease and the deltasnapshot lease as needed when a delta snapshot is taken
+// DeltaSnapshotCaseLeaseUpdate Updates the deltasnapshot lease as needed when a delta snapshot is taken
 func DeltaSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, k8sClientset client.Client, deltaSnapshotLeaseName string, store brtypes.SnapStore) error {
 	_, latestDeltaSnapshotList, err := miscellaneous.GetLatestFullSnapshotAndDeltaSnapList(store)
 	if err == nil {
@@ -344,7 +344,7 @@ func RenewMemberLeasePeriodically(ctx context.Context, stopCh chan struct{}, hco
 			hb.logger.Info("Stopped member lease renewal timer")
 			return nil
 		case <-stopCh:
-			hb.logger.Info("Stoping the member lease renewal")
+			hb.logger.Info("Stopping the member lease renewal")
 			return nil
 		}
 	}
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -286,9 +286,8 @@ func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, full
 		return fullSnapshotLeaseUpdated, &errors.EtcdError{
 			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
 		}
-	} else {
-		return fullSnapshotLeaseUpdated, nil
 	}
+	return fullSnapshotLeaseUpdated, nil
 }
 
 // DeltaSnapshotCaseLeaseUpdate Updates the fullsnapshot lease and the deltasnapshot lease as needed when a delta snapshot is taken
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -144,27 +144,26 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 }
 
 // UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot.
-// The boolean return value indicates whether the lease was updated or not
-func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) (bool, error) {
+func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
 	if k8sClientset == nil {
-		return false, &errors.EtcdError{
+		return &errors.EtcdError{
 			Message: "nil clientset passed",
 		}
 	}
 
 	if fullSnapshot == nil {
-		return false, &errors.EtcdError{
+		return &errors.EtcdError{
 			Message: "can not update full snapshot lease, nil snapshot passed",
 		}
 	}
 
 	namespace, err := utils.GetEnvVarOrError(podNamespace)
 	if err != nil {
-		return false, &errors.EtcdError{
+		return &errors.EtcdError{
 			Message: fmt.Sprintf("Pod namespace env var not present: %v", err),
 		}
 	}
-	fullSnapshotLeaseUpdated := false
+
 	// Retry on conflict is necessary because multiple actors update the full snapshot lease.
 	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		fullSnapLease := &v1.Lease{}
@@ -206,14 +205,13 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
 		}
 
 		logger.Info(logString, " at time ", renewedTime)
-		fullSnapshotLeaseUpdated = true
 		return nil
 	}); err != nil {
-		return false, &errors.EtcdError{
+		return &errors.EtcdError{
 			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
 		}
 	}
-	return fullSnapshotLeaseUpdated, nil
+	return nil
 }
 
 // UpdateDeltaSnapshotLease renews delta snapshot lease and updates the holderIdentity field with the total number or revisions stored in delta snapshots since the last full snapshot was taken
@@ -280,14 +278,13 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 }
 
 // FullSnapshotCaseLeaseUpdate Updates the fullsnapshot lease as needed when a full snapshot is taken
-func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) (bool, error) {
-	fullSnapshotLeaseUpdated, err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName)
-	if err != nil {
-		return fullSnapshotLeaseUpdated, &errors.EtcdError{
+func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
+	if err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName); err != nil {
+		return &errors.EtcdError{
 			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
 		}
 	}
-	return fullSnapshotLeaseUpdated, nil
+	return nil
 }
 
 // DeltaSnapshotCaseLeaseUpdate Updates the deltasnapshot lease as needed when a delta snapshot is taken
--- a/pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
+++ b/pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
@@ -35,8 +35,7 @@ func (ssr *Snapshotter) RenewFullSnapshotLeasePeriodically() {
 				if err := func() error {
 					ctx, cancel := context.WithTimeout(fullSnapshotLeaseUpdateCtx, brtypes.LeaseUpdateTimeoutDuration)
 					defer cancel()
-					err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, logger, ssr.PrevFullSnapshot, ssr.K8sClientset, ssr.healthConfig.FullSnapshotLeaseName)
-					return err
+					return heartbeat.FullSnapshotCaseLeaseUpdate(ctx, logger, ssr.PrevFullSnapshot, ssr.K8sClientset, ssr.healthConfig.FullSnapshotLeaseName)
 				}(); err != nil {
 					//FullSnapshot lease update failed. Retry after interval
 					logger.Warnf("FullSnapshot lease update failed with error: %v", err)
--- a/pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
+++ b/pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
@@ -15,9 +15,9 @@ import (
 
 // RenewFullSnapshotLeasePeriodically has a timer and will periodically call FullSnapshotCaseLeaseUpdate to renew the fullsnapshot lease until it is updated or stopped.
 // The timer starts upon snapshotter initialization and is reset after every full snapshot is taken.
-func (ssr *Snapshotter) RenewFullSnapshotLeasePeriodically() {
+func (ssr *Snapshotter) RenewFullSnapshotLeasePeriodically(FullSnapshotLeaseStopCh chan struct{}) {
 	logger := logrus.NewEntry(logrus.New()).WithField("actor", "FullSnapLeaseUpdater")
-	fullSnapshotLeaseUpdateInterval := ssr.healthConfig.FullSnapshotLeaseUpdateInterval.Duration
+	fullSnapshotLeaseUpdateInterval := ssr.HealthConfig.FullSnapshotLeaseUpdateInterval.Duration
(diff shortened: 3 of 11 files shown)


pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -696,6 +697,29 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
+					}
+					cancel()
+				} else {
+					ssr.logger.Infof("Skipping the FullSnapshot lease update since no full snapshot has been taken yet")
+					ssr.logger.Infof("Resetting the FullSnapshot lease to retry updating after %v", fullSnapshotLeaseUpdateRetryInterval.String())
+					ssr.fullSnapshotLeaseUpdateTimer.Stop()
+					ssr.fullSnapshotLeaseUpdateTimer.Reset(fullSnapshotLeaseUpdateRetryInterval)
+				}
+			}
+
```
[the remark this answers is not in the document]
anveshreddy18: Thanks for pointing this out Ishan. 

I chose this path because the earlier code made calls to `FullSnapshotCaseLeaseUpdate` function from the `snapshotEventHandler()` function itself, so went ahead with a similar view.

But I agree with your points on how it's meant only for managing snapshots and nothing else. I'll move this part of the code to `heartbeat.go` file as suggested. Thanks!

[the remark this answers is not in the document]
anveshreddy18: Updated this in the latest commit. PTAL @ishan16696

pkg/snapshot/snapshotter/snapshot_lease_update.go
```diff
@@ -0,0 +1,70 @@
...
+						ssr.FullSnapshotLeaseUpdateTimer.Reset(fullSnapshotLeaseUpdateRetryInterval)
+					} else {
+						logger.Infof("FullSnapshot lease successfully updated with revision %d", ssr.PrevFullSnapshot.LastRevision)
+						logger.Infof("Stopping the FullSnapshot lease update")
+						ssr.FullSnapshotLeaseUpdateTimer.Stop()
+					}
+					cancel()
+				} else {
+					logger.Infof("Skipping the FullSnapshot lease update since no full snapshot has been taken yet")
+					logger.Infof("Resetting the FullSnapshot lease to retry updating after %v", fullSnapshotLeaseUpdateRetryInterval.String())
```
ishan16696, line 55: Please remove these logs as for every interval these logs will get printed which will make debugging difficult in production as these logs will eat up the space.

pkg/snapshot/snapshotter/snapshot_lease_update.go
```diff
@@ -0,0 +1,70 @@
...
+func (ssr *Snapshotter) RenewFullSnapshotLeasePeriodically(fullSnapshotLeaseUpdateRetryInterval time.Duration) {
+	ssr.FullSnapshotLeaseUpdateTimer = time.NewTimer(fullSnapshotLeaseUpdateRetryInterval)
+	fullSnapshotLeaseUpdateCtx, fullSnapshotLeaseUpdateCancel := context.WithCancel(context.TODO())
+	defer fullSnapshotLeaseUpdateCancel()
+	for {
+		select {
+		case <-ssr.FullSnapshotLeaseUpdateTimer.C:
+			if ssr.healthConfig.SnapshotLeaseRenewalEnabled {
+				if ssr.PrevFullSnapshot != nil {
+					ctx, cancel := context.WithTimeout(fullSnapshotLeaseUpdateCtx, brtypes.LeaseUpdateTimeoutDuration)
```
ishan16696, line 41:
```suggestion
					ctx, cancel := context.WithTimeout(fullSnapshotLeaseUpdateCtx, brtypes.LeaseUpdateTimeoutDuration)
defer cancel()
```

pkg/snapshot/snapshotter/snapshot_lease_update.go
```diff
@@ -0,0 +1,70 @@
...
+						logger.Warnf("FullSnapshot lease update failed : %v", err)
+						logger.Infof("Resetting the FullSnapshot lease to retry updating with revision %d after %v", ssr.PrevFullSnapshot.LastRevision, fullSnapshotLeaseUpdateRetryInterval.String())
+						ssr.FullSnapshotLeaseUpdateTimer.Stop()
+						ssr.FullSnapshotLeaseUpdateTimer.Reset(fullSnapshotLeaseUpdateRetryInterval)
+					} else {
+						logger.Infof("FullSnapshot lease successfully updated with revision %d", ssr.PrevFullSnapshot.LastRevision)
+						logger.Infof("Stopping the FullSnapshot lease update")
+						ssr.FullSnapshotLeaseUpdateTimer.Stop()
+					}
+					cancel()
```
ishan16696, line 52: remove it from here.

pkg/snapshot/snapshotter/snapshot_lease_update.go
```diff
@@ -0,0 +1,70 @@
...
+	logger = logrus.NewEntry(logrus.New()).WithField("actor", "FullSnapLeaseUpdater")
+)
+
+// RenewFullSnapshotLeasePeriodically has a timer and will periodically call FullSnapshotCaseLeaseUpdate to renew the fullsnapshot lease until it is updated or stopped.
+// The timer starts upon snapshotter initialization and is reset after every full snapshot is taken.
+func (ssr *Snapshotter) RenewFullSnapshotLeasePeriodically(fullSnapshotLeaseUpdateRetryInterval time.Duration) {
+	ssr.FullSnapshotLeaseUpdateTimer = time.NewTimer(fullSnapshotLeaseUpdateRetryInterval)
+	fullSnapshotLeaseUpdateCtx, fullSnapshotLeaseUpdateCancel := context.WithCancel(context.TODO())
+	defer fullSnapshotLeaseUpdateCancel()
+	for {
```
ishan16696, line 36: add a log just before starting a for loop

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -196,7 +199,7 @@ func (ssr *Snapshotter) Run(stopCh <-chan struct{}, startWithFullSnapshot bool)
 			return fmt.Errorf("failed to reset full snapshot timer: %v", err)
 		}
 	}
-
+	go ssr.RenewFullSnapshotLeasePeriodically(fullSnapshotLeaseUpdateRetryInterval)
```
ishan16696, line 202: Placing a go routine here cause many problems in following cases: 
1. When snapshotter itself is not enabled ? In `etcd-events` we don't enable the snapshottter.
2. When Snapshot Lease Renewal is not enabled 
3. How to make sure this go routine will graceful terminates ?

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -902,6 +907,159 @@ var _ = Describe("Snapshotter", func() {
 				})
 			})
 		})
+
+		FDescribe("Scenarios to update full snapshot lease", func() {
```
ishan16696, line 911:
```suggestion
		Describe("Scenarios to update full snapshot lease", func() {
```

ishan16696 (changes requested):

ishan16696:
Unit tests failed locally:
```
Summarizing 3 Failures:
  [FAIL] Snapshotter running snapshotter Scenarios to update full snapshot lease Without previous full snapshot [It] cannot update the lease
  /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapshot/snapshotter/snapshotter_test.go:948
  [FAIL] Snapshotter running snapshotter Scenarios to update full snapshot lease With previous full snapshot [It] is able to fetch and update the lease
  /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapshot/snapshotter/snapshotter_test.go:995
  [FAIL] Snapshotter running snapshotter Scenarios to update full snapshot lease With previous full snapshot [It] unable to fetch the lease at first, update the lease in the next attempt
  /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapshot/snapshotter/snapshotter_test.go:1028

Ran 3 of 31 Specs in 0.563 seconds
FAIL! -- 0 Passed | 3 Failed |
```

.ci/unit_test
```diff
@@ -48,7 +48,7 @@ test_with_coverage() {
 ################################################################################
 
 # To run a specific package, run TEST_PACKAGES=<PATH_TO_PACKAGE> make test
-TEST_PACKAGES="${TEST_PACKAGES:-"cmd pkg"}"
+TEST_PACKAGES="${TEST_PACKAGES:-"pkg/snapshot/snapshotter"}"
```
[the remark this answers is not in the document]
anveshreddy18: > Why you are changing this ? Please revert this change

It was a temporary change to test only the snapshotter pkg in a not powerful machines like in concourse because those were needed for unit tests. 

I wanted to update this in the next commit along with few more changes. I will remove them in the next commit.

pkg/snapshot/snapshotter/snapshotter_test.go
anveshreddy18: This was also part of the testing commit. I will revert this too :)

pkg/snapshot/snapshotter/snapshotter.go
anveshreddy18: > When snapshotter itself is not enabled ? In etcd-events we don't enable the snapshottter.

Then the goroutine will also not start, it is run only when the snapshotter starts and ends when the snapshotter closes.

> When Snapshot Lease Renewal is not enabled

If renewal is not enabled, then there won't be any attempts to update the lease, but the timer is called only first time, then after it is not called and goroutine just sits there idle. So no affect on leases. 

> How to make sure this go routine will graceful terminates ?

This is terminated with the help of a stop channel. When the snapshotter stops ( due to any reason ), then as part of the cleanup, a message is sent on the `FullSnapshotLeaseStopCh`. 

**The life of the goroutine is strictly tied to life of the snapshotter**. And we want the lease to be updated while the snapshotter is taking snapshots right? Hence we also want the goroutine to run. That's why goroutine is run when the snapshotter starts, and ends when snapshotter ends. 

Pls correct me if I'm wrong

pkg/snapshot/snapshotter/snapshot_lease_update.go
anveshreddy18: Sure.

pkg/snapshot/snapshotter/snapshot_lease_update.go
anveshreddy18: Makes sense. I'll remove them.

anveshreddy18:
> Unit tests failed locally:
> 
> ```
> Summarizing 3 Failures:
>   [FAIL] Snapshotter running snapshotter Scenarios to update full snapshot lease Without previous full snapshot [It] cannot update the lease
>   /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapshot/snapshotter/snapshotter_test.go:948
>   [FAIL] Snapshotter running snapshotter Scenarios to update full snapshot lease With previous full snapshot [It] is able to fetch and update the lease
>   /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapshot/snapshotter/snapshotter_test.go:995
>   [FAIL] Snapshotter running snapshotter Scenarios to update full snapshot lease With previous full snapshot [It] unable to fetch the lease at first, update the lease in the next attempt
>   /Users/I539698/Desktop/review/etcd-backup-restore/pkg/snapshot/snapshotter/snapshotter_test.go:1028
> 
> Ran 3 of 31 Specs in 0.563 seconds
> FAIL! -- 0 Passed | 3 Failed |
> ```

Can you check now locally @ishan16696, now it's upto date. Earlier there was a temporary commit.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: > Then the goroutine will also not start, it is run only when the snapshotter starts and ends when the snapshotter closes

make sense

> If renewal is not enabled, as from the code below, then there won't be any attempts to update the lease
case <-ssr.FullSnapshotLeaseUpdateTimer.C:
	if ssr.healthConfig.SnapshotLeaseRenewalEnabled {

I'm still not agreeing as we are starting go routine for doing nothing (in case lease renewal is not enabled), then why not check condition this before starting a go routine, WDYT ?

anveshreddy18: > why not check condition this before starting a go routine, WDYT ?

Yea that would be good. I'll address this in the next commit. Thanks!

anveshreddy18: Updated it in the latest [commit](https://github.com/gardener/etcd-backup-restore/pull/711/commits/fd879743025e89517a9bcec447358da7d3e97f97). PTAL

renormalize (changes requested):
Hey @anveshreddy18, just my 2cents on why the tests are failing on the pipeline. Please take a look!

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -902,6 +907,140 @@ var _ = Describe("Snapshotter", func() {
...
+				Expect(os.Unsetenv("POD_NAME")).To(Succeed())
+				Expect(os.Unsetenv("POD_NAMESPACE")).To(Succeed())
+			})
+			Context("Without previous full snapshot", func() {
+				It("cannot update the lease", func() {
+					snapshotterConfig := &brtypes.SnapshotterConfig{
+						FullSnapshotSchedule: fmt.Sprintf("%d %d * * *", 0, 0),
+					}
+					ssr, err = NewSnapshotter(logger, snapshotterConfig, store, etcdConnectionConfig, compressionConfig, healthConfig, snapstoreConfig)
+					Expect(err).ShouldNot(HaveOccurred())
```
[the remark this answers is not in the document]
anveshreddy18: Thanks for the review @renormalize.

I've certainly looked at the call hierarchy as part of debugging and found that `config.GetConfig` is the issue but what I'm unsure of is about the fact that multiple existing testcases which create the snapshotter the same way ([ref](https://github.com/anveshreddy18/etcd-backup-restore/blob/fd879743025e89517a9bcec447358da7d3e97f97/pkg/snapshot/snapshotter/snapshotter_test.go#L886-L887)) passes successfully in pipeline. So I think that provisioning of a cluster may not be the problem. Wdyt?

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -902,6 +907,140 @@ var _ = Describe("Snapshotter", func() {
...
 		})
+
+		Describe("Scenarios to update full snapshot lease", func() {
+			var (
+				ssr   *Snapshotter
+				lease *v1.Lease
+			)
+			BeforeEach(func() {
+				healthConfig.SnapshotLeaseRenewalEnabled = true
+				healthConfig.FullSnapshotLeaseName = brtypes.DefaultFullSnapshotLeaseName
```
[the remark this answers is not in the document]
anveshreddy18: Yup, you're right. I'll make this change.

pkg/snapshot/snapshotter/snapshotter_test.go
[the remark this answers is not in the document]
renormalize: All the other test cases which create a `Snapshotter` through the `NewSnapshotter()` invocation use the default values for `healthConfig` which is initialized by the `NewHealthConfig()` call.

For the test cases that have been added in this PR, the field `healthConfig.SnapshotLeaseRenewalEnabled` is set to `true` (it is `false` for all other test cases). This causes a function call to `GetKubernetesClientSetOrError()` in `NewSnapshotter()` which does not occur in any other test case.

`GetKubernetesClientSetOrError()` tries creating a client to a Kubernetes cluster, which would fail if there is no accessible cluster - which is what I observed in my testing.
Once I exported the `KUBECONFIG` of a running cluster, the test cases which previously failed immediately passed. Creating a client for a cluster is the only difference in `NewSnapshotter()` between the old test cases and the proposed ones in the PR.
So it is of my opinion that the lack of an accessible Kubernetes cluster is the problem.
Also see:
https://github.com/gardener/etcd-backup-restore/blob/edff58c666cef75119c10134ad9a91d92e01363c/pkg/miscellaneous/miscellaneous.go#L221-L224

which is to be used to create a client for unit tests, which is not in the call chain to fetch a client for the test cases that have been added.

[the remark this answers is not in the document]
anveshreddy18: Thanks. I didn't notice this difference. I'll modify it to use fakeclient for this.

ishan16696 (changes requested):

pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
```diff
@@ -0,0 +1,72 @@
...
+				ssr.FullSnapshotLeaseUpdateTimer.Stop()
+				ssr.FullSnapshotLeaseUpdateTimer.Reset(fullSnapshotLeaseUpdateInterval)
+			}
+
+		case <-ssr.FullSnapshotLeaseStopCh:
+			logger.Info("Closing the full snapshot lease renewal")
+			if ssr.FullSnapshotLeaseUpdateTimer != nil {
+				ssr.FullSnapshotLeaseUpdateTimer.Stop()
+				ssr.FullSnapshotLeaseUpdateTimer = nil
+			}
```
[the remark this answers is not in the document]
anveshreddy18: I think having this part of the code in the ```case <-ssr.FullSnapshotLeaseStopCh:``` condition is better in terms of readability which lets the reader know for sure that the timer gets closed at this place. 

Also since there are no other places in the function where we close the timer apart from this, so putting it in defer() may imply to the reader that it gets called in several path ways. 

These are just my views, I don't have any hard opinion on this, we can change if we want WDYT?

pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
```diff
@@ -0,0 +1,72 @@
...
+	ssr.FullSnapshotLeaseUpdateTimer = time.NewTimer(fullSnapshotLeaseUpdateInterval)
+	fullSnapshotLeaseUpdateCtx, fullSnapshotLeaseUpdateCancel := context.WithCancel(context.TODO())
+	defer fullSnapshotLeaseUpdateCancel()
+	logger.Infof("Starting the FullSnapshot lease renewal with interval %v", fullSnapshotLeaseUpdateInterval)
+	for {
+		select {
+		case <-ssr.FullSnapshotLeaseUpdateTimer.C:
+			if ssr.PrevFullSnapshot != nil {
+				ctx, cancel := context.WithTimeout(fullSnapshotLeaseUpdateCtx, brtypes.LeaseUpdateTimeoutDuration)
+				defer cancel()
```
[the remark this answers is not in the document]
anveshreddy18: 1. Yeah true, I think a simple way is to put the `cancel()` func at the end of the if condition where the context is being created which I originally added but then removed on [request](https://github.com/gardener/etcd-backup-restore/pull/711#discussion_r1482829479). 
2. The reason for the context getting created in a for loop is that before calling the `FullSnapshotCaseLeaseUpdate` func, it needs to have a `60s` timeout context. . But I feel doing the above (1) will make it deterministic that the ctx will get cancelled either by `timeout of 60s` or `cancel() func`. 

If you still feel it's not memory safe, can you pls suggest a safe way I can do this. Thanks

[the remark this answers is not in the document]
ishan16696: please use inline function. you can refer to this:
https://github.com/gardener/etcd-backup-restore/blob/a73fb83f171edf3bddfae43f11b6c163474d1f34/pkg/etcdutil/etcdutil.go#L181-L190

pkg/health/heartbeat/heartbeat.go
```diff
@@ -153,27 +153,28 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 	return nil
 }
 
-// UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot
-func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
+// UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot.
+// The boolean return value indicates whether the lease was updated or not
+func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) (bool, error) {
```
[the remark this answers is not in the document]
ishan16696: I would suggest if you are not taking different action on basis of this boolean and err then I would suggest to revert this change.

pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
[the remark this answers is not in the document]
ishan16696: It's not about readability, it's about closing the resource gracefully, that's why you need to put them inside defer().

> Also since there are no other places in the function where we close the timer apart from this, so putting it in defer() may imply to the reader that it gets called in several path ways but it won't.

what makes you think so, defer inside this func, when func will get close defer will get execute.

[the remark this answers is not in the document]
anveshreddy18: > what makes you think so, defer inside this func, when func will get close defer will get execute.

Sorry I didn't explain properly. But yeah having it in defer() is much safer than handling it in a if condition. I agree 👍

pkg/health/heartbeat/heartbeat.go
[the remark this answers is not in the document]
anveshreddy18: Yeah I'm not taking any different action on this. I'll revert it.

pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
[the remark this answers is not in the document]
ishan16696: > Yeah true, I think a simple way is to put the cancel() func at the end of the if condition where the context is being created which I originally added but then removed on https://github.com/gardener/etcd-backup-restore/pull/711#discussion_r1482829479.

make sense, then I will leave it to you. Either use inline function or removed the defer from cancel. I have missed this in this comment https://github.com/gardener/etcd-backup-restore/pull/711#discussion_r1482829479.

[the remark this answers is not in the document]
anveshreddy18: I went ahead with the inline approach as you suggested. Thanks

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,148 @@ var _ = Describe("Snapshotter", func() {
...
+			)
+			BeforeEach(func() {
+				healthConfig.SnapshotLeaseRenewalEnabled = true
+				snapstoreConfig = &brtypes.SnapstoreConfig{Container: path.Join(outputDir, "default.bkp")}
+				store, err = snapstore.GetSnapstore(snapstoreConfig)
+				Expect(err).ShouldNot(HaveOccurred())
+
+				Expect(os.Setenv("POD_NAME", "test_pod")).To(Succeed())
+				Expect(os.Setenv("POD_NAMESPACE", "test_namespace")).To(Succeed())
+				os.Setenv("UNIT_TEST", "true")
```
ishan16696, line 914: I removed this and then run the tests, they are working fine locally.
So, IMO you can try removing this from usage and it will work as long as snapshotter object has fakeK8sClient which you are updating it separately.
```
ssr.K8sClientset = fake.NewClientBuilder().Build()
```

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,148 @@ var _ = Describe("Snapshotter", func() {
...
+
+					prevFullSnap := &brtypes.Snapshot{
+						Kind:          brtypes.SnapshotKindFull,
+						CreatedOn:     time.Now(),
+						StartRevision: 0,
+						LastRevision:  123,
+					}
+					prevFullSnap.GenerateSnapshotName()
+					ssr.PrevFullSnapshot = prevFullSnap
+					ssr.FullSnapshotLeaseStopCh = make(chan struct{})
```
ishan16696, line 979: While calling func `NewSnapshotter()`, this stop channel is already getting initialised then why you to initialised it seprately again, I guess you can remove this line.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,148 @@ var _ = Describe("Snapshotter", func() {
...
```
ishan16696, line 1014: same as above and for every other places.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,148 @@ var _ = Describe("Snapshotter", func() {
...
+				Expect(os.Unsetenv("POD_NAME")).To(Succeed())
+				Expect(os.Unsetenv("POD_NAMESPACE")).To(Succeed())
+				Expect(os.Unsetenv("UNIT_TEST")).To(Succeed())
+			})
+			Context("Without previous full snapshot", func() {
+				It("cannot update the lease", func() {
+					snapshotterConfig := &brtypes.SnapshotterConfig{
+						FullSnapshotSchedule: fmt.Sprintf("%d %d * * *", 0, 0),
+					}
+					ssr, err = NewSnapshotter(logger, snapshotterConfig, store, etcdConnectionConfig, compressionConfig, healthConfig, snapstoreConfig)
```
ishan16696, line 938: you are calling this in every tests, why not put this in BeforeEach ?

pkg/snapshot/snapshotter/fullsnapshotleaseupdate.go
```diff
@@ -0,0 +1,67 @@
...
+	logger.Infof("Starting the FullSnapshot lease renewal with interval %v", fullSnapshotLeaseUpdateInterval)
+	for {
+		select {
+		case <-ssr.FullSnapshotLeaseUpdateTimer.C:
+			if ssr.PrevFullSnapshot != nil {
+				if err := func() error {
+					ctx, cancel := context.WithTimeout(fullSnapshotLeaseUpdateCtx, brtypes.LeaseUpdateTimeoutDuration)
+					defer cancel()
+					err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, logger, ssr.PrevFullSnapshot, ssr.K8sClientset, ssr.healthConfig.FullSnapshotLeaseName)
+					return err
```
ishan16696, line 39:
```suggestion
					return heartbeat.FullSnapshotCaseLeaseUpdate(ctx, logger, ssr.PrevFullSnapshot, ssr.K8sClientset, ssr.healthConfig.FullSnapshotLeaseName)
```

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,148 @@ var _ = Describe("Snapshotter", func() {
...
+					Expect(err).ShouldNot(HaveOccurred())
+
+					ssr.PrevFullSnapshot = nil
+					ssr.FullSnapshotLeaseStopCh = make(chan struct{})
+					ssr.K8sClientset = fake.NewClientBuilder().Build()
```
ishan16696, line 943: this can also be a part of BeforeEach.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,148 @@ var _ = Describe("Snapshotter", func() {
...
+					ssr.FullSnapshotLeaseStopCh <- struct{}{}
+
+					l := &v1.Lease{}
+					Expect(ssr.K8sClientset.Get(ctx, client.ObjectKey{
+						Namespace: lease.Namespace,
+						Name:      lease.Name,
+					}, l)).To(Succeed())
+					Expect(*l.Spec.HolderIdentity).To(Equal(strconv.FormatInt(prevFullSnap.LastRevision, 10)))
+				})
+				It("unable to fetch the lease at first, update the lease in the next attempt", func() {
```
ishan16696, line 999:
```suggestion
				It("should not able to fetch the lease at first but should be able to update the lease in the next attempt", func() {
```

ishan16696 (changes requested):

ishan16696:
Please include the flag you have introduce`--full-snapshot-lease-update-interval` and its default value in release note.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,132 @@ var _ = Describe("Snapshotter", func() {
...
+				Expect(err).ShouldNot(HaveOccurred())
+				ssr.PrevFullSnapshot = nil
+				ssr.K8sClientset = fake.NewClientBuilder().Build()
+				ssr.HealthConfig.SnapshotLeaseRenewalEnabled = true
+				FullSnapshotLeaseStopCh = make(chan struct{})
+			})
+			AfterEach(func() {
+				Expect(os.Unsetenv("POD_NAME")).To(Succeed())
+				Expect(os.Unsetenv("POD_NAMESPACE")).To(Succeed())
+				Expect(os.Unsetenv("UNIT_TEST")).To(Succeed())
```
ishan16696, line 940: now, I guess you can remove this

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -892,6 +897,132 @@ var _ = Describe("Snapshotter", func() {
...
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      brtypes.DefaultFullSnapshotLeaseName,
+						Namespace: os.Getenv("POD_NAMESPACE"),
+					},
+				}
+				snapshotterConfig := &brtypes.SnapshotterConfig{
+					FullSnapshotSchedule: fmt.Sprintf("%d %d * * *", 0, 0),
+				}
+				os.Setenv("UNIT_TEST", "true")
```
ishan16696, line 929: same goes for here

ishan16696 (changes requested):
Overall looks good, just few nits.

pkg/snapshot/snapshotter/snapshotter_test.go
anveshreddy18: @ishan16696 Sorry, I forgot to remove those. Updated

ishan16696 (approved):
LGTM!!

renormalize (approved):

PR #715, opened 2024-02-27
Modify URL parsing for `advertise-urls` used by `etcd` sidecar

anveshreddy18:

This PR modifies the parsing logic for the `etcd configmap` parameters `initial-advertise-peer-urls` and `advertise-client-urls` as the druid PR [#771](https://github.com/gardener/etcd-druid/pull/771) now uses proper URLs for these flags instead of `@` as separator currently being used. 

Fixes #

**Release note**:

```improvement operator
Enhances parsing logic for Etcd ConfigMap parameters `initial-advertise-peer-urls` and `advertise-client-urls` to support proper URL formatting
```

### DIFF
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -541,12 +541,23 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
 
 // ParsePeerURL forms a PeerURL, given podName by parsing the initial-advertise-peer-urls
 func ParsePeerURL(initialAdvertisePeerURLs, podName string) (string, error) {
-	tokens := strings.Split(initialAdvertisePeerURLs, "@")
-	if len(tokens) < 4 {
-		return "", fmt.Errorf("invalid peer URL : %s", initialAdvertisePeerURLs)
+	tokens := strings.Split(initialAdvertisePeerURLs, "://")
+	if len(tokens) < 2 {
+		return "", fmt.Errorf("invalid URL format : %s", initialAdvertisePeerURLs)
 	}
-	domaiName := fmt.Sprintf("%s.%s.%s", tokens[1], tokens[2], "svc")
-	return fmt.Sprintf("%s://%s.%s:%s", tokens[0], podName, domaiName, tokens[3]), nil
+	protocol := tokens[0]
+	tokens = strings.Split(tokens[1], ".")
+	if len(tokens) < 2 {
+		return "", fmt.Errorf("invalid URL format : %s", initialAdvertisePeerURLs)
+	}
+	svcName := tokens[0]
+	tokens = strings.Split(tokens[1], ":")
+	if len(tokens) < 2 {
+		return "", fmt.Errorf("invalid URL format : %s", initialAdvertisePeerURLs)
+	}
+	namespace, peerPort := tokens[0], tokens[1]
+	domaiName := fmt.Sprintf("%s.%s.%s", svcName, namespace, "svc")
+	return fmt.Sprintf("%s://%s.%s:%s", protocol, podName, domaiName, peerPort), nil
 }
 
 // IsPeerURLTLSEnabled checks whether the peer address is TLS enabled or not.
--- a/pkg/miscellaneous/miscellaneous_test.go
+++ b/pkg/miscellaneous/miscellaneous_test.go
@@ -608,14 +608,14 @@ var _ = Describe("Miscellaneous Tests", func() {
 
 		Context("parse peer url", func() {
 			It("parsing well-defined initial-advertise-peer-urls", func() {
-				initialAdPeerURL = "https@etcd-events-peer@shoot--dev--test@2380"
+				initialAdPeerURL = "https://etcd-events-peer.shoot--dev--test:2380"
 				peerURL, err := ParsePeerURL(initialAdPeerURL, podName)
 				Expect(err).To(BeNil())
 				Expect(peerURL).To(Equal("https://etcd-test-pod-0.etcd-events-peer.shoot--dev--test.svc:2380"))
 			})
 
 			It("parsing malformed initial-advertise-peer-urls", func() {
-				initialAdPeerURL = "https@etcd-events-peer@shoot--dev--test"
+				initialAdPeerURL = "https://etcd-events-peer.shoot--dev--test"
 				_, err := ParsePeerURL(initialAdPeerURL, podName)
 				Expect(err).ToNot(BeNil())
 			})
@@ -724,7 +724,7 @@ var _ = Describe("Miscellaneous Tests", func() {
 		Context("with non-TLS enabled peer url", func() {
 			BeforeEach(func() {
 				etcdConfigYaml := `name: etcd1
-initial-advertise-peer-urls: http@etcd-main-peer@default@2380
+initial-advertise-peer-urls: http://etcd-main-peer.default:2380
 initial-cluster: etcd1=http://0.0.0.0:2380`
 				err := os.WriteFile(outfile, []byte(etcdConfigYaml), 0755)
 				Expect(err).ShouldNot(HaveOccurred())
@@ -740,7 +740,7 @@ initial-cluster: etcd1=http://0.0.0.0:2380`
 		Context("with TLS enabled peer url", func() {
 			BeforeEach(func() {
 				etcdConfigYaml := `name: etcd1
-initial-advertise-peer-urls: https@etcd-main-peer@default@2380
+initial-advertise-peer-urls: https://etcd-main-peer.default:2380
 initial-cluster: etcd1=https://0.0.0.0:2380`
 				err := os.WriteFile(outfile, []byte(etcdConfigYaml), 0755)
 				Expect(err).ShouldNot(HaveOccurred())
--- a/pkg/server/httpAPI.go
+++ b/pkg/server/httpAPI.go
@@ -439,7 +439,7 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	config["name"] = podName
 
 	initAdPeerURL := config["initial-advertise-peer-urls"]
-	protocol, svcName, namespace, peerPort, err := parsePeerURL(fmt.Sprint(initAdPeerURL))
+	protocol, svcName, namespace, peerPort, err := ParseAdvertiseURL(fmt.Sprint(initAdPeerURL))
 	if err != nil {
 		h.Logger.Warnf("Unable to determine service name, namespace, peer port from advertise peer urls : %v", err)
 		rw.WriteHeader(http.StatusInternalServerError)
@@ -449,7 +449,7 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	config["initial-advertise-peer-urls"] = fmt.Sprintf("%s://%s.%s:%s", protocol, podName, domaiName, peerPort)
 
 	advClientURL := config["advertise-client-urls"]
-	protocol, svcName, namespace, clientPort, err := parseAdvClientURL(fmt.Sprint(advClientURL))
+	protocol, svcName, namespace, clientPort, err := ParseAdvertiseURL(fmt.Sprint(advClientURL))
 	if err != nil {
 		h.Logger.Warnf("Unable to determine service name, namespace, peer port from advertise client url : %v", err)
 		rw.WriteHeader(http.StatusInternalServerError)
@@ -578,20 +578,24 @@ func getInitialCluster(ctx context.Context, initialCluster string, etcdConn brty
 	return initialCluster
 }
 
-func parsePeerURL(peerURL string) (string, string, string, string, error) {
-	tokens := strings.Split(peerURL, "@")
-	if len(tokens) < 4 {
-		return "", "", "", "", fmt.Errorf("total length of tokens is less than four")
+// ParseAdvertiseURL parses the advertise URL and returns the protocol, service name, namespace and port.
+func ParseAdvertiseURL(url string) (string, string, string, string, error) {
+	tokens := strings.Split(url, "://")
+	if len(tokens) < 2 {
+		return "", "", "", "", fmt.Errorf("invalid URL format : %s", url)
 	}
-	return tokens[0], tokens[1], tokens[2], tokens[3], nil
-}
-
-func parseAdvClientURL(advClientURL string) (string, string, string, string, error) {
-	tokens := strings.Split(advClientURL, "@")
-	if len(tokens) < 4 {
-		return "", "", "", "", fmt.Errorf("total length of tokens is less than four")
+	protocol := tokens[0]
+	tokens = strings.Split(tokens[1], ".")
+	if len(tokens) < 2 {
+		return "", "", "", "", fmt.Errorf("invalid URL format : %s", url)
 	}
-	return tokens[0], tokens[1], tokens[2], tokens[3], nil
+	svcName := tokens[0]
+	tokens = strings.Split(tokens[1], ":")
+	if len(tokens) < 2 {
+		return "", "", "", "", fmt.Errorf("invalid URL format : %s", url)
+	}
+	namespace, port := tokens[0], tokens[1]
+	return protocol, svcName, namespace, port, nil
 }
 
 // delegateReqToLeader forwards the incoming http/https request to BackupLeader.
--- a/pkg/member/member_control_test.go
+++ b/pkg/member/member_control_test.go
@@ -65,7 +65,7 @@ var _ = Describe("Membercontrol", func() {
     quota-backend-bytes: 1073741824
     listen-client-urls: http://0.0.0.0:2379
     advertise-client-urls: http://0.0.0.0:2379
-    initial-advertise-peer-urls: http@etcd-main-peer@default@2380
+    initial-advertise-peer-urls: http://etcd-main-peer.default:2380
     initial-cluster: etcd1=http://0.0.0.0:2380
     initial-cluster-token: new
     initial-cluster-state: new
--- a/test/e2e/integration/cloud_backup_test.go
+++ b/test/e2e/integration/cloud_backup_test.go
@@ -139,7 +139,7 @@ enable-v2: false
 quota-backend-bytes: 1073741824
 listen-client-urls: http://0.0.0.0:2379
 advertise-client-urls: http://0.0.0.0:2379
-initial-advertise-peer-urls: http@etcd-main-peer@default@2380
+initial-advertise-peer-urls: http://etcd-main-peer.default:2380
 initial-cluster: etcd1=http://0.0.0.0:2380
 initial-cluster-token: new
 initial-cluster-state: new


pkg/miscellaneous/miscellaneous.go
```diff
@@ -541,12 +542,18 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+	tokens := re.FindStringSubmatch(initialAdvertisePeerURLs)
+	if len(tokens) != 5 {
+		return "", fmt.Errorf("invalid URL format: %s", initialAdvertisePeerURLs)
+	}
+	scheme = tokens[1]
+	peerSvcName = tokens[2]
+	namespace = tokens[3]
+	peerPort = tokens[4]
+	domainName := fmt.Sprintf("%s.%s.%s", peerSvcName, namespace, "svc")
+	return fmt.Sprintf("%s://%s.%s:%s", scheme, podName, domainName, peerPort), nil
```
renormalize, line 556: The intent of the functionality is clearer to the reader when we perform the parsing manually instead of using regex.

If the input is malformed, the user would see more directed error messages like `the hostname for the initialAdvertisePeerURLs must be of the form "peerServiceName.namespace"`, and `the peer port must be specified` instead of getting a regex error.

What are your opinions on this way of doing it?

```suggestion
	u, err := url.Parse(initialAdvertisePeerURLs)
	if err != nil {
		return "", err
	}

	hostParts := strings.Split(u.Hostname(), ".")
	if len(hostParts) != 2 {
		return "", fmt.Errorf("the hostname for the initialAdvertisePeerURLs must be of the form \"peerServiceName.namespace\"")
	}
	peerSvcName, ns := hostParts[0], hostParts[1]

	if u.Port() == "" {
		return "", fmt.Errorf("the peer port must be specified")
	}

	domainName := fmt.Sprintf("%s.%s.%s", peerSvcName, ns, "svc")
	return fmt.Sprintf("%s://%s.%s:%s", u.Scheme, podName, domainName, u.Port()), nil
```

renormalize (changes requested):
Thanks for your PR @anveshreddy18!

I'm personally not a fan of regular expressions so I've suggested a different way of doing it in my review. If you don't have any objections and think this method is clearer, please go ahead and adapt these changes to the two functions that are using regex.

Thanks.

pkg/miscellaneous/miscellaneous.go
anveshreddy18: Addressed [96818c](https://github.com/gardener/etcd-backup-restore/pull/715/commits/96818c894117b3f67791590419f5f4cd6a569583)

pkg/server/httpAPI.go
```diff
@@ -439,7 +439,7 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	config["name"] = podName
 
 	initAdPeerURL := config["initial-advertise-peer-urls"]
-	protocol, svcName, namespace, peerPort, err := parsePeerURL(fmt.Sprint(initAdPeerURL))
+	protocol, svcName, namespace, peerPort, err := miscellaneous.ParseAdvertiseURL(fmt.Sprint(initAdPeerURL))
```
renormalize, line 442: Can this be replaced with a call to `CreateMemberPeerURL` instead? The logic here after calling `ParseAdvertiseURL` is identical to that in `CreateMemberPeerURL`.

Could you also add a simple comment before the function call like:
```go
// validate the URLs in the configuration file
```

pkg/server/httpAPI.go
```diff
@@ -449,7 +449,7 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	config["initial-advertise-peer-urls"] = fmt.Sprintf("%s://%s.%s:%s", protocol, podName, domaiName, peerPort)
 
 	advClientURL := config["advertise-client-urls"]
-	protocol, svcName, namespace, clientPort, err := parseAdvClientURL(fmt.Sprint(advClientURL))
+	protocol, svcName, namespace, clientPort, err := miscellaneous.ParseAdvertiseURL(fmt.Sprint(advClientURL))
```
renormalize, line 452: https://github.com/gardener/etcd-backup-restore/pull/715#discussion_r1774884199

pkg/miscellaneous/miscellaneous.go
```diff
@@ -539,14 +539,32 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
 	return c, nil
 }
 
-// ParsePeerURL forms a PeerURL, given podName by parsing the initial-advertise-peer-urls
-func ParsePeerURL(initialAdvertisePeerURLs, podName string) (string, error) {
-	tokens := strings.Split(initialAdvertisePeerURLs, "@")
-	if len(tokens) < 4 {
-		return "", fmt.Errorf("invalid peer URL : %s", initialAdvertisePeerURLs)
+// ParseAdvertiseURL parses the advertise URL and returns the protocol, service name, namespace and port.
+func ParseAdvertiseURL(advUrl string) (string, string, string, string, error) {
```
renormalize, line 543: If https://github.com/gardener/etcd-backup-restore/pull/715#discussion_r1774884199 makes sense to you, then you can go ahead and make this helper function `parseAdvertiseURL`, or simply make it a part of the function body of `CreateMemberPeerURL` instead of seperating it out into its own function. Either works for me. You can also just leave it as an exported function if you would prefer that.

renormalize (changes requested):
Just a couple opinions from my side after a second, closer look.
Things look great otherwise. @anveshreddy18.

pkg/server/httpAPI.go
anveshreddy18: Done

renormalize (approved):
Thanks for addressing all the comments @anveshreddy18!

seshachalam-yv:
@anveshreddy18 
From the perspective of `etcd-backup-restore`, there is no ConfigMap involved. It should be referred to as a config file. Can you update the release note.

ishan16696 (changes requested):
Can you also update the charts: [here](https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/etcd-configmap.yaml#L5) and [here](https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/etcd-statefulset.yaml#L317)

pkg/server/backuprestoreserver.go
```diff
@@ -571,18 +572,14 @@ func handleSsrStopRequest(ctx context.Context, handler *HTTPHandler, _ *snapshot
 }
 
 func hasPeerURLChanged(ctx context.Context, m member.Control, cli client.ClusterCloser) (bool, error) {
-	podName, err := miscellaneous.GetEnvVarOrError("POD_NAME")
-	if err != nil {
-		return false, fmt.Errorf("error reading POD_NAME env var : %v", err)
-	}
-
-	peerURLsFromEtcdConfig, err := miscellaneous.GetMemberPeerURL(miscellaneous.GetConfigFilePath(), podName)
+	peerURLsFromEtcdConfig, err := miscellaneous.GetAdvertiseURLs("initial-advertise-peer-urls", miscellaneous.GetConfigFilePath())
```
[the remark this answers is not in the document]
anveshreddy18: I will be modifying this to have separate functions as @seshachalam-yv has recommended in [comment](https://github.com/gardener/etcd-backup-restore/pull/715#discussion_r1812706929). Wdyt?

pkg/miscellaneous/testdata/valid_config.yaml
```diff
@@ -52,7 +58,13 @@ listen-peer-urls: https://0.0.0.0:2380
 
 # List of this member's peer URLs to advertise to the public.
 # The URLs needed to be a comma-separated list.
-initial-advertise-peer-urls: https@etcd-events-peer@shoot--dev--test@2380
+initial-advertise-peer-urls:
+  etcd-events-0:
+    - https://etcd-events-0.etcd-events-peer.shoot--dev--test.svc:2380
```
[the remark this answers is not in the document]
anveshreddy18: Yes, will do.

pkg/server/backuprestoreserver.go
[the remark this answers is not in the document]
ishan16696: ok, then I will take a look after that.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +535,62 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+		for _, url := range urlsList {
+			urlStr, ok := url.(string)
+			if !ok {
+				return "", fmt.Errorf("url is not a string")
+			}
+			resultMap[podName] = append(resultMap[podName], urlStr)
+		}
+	}
+	podUrls, ok := resultMap[memberName]
+	if !ok {
```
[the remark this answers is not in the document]
anveshreddy18: I have previously used two functions like you suggested, but reverted thinking it would be redundant but I agree it is much more readable to have separate functions. Hence, I have now followed your suggestion.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +540,81 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
 	}
-	domaiName := fmt.Sprintf("%s.%s.%s", tokens[1], tokens[2], "svc")
-	return fmt.Sprintf("%s://%s.%s:%s", tokens[0], podName, domaiName, tokens[3]), nil
+	return &advURLsConfig, nil
 }
 
-// IsPeerURLTLSEnabled checks whether the peer address is TLS enabled or not.
-func IsPeerURLTLSEnabled() (bool, error) {
-	podName, err := GetEnvVarOrError("POD_NAME")
+// GetInitialAdvertisePeerURLs retrieves the initial advertise peer URLs for the etcd member.
```
seshachalam-yv, line 557:
```suggestion
// GetInitialAdvertisePeerURLs retrieves the initial advertise peer URLs for the etcd member using the POD_NAME environment variable.
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +540,81 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+func GetInitialAdvertisePeerURLs(configFile string) ([]string, error) {
+	memberName, err := GetEnvVarOrError("POD_NAME")
 	if err != nil {
-		return false, err
+		return nil, err
```
seshachalam-yv, line 561:
```suggestion
		return nil, fmt.Errorf("failed to retrieve POD_NAME: %w", err)
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +540,81 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
 	}
 
-	configFile := GetConfigFilePath()
+	advURLsConfig, err := parseAdvertiseURLsConfig(configFile)
+	if err != nil {
+		return nil, err
```
seshachalam-yv, line 566:
```suggestion
		return nil, fmt.Errorf("failed to parse advertise URLs config: %w", err)
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +540,81 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+	}
+
+	peerURLs, ok := advURLsConfig.InitialAdvertisePeerURLs[memberName]
+	if !ok || len(peerURLs) == 0 {
+		return nil, fmt.Errorf("peer url not found for pod %s", memberName)
```
seshachalam-yv, line 571:
```suggestion
		return nil, fmt.Errorf("no peer URLs found for pod %s", memberName)
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +540,81 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+	}
 
-	config, err := ReadConfigFileAsMap(configFile)
+	for _, peerURL := range peerURLs {
+		if _, err := url.Parse(peerURL); err != nil {
+			return nil, fmt.Errorf("peer url %s is not valid", peerURL)
```
seshachalam-yv, line 576:
```suggestion
			return nil, fmt.Errorf("invalid peer URL %s: %w", peerURL, err)
```

seshachalam-yv (changes requested):

pkg/miscellaneous/miscellaneous.go
seshachalam-yv: Same changes applicable to `GetAdvertiseClientURLs` func

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,41 +540,81 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+}
+
+// IsPeerURLTLSEnabled checks whether the peer address is TLS enabled or not.
+func IsPeerURLTLSEnabled() (bool, error) {
+	configFile := GetConfigFilePath()
+	memberPeerURLs, err := GetInitialAdvertisePeerURLs(configFile)
+	if err != nil {
+		return false, err
+	}
+	peerURL, err := url.Parse(memberPeerURLs[0])
```
seshachalam-yv, line 614: Check whether any of the peer URLs are TLS-enabled, you can adjust the function to iterate over all provided URLs and return true as soon as it finds one with the HTTPS scheme

pkg/miscellaneous/miscellaneous_test.go
```diff
@@ -586,28 +587,164 @@ var _ = Describe("Miscellaneous Tests", func() {
...
 		)
-
-		BeforeEach(func() {
-			podName = "etcd-test-pod-0"
+		type testCase struct {
+			name     string
+			field    string
+			function func(string) ([]string, error)
+		}
+		testCases := []testCase{
```
seshachalam-yv, line 600: You can use [ginkgo Table Specs  ](https://onsi.github.io/ginkgo/#table-specs) for this case.

seshachalam-yv (changes requested):

pkg/miscellaneous/miscellaneous.go
anveshreddy18: The error itself already has the format `fmt.Errorf("missing environment variable %s", varName)` inside the `GetEnvVarOrError` , so I feel this extra logging is not needed. Wdyt?

pkg/miscellaneous/miscellaneous.go
anveshreddy18: To be more accurate, I'm now iterating over the urls and even if one of them is `http` I'm returning false. Since, for us to say peer urls are TLS enabled, every one of them should be `https`. I hope this is okay.

pkg/miscellaneous/miscellaneous_test.go
anveshreddy18: I've now used ginkgo table driven tests, pls check.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,42 +540,95 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
-	return peerURL.Scheme == https, nil
+	sort.Strings(peerURLsSchemes)
+
+	if peerURLsSchemes[0] != peerURLsSchemes[len(peerURLsSchemes)-1] {
+		return false, fmt.Errorf("peer URLs have different schemes")
+	}
+	if peerURLsSchemes[0] == https {
+		return true, nil
+	}
+	return false, nil
```
seshachalam-yv, line 631: WDYT ?

```
// IsPeerURLTLSEnabled checks whether all peer URLs are TLS-enabled (i.e., use the "https" scheme).
func IsPeerURLTLSEnabled() (bool, error) {
	memberPeerURLs, err := GetInitialAdvertisePeerURLs(GetConfigFilePath())
	if err != nil {
		return false, fmt.Errorf("failed to get initial advertise peer URLs: %w", err)
	}

	peerURLSchemes := make(map[string]bool)
	for _, peerURL := range memberPeerURLs {
		parsedPeerURL, err := url.Parse(peerURL)
		if err != nil {
			return false, fmt.Errorf("failed to parse peer URL %s: %w", peerURL, err)
		}
		peerURLSchemes[parsedPeerURL.Scheme] = true
	}

	// Check for mixed schemes
	if len(peerURLSchemes) > 1 {
		return false, fmt.Errorf("peer URLs have different schemes")
	}

	// Check if the single scheme is "https"
	if _, isTLSEnabled := peerURLSchemes[https]; isTLSEnabled {
		return true, nil
	}

	return false, nil
}
```

seshachalam-yv (approved):
/lgtm

pkg/member/member_control.go
```diff
@@ -116,7 +116,7 @@ func NewMemberControl(etcdConnConfig *brtypes.EtcdConnectionConfig) Control {
 // AddMemberAsLearner add a member as a learner to the etcd cluster
 func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
 	//Add member as learner to cluster
-	memberURL, err := miscellaneous.GetMemberPeerURL(m.configFile, m.podName)
+	memberPeerURLs, err := miscellaneous.GetInitialAdvertisePeerURLs(m.configFile)
```
ishan16696, line 119: why you have renamed this function ? I found `GetMemberPeerURL` name more intuitive .

pkg/member/member_control.go
```diff
@@ -130,10 +130,10 @@ func (m *memberControl) AddMemberAsLearner(ctx context.Context) error {
 	memAddCtx, cancel := context.WithTimeout(ctx, EtcdTimeout)
 	defer cancel()
 	start := time.Now()
-	response, err := cli.MemberAddAsLearner(memAddCtx, []string{memberURL})
+	response, err := cli.MemberAddAsLearner(memAddCtx, memberPeerURLs)
 	if err != nil {
 		if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCPeerURLExist)) || errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCMemberExist)) {
-			m.logger.Infof("Member %s already part of etcd cluster", memberURL)
+			m.logger.Infof("Member %s already part of etcd cluster", m.podName)
```
ishan16696, line 136: log should also contains the member URL as `rpctypes.Error(rpctypes.ErrGRPCPeerURLExist)` refering to URL

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,42 +540,97 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+	advURLsConfig, err := parseAdvertiseURLsConfig(configFile)
 	if err != nil {
-		return false, err
+		return nil, fmt.Errorf("failed to parse advertise URLs config: %w", err)
 	}
 
-	configFile := GetConfigFilePath()
+	peerURLs, ok := advURLsConfig.InitialAdvertisePeerURLs[memberName]
+	if !ok || len(peerURLs) == 0 {
+		return nil, fmt.Errorf("no peer URLs found for pod %s", memberName)
```
ishan16696, line 571: same config is provided accross to all members ... right ? then error saying `"no peer URLs found for pod %s", memberName` is wrong.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,42 +540,97 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
-	memberPeerURL, err := ParsePeerURL(initAdPeerURL.(string), podName)
+	advURLsConfig, err := parseAdvertiseURLsConfig(configFile)
 	if err != nil {
-		return false, err
+		return nil, fmt.Errorf("failed to parse advertise URLs config: %w", err)
+	}
+
+	clientURLs, ok := advURLsConfig.AdvertiseClientURLs[memberName]
+	if !ok || len(clientURLs) == 0 {
+		return nil, fmt.Errorf("no client URLs found for pod %s", memberName)
```
ishan16696, line 596: same as above

ishan16696 (changes requested):
few nits

ishan16696:
@anveshreddy18 please also resolves the merge conflicts.

pkg/miscellaneous/miscellaneous.go
anveshreddy18: Here the `peerURLs` contains the list of this particular member's peer urls which is fetched by using the member's name `memberName`. So I think the comment makes sense as we are not fetching peerURLs of all members. wdyt?

For context: We will start allowing multiple peer urls for a single member in the following format 
```
initial-advertise-peer-urls:
  etcd-main-0:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-1:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-2:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
advertise-client-urls:
  etcd-main-0:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-1:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-2:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
```

pkg/member/member_control.go
anveshreddy18: I only changed it because the code is using podname in the next few lines [here](https://github.com/gardener/etcd-backup-restore/blob/0bcd93ae1f3f60acce9733df9363d76c833c0568/pkg/member/member_control.go#L138-L141)

Do you want to return podname and urls ( multiple urls is possible in few cases )  both?

pkg/member/member_control.go
anveshreddy18: Just wanted to be consistent with the actual name. But no strong opinion, I can change this.

pkg/miscellaneous/miscellaneous.go
ishan16696: ok, got it.

ishan16696 (changes requested):

pkg/miscellaneous/miscellaneous.go
```diff
@@ -535,42 +540,97 @@ func ReadConfigFileAsMap(path string) (map[string]interface{}, error) {
...
+		if err != nil {
+			return false, fmt.Errorf("failed to parse peer URL %s: %w", peerURL, err)
+		}
+		peerURLSchemes[parsedPeerURL.Scheme] = true
+	}
+
+	// Check for mixed schemes
+	if len(peerURLSchemes) > 1 {
+		return false, fmt.Errorf("peer URLs have different schemes")
+	}
```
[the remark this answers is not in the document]
anveshreddy18: Yes, thanks for the find. I'll return `nil` as error here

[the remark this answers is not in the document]
anveshreddy18: If there are anymore change requests, I'll push all of them together.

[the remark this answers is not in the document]
ishan16696: > Originally thought this shouldn't be allowed for some reason, but you are right, for in-cluster peer url someone may use non-tls enabled url and for external communication, they may use tls-enabled url. And also as you said, this shouldn't be checked here in this func.
Thanks for the find. I'll return nil as error here.

yeah thanks, can you also simplfy this function also:
```go
// IsPeerURLTLSEnabled checks whether all peer URLs are TLS-enabled.
func IsPeerURLTLSEnabled() (bool, error) {
    memberPeerURLs, err := GetInitialAdvertisePeerURLs(GetConfigFilePath())
    if err != nil {
        return false, fmt.Errorf("failed to get initial advertise peer URLs: %w", err)
    }

    for _, peerURL := range memberPeerURLs {
        parsedPeerURL, err := url.Parse(peerURL)
        if err != nil {
            return false, fmt.Errorf("failed to parse peer URL %s: %w", peerURL, err)
        }
        if parsedPeerURL.Scheme != https {
            return false, nil
        }
    }

    return true, nil
}
```

> If there are anymore change requests, I'll push all of them together.

ok, I will take a final look.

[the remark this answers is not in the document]
seshachalam-yv: @ishan16696 @anveshreddy18 
Does etcd allow both HTTP and HTTPS for a single URL?

ishan16696 (changes requested):

pkg/miscellaneous/miscellaneous.go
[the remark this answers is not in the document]
ishan16696: it's a list of URL not a single url , so it's can be possible that 1 url is of TLS enabled and 2 url isn't TLS enabled.
Anyway, that's a separate discussion, my point was mainly of this func `IsPeerURLTLSEnabled` should only do what it is define to do.

anveshreddy18:
> Is backup-restore parsing of config-map backward compatible ?
Will new configmap works if it's given to older backup-restore version ?

It won't, in one of the meets between me, Sesha & Madhav, we decided to upgrade & downgrade druid along with the compatiable `backup-restore` versions. So if we have to downgrade `backup-restore`, then we need to downgrade `etcd-druid` as well. I hope this is fine as we do not downgrade as much and even if we do in rare case, we can downgrade both of them combined.

pkg/miscellaneous/miscellaneous.go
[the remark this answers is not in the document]
anveshreddy18: Updated!

ishan16696:
currently the edge case which we found in our production can also affect this feature of updating the config-map, let me explain:
1. etcd-druid updates the config map with new changes.
2. One pod is already down due to some reason, druid waits until all pods become ready.
3. Pod comes up with older backup-restore version and tries to parse the config map but config map has been updated, so backup-restore will fails

ishan16696:
> currently the edge case which we found in our production can also affect this feature of updating the config-map, let me explain:
etcd-druid updates the config map with new changes.
One pod is already down due to some reason, druid waits until all pods become ready.
Pod comes up with older backup-restore version and tries to parse the config map but config map has been updated, so backup-restore will fails
so, it's no longer about downgrading the druid, it can affect the upgrading as well.

Regarding this @anveshreddy18 and I had a call and we found that this case may not be possible but we need test this or any other edge case on similar line of thought.

ishan16696 (approved):
LGTM!!

Issue #712, opened 2024-02-02
[Enhancement] Update member URL should also restart the embedded etcd run by etcd-wrapper via restart endpoint


unmarshall:
**Enhancement (What you would like to be added):**
`etcd-wrapper` is going to support a `restart` HTTP endpoint. See https://github.com/gardener/etcd-wrapper/issues/24.
`etcd-backup-restore` should now do the following:
It should create a new function to ensure that the member peer URL is reflected in the embedded etcd process. This function should do 2 things:

1. Call member API to update the member URL (we already do this [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/backuprestoreserver.go#L231-L244))
2. Call the newly introduced `restart` endpoint as per https://github.com/gardener/etcd-wrapper/issues/24.

In addition `etcd-backup-restore` should no longer check the mounted etcd configuration to update the peer URL TLS enablement status via member lease. Instead it should make a member API call. See [code](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/miscellaneous/miscellaneous.go#L553-L578).

**Motivation (Why is this needed?):**
This will ensure that the embedded etcd managed by `etcd-wrapper` sees the updated member peer URL and will thus prevent a second restart from `etcd-druid`, thus reducing the complexity in druid.

**Approach/Hint to the implement solution (optional):**

PR #716, opened 2024-02-29
Update to member lease annotation for peer URL TLS is now done by making a member API call

unmarshall:
Update to member lease annotation for peer URL TLS is now done by making a member API call which accurately depicts the status of the enablement of change of peer URL TLS as seen by the etcd process. Previously only mounted configmap was looked at which is not sufficient  as an additional restart of the etcd process is required for this to be reflected in the etcd member.

Fixes Part of #712

**Release note**:

```improvement operator
Update to member lease annotation for peer URL TLS is now done by making a member API call
```


renormalize:
In an out of bands discussion it was mentioned that this PR would not be necessary for https://github.com/gardener/etcd-druid/pull/777 to be merged, could this be elaborated further here for documenting why so? Please correct me if my understanding is wrong.

unmarshall:
> In an out of bands discussion it was mentioned that this PR would not be necessary for [gardener/etcd-druid#777](https://github.com/gardener/etcd-druid/pull/777) to be merged, could this be elaborated further here for documenting why so? Please correct me if my understanding is wrong.

As per [Update advertise peer URLs](https://etcd.io/docs/v3.3/op-guide/runtime-configuration/) it was understood that one needs to call member API to update the peer URL and then restart the member for it come into effect. In the current code in master STS component first updates the STS mounting the peer secret volume (this causes a rolling update thus resulting in backup-restore taking the latest config map which contains the changed peer url). Once the backup-restore container restarts as part of initialization it makes the `update member API ` call. Then druid deletes the pods to force a second restart.  When working on https://github.com/gardener/etcd-druid/pull/777 it was observed that in the case where we move from replicas 1->3 (with scheme of peer URL changes from http to https) then the second restart was not really required. Druid continues to patch the STS mounting (preserving the original replica count of 1). This causes the single member to restart and while it restarts the embedded etcd already picks up the latest peer URL (with TLS enabled) and also correctly gets the tls resources that it then uses to start etcd. This effectively makes the explicit call to update the member URL and force a second restart quite useless.

In another case where there is a 3 member cluster and now replicas are increased to 5 and TLS is enabled then one restart would no longer be sufficient. However in this case this PR will still not be useful.

PR #721, opened 2024-03-21
Fix statefulset to treat provider config env vars as optional

shreyas-s-rao:
Fix statefulset to treat provider config env vars as optional. This allows users to not define certain non-mandatory fields in the provider secrets, which the statefulset can then ignore, instead of forcing the fields to be present in the provider secrets. 

Fixes #

/cc @renormalize @anveshreddy18 

**Release note**:

```improvement operator
NONE
```


anveshreddy18:
Thanks @shreyas-s-rao for the PR. Really needed for non-druid users. Makes the point for maintaining the charts as well. 
/LGTM

anveshreddy18 (approved):
Thanks @shreyas-s-rao for the PR. Really needed for non-druid users. Makes the point for maintaining the charts as well.
LGTM!!

PR #719, opened 2024-03-15
Setup SSE with customer provided keys (SSE-C) for S3-compatible providers

amold1:
**What this PR does / why we need it**:
This PR enables setting up Server-Side Encryption using SSE-C(Not AWS managed keys) keys for S3-compatible providers

**Which issue(s) this PR fixes**:
Fixes #718

**Special notes for your reviewer**:
None
**Release note**:

```improvement operator
etcd-backup-restore now supports server-side encryption using customer provided keys (SSE-C) for S3-compatible providers
```

PR #723, opened 2024-04-04
Add documentation to run `etcdbrctl` as a process locally

renormalize:

Running `etcdbrctl` locally as a process is quite useful for testing, since it enables faster iteration and smaller build times.

Lack of this documentation was a problem for the contributor in #719, and the diff here was added as a comment in it. The comment is repurposed and formatted to fit the docs here.

The rest of `docs/development/local_setup.md` was updated in relevant places, like minimum Go version, output of `etcdbrctl -v`.

Fixes #

**Release note**:
```improvement developer
Added documentation to use `etcdbrctl` as a process for testing locally, for a better developer experience and faster iteration speeds.
```

### DIFF
--- a/docs/development/local_setup.md
+++ b/docs/development/local_setup.md
@@ -86,7 +86,7 @@ You can verify the installation by running following command:
 
 Taking the example of S3 here, one has to:
 
-* Set the `AWS_APPLICATION_CREDENTIALS` environment variable if the credentials are to be stored in a directory, where each file in the directory holds a different credential
+* Set the `AWS_APPLICATION_CREDENTIALS` environment variable if the credentials are to be stored in a directory, where each file in the directory holds a different credential.
 * Set the `AWS_APPLICATION_CREDENTIALS_JSON` if the credentials are to be stored in `.json` file.
 
 Export the following environment variables according to your host OS:
@@ -100,8 +100,9 @@ Export the following environment variables according to your host OS:
 
 > For example: `export ETCD_CONF=example/01-etcd-config.yaml`, etc. on Unix based OS-es, and using the `set` command in Windows.
 
-`example/01-etcd-config.yaml` has the configuration required to set up `etcdbrctl` while running `etcd` as a process alongside `etcdbrctl`.
 `POD_NAME`, and `POD_NAMESPACE` are environment variables that `etcdbrctl` expects since it is typically run as a sidecar container alongside a [`etcd`](https://github.com/gardener/etcd-wrapper) in a pod, as an instance of the `Etcd` CR as defined in [`etcd-druid`](https://github.com/gardener/etcd-druid). These values are provided to ensure `etcdbrctl` runs.
+
+`example/01-etcd-config.yaml` has the configuration required to set up `etcdbrctl` while running `etcd` as a process alongside `etcdbrctl`. This provides information to `etcdbrctl` like the endpoint at which the `etcd` process is running, the initial cluster state of the `etcd` cluster, the directory that stores the `etcd` data and so on. In scenarios where `etcdbrctl` is used as sidecar to a `etcd` in a `Etcd` (`etcd-druid`), this information is fetched from a `ConfigMap` that is mounted to the `Etcd` pod at `/var/etcd/config/etcd.conf.yaml`.
 
 ### Credential structure
 
@@ -117,7 +118,7 @@ s3credentials
 ```
 
 > [!TIP]  
-> Each file mentioned above will consist of the credential strings only. Ensure that there are no newline/carriage return characters at the end of the files while saving these credential files. Modern text editors are typically configured to add newline characters at the end of files on format/save. The newline characters will completely change the credentials and will cause the requests with S3 to not get authenticated. For example, the region file would contain only the string eu-west-1, without any trailing characters.
+> Each file mentioned above will consist of the credential strings only. Ensure that there are no newline/carriage return characters at the end of the files while saving these credential files. Modern text editors are typically configured to add newline characters at the end of files on format/save. The newline characters will completely change the credentials and will cause the requests with S3 to not get authenticated. For example, the region file would contain only the string `eu-west-1`, without any trailing characters.
 
 #### Credentials in a JSON file
 
--- a/docs/development/local_setup.md
+++ b/docs/development/local_setup.md
@@ -1,10 +1,12 @@
+# Local Setup
+
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


docs/development/local_setup.md
```diff
@@ -68,9 +70,67 @@ Currently there are no binary builds available, but it is fairly simple to build
...
+```console
+s3credentials
+├── accessKeyID
+├── region
+└── secretAccessKey
+```
+
+Each file mentioned above will consist of the credential strings only. Ensure that there are no newline/carriage return characters at the end of the files while saving these credential files. Modern text editors are typically configured to add newline characters at the end of files on format/save. The newline characters will completely change the credentials and will cause the requests with S3 to not get authenticated.
+
+For example, the `region` file would contain *only* the string `eu-west-1`, without any trailing characters.
```
anveshreddy18, line 121: Thanks for the PR @renormalize. It looks like this part of the document is repeated twice, here and just below this. PTAL

docs/development/local_setup.md
```diff
@@ -68,9 +70,67 @@ Currently there are no binary builds available, but it is fairly simple to build
...
+
+> [!WARNING]  
+> Ensure that you do not add your `credentialDirectory` or `credential_file.json` to your version control. Easiest way to ensure this, is to avoid using `git add .`. You could add these files/directory as `.gitignore` targets, or have the credential file/directory exist in a directory which is not tracked by `git`.
+
+### Environment variables
+
+Taking the example of S3 here, one has to:
+
+* Set the `AWS_APPLICATION_CREDENTIALS` environment variable if the credentials are to be stored in a directory, where each file in the directory holds a different credential
+* Set the `AWS_APPLICATION_CREDENTIALS_JSON` if the credentials are to be stored in `.json` file.
```
anveshreddy18, line 90: I think it is better to also include the sole environment variables way of setting the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION`, and `AWS_ENDPOINT_URL_S3` ( if the user wants to test using an instance of localstack emulator ) environment variables.

Maybe we can say that this should be the preferred way as otherwise, there is always a chance of exposing the credentials in public if we are using files to authenticate.

anveshreddy18 (changes requested):
Thanks for the PR @renormalize, a couple comments from me. PTAL

docs/development/local_setup.md
renormalize: In this specific case, I only wanted to give an example of how the credentials must look, by taking the example of S3. Usage of `etcdbrctl` with a cloud provider is out of scope for this section, in my opinion.

I don't want to mix up documentation of running `etcdbrctl` as a process with using `etcdbrctl` with emulators. I think there could be a separate section which explains how to use `etcdbrctl` with all three emulators? In this section we can describe the environment variable method of passing credentials as you mentioned.

What would you suggest?

renormalize (commented):
Thanks for your review comments @anveshreddy18! Please take a look at my responses and let me know if you agree with my comments or you'd like me to go some other way.

docs/development/local_setup.md
anveshreddy18: Yeah I'm okay with not including emulators in this doc. My main point was to use environment variables for the creds, which you are okay with. All good.

renormalize: Passing credentials purely through environment variables `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_REGION` is currently not supported in `etcd-backup-restore`. For that reason, @anveshreddy18 and I decided that it wouldn't be necessary to add this to the current documentation.

docs/development/local_setup.md
```diff
@@ -68,9 +70,63 @@ Currently there are no binary builds available, but it is fairly simple to build
...
+* `ETCD_CONF=example/01-etcd-config.yaml`
+* `POD_NAME=POD_0`
+* `POD_NAMESPACE=POD_NAMESPACE_0`
+* One of the following based on the authentication type:
+  * `AWS_APPLICATION_CREDENTIALS=/path/to/your/credential/file/directory/s3credentials`
+  * `AWS_APPLICATION_CREDENTIALS_JSON=/path/to/your/credential/file/credential_file.json`
+
+> For example: `export ETCD_CONF=example/01-etcd-config.yaml`, etc. on Unix based OS-es, and using the `set` command in Windows.
+
+`example/01-etcd-config.yaml` has the configuration required to set up `etcdbrctl` while running `etcd` as a process alongside `etcdbrctl`.
```
anveshreddy18, line 103: A small suggestion, I feel it would be better to say explicitly that, this config is a dummy of the actual configmap which is required by `etcd` process. Currently it is not clear as to who uses that out of the two processes. This is assuming the reader of this doc is someone new to `etcdbr`

docs/development/local_setup.md
```diff
@@ -68,9 +70,63 @@ Currently there are no binary builds available, but it is fairly simple to build
...
+`POD_NAME`, and `POD_NAMESPACE` are environment variables that `etcdbrctl` expects since it is typically run as a sidecar container alongside a [`etcd`](https://github.com/gardener/etcd-wrapper) in a pod, as an instance of the `Etcd` CR as defined in [`etcd-druid`](https://github.com/gardener/etcd-druid). These values are provided to ensure `etcdbrctl` runs.
+
+### Credential structure
```
anveshreddy18, line 106: Can we move this section to somewhere above to the line `One of the following based on the authentication type:` .Since we are asking them to export the credential file/dir, so it would it good to tell them before that, the structure and content of those files. Ideally having it in bottom-up fashion will make it easier for the reader. WDYT?

anveshreddy18 (changes requested):

docs/development/local_setup.md
renormalize: Alright, will add that!
This ConfigMap is mounted into the `etcd-backup-restore` container in the `Etcd` pod which gives information about the `etcd` container to `etcd-backup-restore`.

docs/development/local_setup.md
renormalize: The document first explains a higher level overview before getting into the specifics, so I think it is fine in this particular case. The section immediately after the `Environment variables` is `Credential structure` anyway.

anveshreddy18: that is fine as well

anveshreddy18 (approved):
Thanks for the changes, LGTM!!

PR #730, opened 2024-04-28
Add support to read JSON credential files present in a directory passed through the `<PROVIDER>_APPLICATION_CREDENTIALS` environment variable

### LINKED ISSUE #729: [BUG] Temporarily enable passing of cloud provider credentials in JSON through the directory pointed by `${PROVIDER}_APPLICATION_CREDENTIALS`
**Describe the bug:**

[Example](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/01-aws-s3-storage-secret.yaml) provides 2 options to specify the backup bucket secret. Option 2 (with data as JSON) is not supported today and should be removed.
> NOTE: Please make the correction for all object store providers.

This only creates confusion for adopters. See https://kubernetes.slack.com/archives/CB57N0BFG/p1713778819927999

renormalize:

The dual way of passing credentials to etcd-backup-restore, but the file-type agnostic way of passing credentials in [gardener/etcd-druid](https://github.com/gardener/etcd-druid) caused confusion in the community consuming both etcd-druid and etcd-backup-restore.

Passing credentials through a JSON file in etcd-druid does not cause it to error, but is not supported by etcd-druid - and thereby causes etcd-backup-restore to error as a result, even though etcd-backup-restore supports credentials in JSON files. 
If consumers familiar with etcd-backup-restore who pass credentials as JSON typically, use the same method to pass credentials in etcd-druid, etcd-backup-restore errors.

To temporarily solve this, if credentials are passed through a JSON file in etcd-druid (i.e. as a file in a directory whose path is exported in `<PROVIDER>_APPLICATION_CREDENTIALS`), etcd-backup-restore handles this and fetches the credentials from the JSON file. If a JSON file is present in the directory, all other files are ignored.

However, as mentioned in #729, credentials will only be passed to etcd-backup-restore as individual files in a directory in 3 releases from the next (as of writing, v0.31.0), and credentials in the form of JSON files will be deprecated.

This PR makes the following changes:

* Add support to read JSON credential files present in a directory which is passed through the `<PROVIDER>_APPLICATION_CREDENTIALS` environment variable for all (applicable) providers.

* Remove examples which suggest that cloud provider credentials can be passed through a JSON file since this will be deprecated in future releases.

* Improve error handling in various `SnapStore` related files for each cloud provider.

* Updated docs to discourage usage of JSON files to pass credentials.

* Housekeeping in `pkg/server/backuprestoreserver.go`. Changed an unused named function parameter to `_`, and removed a tautological if condition.

Tested with the following cloud providers to test the feature and check for regressions:

- [x] AWS S3
- [x] Azure Blob Storage
- [x] GCS (currently only a `service_account.json` is accepted so this issue will not be faced.)
- [x] OpenStack Swift

Fixes #729 

Please update the block header and the release note to something that you feel is more appropriate.

**Release note**:

```noteworthy|action user|operator
Passing object store credentials to etcd-backup-restore in the form of a JSON file is now deprecated. Support to pass JSON credentials via file mount (File Path Set through Env: <ProviderName>_APPLICATION_CREDENTIALS_JSON) will be dropped by v0.31.0.
```

```noteworthy|action user|operator
Please switch to passing credentials through non-JSON file as mentioned in https://github.com/gardener/etcd-backup-restore/tree/master/example/storage-provider-secrets and Set File Path through Env: `<ProviderName>_APPLICATION_CREDENTIALS`
```

### DIFF
--- a/docs/deployment/getting_started.md
+++ b/docs/deployment/getting_started.md
@@ -1,24 +1,26 @@
 # Getting started
 
-Currently we don't publish the binary build with the release, but it is pretty straight forward to build it by following the steps mentioned [here](../development/local_setup.md#build). But we do publish the docker image with each release, please check the [release page](https://github.com/gardener/etcd-backup-restore/releases) for the same. Currently, release docker images are pushed to `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl` to container registry.
+The binary builds are not published with each release, but it is pretty straight forward to build it by following the steps mentioned [here](../development/local_setup.md#build). But we do publish the docker image with each release, please check the [release page](https://github.com/gardener/etcd-backup-restore/releases) for the same. Currently, release docker images are pushed to `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl` to container registry.
 
 ## Usage
 
-You can follow the `help` flag on `etcdbrctl` command and its sub-commands to know the usage details. Some of the common use cases are mentioned below. Although examples below use `AWS S3` as storage provider, etcd-backup-restore supports AWS, GCS, Azure, Openstack swift and Alicloud OSS object store. It also supports local disk as storage provider for development purposes, but it is not recommended to use this in a production environment.
+You can follow the `help` flag on `etcdbrctl` command and its sub-commands to know the usage details. Some common use cases are mentioned below. Although examples below use `AWS S3` as storage provider, etcd-backup-restore supports AWS, GCS, Azure, OpenStack Swift, and AliCloud OSS object store. It also supports local disk as storage provider for development purposes, but it is not recommended to use this in a production environment.
 
 ### Cloud Provider Credentials
 
-The procedure to provide credentials to access the cloud provider object store varies for different providers, there are various ways to pass credentials([described below](#various-ways-to-pass-credentials)), you can choose either ways but we recommend you to pass credentials through a file.
+The procedure to provide credentials to access the cloud provider object store varies for different providers, there are various ways to pass credentials([described below](#various-ways-to-pass-credentials)), however, it is *highly* recommended passing each credential as an individual file in a directory.
 
+### Various ways to pass Credentials
 
-### Various ways to pass Credentials:
+> [!WARNING]
+> It is recommended to pass credentials as files in a directory. Please avoid passing credentials as a JSON file, since this feature is planned to be deprecated soon (in *v0.31.0*).
 
-* For `AWS S3`: 
+* For `AWS S3`:
    1. The secret file should be provided, and the file path should be made available as environment variables: `AWS_APPLICATION_CREDENTIALS` or `AWS_APPLICATION_CREDENTIALS_JSON`.
    2. For `S3-compatible providers` such as MinIO, `endpoint`, `s3ForcePathStyle`, `insecureSkipVerify` and `trustedCaCert`, can also be made available in a above file to configure the S3 client to communicate to a non-AWS provider.
    3. To enable Server-Side Encryption using Customer Managed Keys for `S3-compatible providers`, use `sseCustomerKey` and `sseCustomerAlgorithm` in the credentials file above. For example, `sseCustomerAlgorithm` could be set to `AES256`, and correspondingly the `sseCustomerKey` is set to a valid AES-256 key.
 
-* For  `Google Cloud Storage`: 
+* For `Google Cloud Storage`:
    1. The service account json file should be provided in the `~/.gcp` as a `service-account-file.json` file.
    2. The service account json file should be provided, and the file path should be made available as environment variable `GOOGLE_APPLICATION_CREDENTIALS`.
    3. If using a storage API [endpoint override](https://pkg.go.dev/cloud.google.com/go#hdr-Endpoint_Override), such as a [regional endpoint](https://cloud.google.com/storage/docs/regional-endpoints) or a local GCS emulator endpoint, then the endpoint must be made available via environment variable `GOOGLE_STORAGE_API_ENDPOINT`, in the format `http[s]://host[:port]/storage/v1/`.
@@ -39,12 +41,11 @@ The procedure to provide credentials to access the cloud provider object store v
   1. The secret file should be provided, and the file path should be made available as environment variables: `OPENSHIFT_APPLICATION_CREDENTIALS` or `OPENSHIFT_APPLICATION_CREDENTIALS_JSON`.
   For development purposes, the environment variables `OCS_DISABLE_SSL` and `OCS_INSECURE_SKIP_VERIFY` can also be set to "true" or "false".
 
-
 Check the [example of storage provider secrets](https://github.com/gardener/etcd-backup-restore/tree/master/example/storage-provider-secrets)
 
 ### Taking scheduled snapshot
 
-Sub-command `snapshot` takes scheduled backups, or `snapshots` of a running `etcd` cluster, which are pushed to one of the storage providers specified above (please note that `etcd` should already be running). One can apply standard cron format scheduling for regular backup of etcd. The cron schedule is used to take full backups. The delta snapshots are taken at regular intervals in the period in between full snapshots as indicated by the `delta-snapshot-period` flag. The default for the same is 20 seconds.
+Sub-command `snapshot` takes scheduled backups, or `snapshots` of a running `etcd` cluster, which are pushed to one of the storage providers specified above (please note that `etcd` should already be running). One can apply standard Cron format scheduling for regular backup of etcd. The Cron schedule is used to take full backups. The delta snapshots are taken at regular intervals in the period in between full snapshots as indicated by the `delta-snapshot-period` flag. The default for the same is 20 seconds.
 
 etcd-backup-restore has two garbage collection policies to clean up existing backups from the cloud bucket. The flag `garbage-collection-policy` is used to indicate the desired garbage collection policy.
 
--- a/docs/deployment/getting_started.md
+++ b/docs/deployment/getting_started.md
@@ -8,16 +8,13 @@ You can follow the `help` flag on `etcdbrctl` command and its sub-commands to kn
 
 ### Cloud Provider Credentials
 
-The procedure to provide credentials to access the cloud provider object store varies for different providers, there are various ways to pass credentials([described below](#various-ways-to-pass-credentials)), however, it is *highly* recommended passing each credential as an individual file in a directory.
+The procedure to provide credentials to access the cloud provider object store varies for different providers, the method to pass credentials for each provider is [described below](#passing-credentials).
 
-### Various ways to pass Credentials
-
-> [!WARNING]
-> It is recommended to pass credentials as files in a directory. Please avoid passing credentials as a JSON file, since this feature is planned to be deprecated soon (in *v0.31.0*).
+### Passing Credentials
 
 * For `AWS S3`:
-   1. The secret file should be provided, and the file path should be made available as environment variables: `AWS_APPLICATION_CREDENTIALS` or `AWS_APPLICATION_CREDENTIALS_JSON`.
-   2. For `S3-compatible providers` such as MinIO, `endpoint`, `s3ForcePathStyle`, `insecureSkipVerify` and `trustedCaCert`, can also be made available in a above file to configure the S3 client to communicate to a non-AWS provider.
+   1. The secret file should be provided, and the file path should be made available as an environment variable: `AWS_APPLICATION_CREDENTIALS`.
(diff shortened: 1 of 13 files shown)


pkg/snapstore/abs_snapstore.go
```diff
@@ -116,6 +115,23 @@ func getCredentials(prefixString string) (string, string, error) {
...
+			credentials, err := readABSCredentialsJSON(jsonCredentialFile)
+			if err != nil {
+				return "", "", fmt.Errorf("error getting credentials using %v JSON file in a directory with error: %w", jsonCredentialFile, err)
+			}
+			return credentials.StorageAccount, credentials.SecretKey, nil
+		}
+		// Non JSON credential files might exist in the credential directory, do not return
+	}
+
 	if dir, isSet := os.LookupEnv(prefixString + absCredentialDirectory); isSet {
```
anveshreddy18, line 135: Can I know why can't we just club these two `if` statements together, since then there would be no code redundancy and also it will be easy to understand the flow of fetching the credentials. As the flow goes like 

```
See if `absCredentialDirectory` is set, if yes -> Check if `jsonCredentialFile` is present, if yes, read from there, if no -> proceed to read the credentials from individual files
```

I presume you wanted to keep it this way since it's a temporary change and will be reverted back in a few releases but I think 3 releases is long enough time considering there are security issues and some major upgrades going on due to which the releases won't be as frequent. I believe even if it's temporary, it will be a bit easier for the users to understand the change in a neat flow. WDYT?

pkg/snapstore/abs_snapstore.go
```diff
@@ -375,6 +391,19 @@ func (a *ABSSnapStore) Delete(snap brtypes.Snapshot) error {
...
+			return time.Time{}, err
+		}
+		if !modificationTimeStamp.IsZero() {
+			return modificationTimeStamp, nil
+		}
+		// Non JSON credential files might exist in the credential directory, do not return
+	}
+
 	if dir, isSet := os.LookupEnv(absCredentialDirectory); isSet {
 		// credential files which are essential for creating the snapstore
```
anveshreddy18, line 408: Same as . And for all other providers.

anveshreddy18 (changes requested):
Thanks for the PR @renormalize. I left a couple of comments. PTAL

pkg/snapstore/abs_snapstore.go
renormalize: I don't have a strong preference for either, but I feel this coding style is easier to read. Seperating this handling into a different `if` makes the intent clear that this handling is sepearate from the other two, i.e. an edge case that is being handled. If you strongly feel clubbing makes the code look cleaner, I will make the necessary changes.

renormalize: I don't have a strong opinion on this, but I personally feel that this is easier to read, than having a huge blob in an `if`. A separate `if` for this case makes the intent clear that the case being handled is different from the other two, because it is different from what is typically understood of how credentials are sent to etcd-backup-restore.

If you have a strong preference for this code in the same `if` condition, I'll make the changes accordingly.

renormalize:
Would 3 releases be necessary for deprecation @gardener/etcd-druid-maintainers? We're still in a 0 major version and given the release cadence of etcd-backup-restore, this will take a *long* time as @anveshreddy18 has mentioned in a review comment.

Being in "alpha" gives us the freedom to deprecate features without any repercussions. However, I do understand that etcd-backup-restore has been used by a lot of members who have come to rely on their particular workflows with regards to passing credentials, and 1 release is not enough time. Would 2 be good enough?

docs/deployment/getting_started.md
```diff
@@ -1,24 +1,26 @@
...
+You can follow the `help` flag on `etcdbrctl` command and its sub-commands to know the usage details. Some common use cases are mentioned below. Although examples below use `AWS S3` as storage provider, etcd-backup-restore supports AWS, GCS, Azure, OpenStack Swift, and AliCloud OSS object store. It also supports local disk as storage provider for development purposes, but it is not recommended to use this in a production environment.
 
 ### Cloud Provider Credentials
 
-The procedure to provide credentials to access the cloud provider object store varies for different providers, there are various ways to pass credentials([described below](#various-ways-to-pass-credentials)), you can choose either ways but we recommend you to pass credentials through a file.
+The procedure to provide credentials to access the cloud provider object store varies for different providers, there are various ways to pass credentials([described below](#various-ways-to-pass-credentials)), however, it is *highly* recommended passing each credential as an individual file in a directory.
```
ishan16696, line 11: > however, it is *highly* recommended passing each credential as an individual file in a directory.

I got your point here, but IMO this could be little ambiguous for community user: "passing each credential as an individual file".  Actually, user is passing secret in a 1 file but secrets are getting mounted as individual file. IMO, previous statement is correct. If you want it to be more accurate then mention the [examples of storage-provider-secrets](https://github.com/gardener/etcd-backup-restore/tree/master/example/storage-provider-secrets).

example/storage-provider-secrets/01-aws-s3-storage-secret.yaml
```diff
@@ -12,30 +12,3 @@ data:
 # s3ForcePathStyle: # set to `true` for S3 compliant providers
 # insecureSkipVerify: # set to `true` for S3 private deployments to skip certificate check
 # trustedCaCert: # set a trusted Certificate authority for private deployments like minio
-
-#### OR ####
```
ishan16696, line 16: I would suggest if you are removing it from examples then you can remove it from doc as well: https://github.com/gardener/etcd-backup-restore/blob/master/docs/deployment/getting_started.md#various-ways-to-pass-credentials for all storage providers.

pkg/snapstore/abs_snapstore.go
```diff
@@ -116,6 +115,23 @@ func getCredentials(prefixString string) (string, string, error) {
...
+		if jsonCredentialFile != "" {
+			credentials, err := readABSCredentialsJSON(jsonCredentialFile)
+			if err != nil {
+				return "", "", fmt.Errorf("error getting credentials using %v JSON file in a directory with error: %w", jsonCredentialFile, err)
+			}
+			return credentials.StorageAccount, credentials.SecretKey, nil
+		}
+		// Non JSON credential files might exist in the credential directory, do not return
+	}
+
```
ishan16696, line 134: I would suggest put this block of code before checking the `os.LookupEnv(prefixString + absCredentialJSONFile)`. This will save you from extra part of code.
```
See if `absCredentialDirectory` is set, if yes -> Check if `jsonCredentialFile` is present, if yes --> set the ENV `absCredentialJSONFile` and continue the code as it is
```
By this you don't have to handle to write piece of code:
```
		if jsonCredentialFile != "" {
			credentials, err := readABSCredentialsJSON(jsonCredentialFile)
			if err != nil {
				return "", "", fmt.Errorf("error getting credentials using %v JSON file in a directory with error: %w", jsonCredentialFile, err)
			}
			return credentials.StorageAccount, credentials.SecretKey, nil
		}
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -540,6 +556,19 @@ func (s *S3SnapStore) Delete(snap brtypes.Snapshot) error {
...
+		modificationTimeStamp, err := getJSONCredentialModifiedTime(dir, "AWS")
+		if err != nil {
+			return time.Time{}, err
+		}
+		if !modificationTimeStamp.IsZero() {
+			return modificationTimeStamp, nil
+		}
+		// Non JSON credential files might exist in the credential directory, do not return
+	}
+
```
ishan16696, line 571: If you follow my above suggestion then here you don't need extra handling of this, may be you need to just swap the code .... check `JSON` first then `non-JSON`

pkg/snapstore/abs_snapstore.go
renormalize: I'd specifically not chosen to do this because the environment variable is set by the consumer of etcd-backup-restore.

etcd-backup-restore changing the environment variables that it consumes from the user as a part of configuration does not feel right to me.

Let's say that a consumer of etcd-druid provides a JSON file as a part of the credential secret. During a credential rotation, they choose to provide the credentials as normal data fields which are mounted as individual files in the etcd-backup-restore container. If etcd-backup-restore sets the JSON environment variable, it will not be unset when this credential rotation happens, and the credentials will still be looked for as a JSON file.

Due to configuration to the binary being changed by the binary causing possible confusion, and the above case, I wrote the code as it is right now.

If you still feel the flow would be made simpler due to the suggestion, I will change the code.

docs/deployment/getting_started.md
renormalize: You are right. I will make the corresponding changes by pointing the user to the example secret files.
Thanks for the suggestion!

example/storage-provider-secrets/01-aws-s3-storage-secret.yaml
renormalize: Good catch! Thanks for the suggestion.

I'll remove it from the mentioned doc as well.

pkg/snapstore/abs_snapstore.go
ishan16696: good edge catch but can be easily handled by this:
```
See if `absCredentialDirectory` is set, if yes -> Check if `jsonCredentialFile` is present, if yes --> set the ENV `absCredentialJSONFile` and continue the code as it is else unset ENV for JSON.
```
And secret rotation detection shouldn't depend on any `ENV`, it should list all files and check if any file get updated or not.

ishan16696: But I leave this decision to you, basically my point was to make less code change.

renormalize: Alright, I'll take a call on this. I don't mind the larger change since it's easier to follow.

pkg/snapstore/utils.go
```diff
@@ -211,3 +212,40 @@ func getLatestCredentialsModifiedTime(credentialFiles []string) (time.Time, erro
...
+		}
+	}
+	// return the first file found with the extension
+	return credentialFile, nil
+}
+
+// getJSONCredentialModifiedTime returns the modification time of a JSON file if it is present in the directory dir.
+// This function is introduced only to support JSON files being present in the directory which is passed through the
+// PROVIDER_APPLICATION_CREDENTIAL environment variable. Will be removed by v0.31.0.
+func getJSONCredentialModifiedTime(dir, providerErrorString string) (time.Time, error) {
```
ishan16696, line 236: I don't understand the point of taking `providerErrorString` as an argument as this function `getJSONCredentialModifiedTime` doesn't look like it is using this argument. It just using it in error string which I guess caller can do it.

ishan16696 (changes requested):

pkg/snapstore/utils.go
renormalize: Alright, will make that change! Thanks for the suggestion.

pkg/snapstore/abs_snapstore.go
anveshreddy18: Since I don't have strong opinion on this and the below comment, I leave it to you to decide. You can close or address this depending on your opinion when you come to decide on this section of code since Ishan also has some reviews on this piece of code.

renormalize: I've expressed the same opinion with Ishan as well and he's okay with it. I'll be keeping this section as is. Thanks for the review!

pkg/snapstore/utils.go
renormalize: [81620c7](https://github.com/gardener/etcd-backup-restore/pull/730/commits/81620c7e6af279c5a0e631b8fa6dfa2140f58b3f).

renormalize:
I've addressed your comments @ishan16696, feel free to let me know if you have any more suggestions!

pkg/snapstore/abs_snapstore.go
```diff
@@ -116,6 +115,23 @@ func getCredentials(prefixString string) (string, string, error) {
 		return credentials.StorageAccount, credentials.SecretKey, nil
 	}
 
+	// TODO: @renormalize Remove this extra handling in v0.31.0
```
ishan16696, line 118: I would suggest open a issue to drop the support for secret passing via JSON credential file, so that it become easy to track

ishan16696 (commented):
Overall looks good. just one nit

pkg/snapstore/abs_snapstore.go
ishan16696: assign that issue a milestone of `v0.31.0`

renormalize: Yeah will be doing that. Thanks for the suggestion!

renormalize: I've created https://github.com/gardener/etcd-backup-restore/issues/741 to track this.

ishan16696 (approved):
LGTM!!

anveshreddy18 (approved):
LGTM!! Thanks for the PR. 

Before merging, pls make sure to test with OpenStack Swift as well. Thanks

PR #744, opened 2024-07-01
Remove the `full-snapshot-lease-update-interval` flag from etcdbr API

anveshreddy18:

This PR removes the flag `full-snapshot-lease-update-interval` from the etcdbr API. The flag was introduced in PR #711 to let druid configure the interval for retrying to update full snapshot lease when the lease is stale. But from the internal discussions among the team, we've decided that this flag is of not much help and unnecessarily extends the API. So this PR removes the flag. 

Also update the default retry interval from `3 min` to `1 min` to allow faster periodic retry. For more information refer PR #711.

Fixes #

**Release note**:

```improvement operator
Reverted the decision to have the flag `full-snapshot-lease-update-interval` to configure periodic interval for full snapshot lease update. And reduce the retry interval from `3 minutes` to `1 minute`
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter_test.go
+++ b/pkg/snapshot/snapshotter/snapshotter_test.go
@@ -945,8 +945,8 @@ var _ = Describe("Snapshotter", func() {
 					err := ssr.K8sClientset.Create(ctx, lease)
 					Expect(err).ShouldNot(HaveOccurred())
 
-					ssr.HealthConfig.FullSnapshotLeaseUpdateInterval.Duration = 2 * time.Second
-					go ssr.RenewFullSnapshotLeasePeriodically(FullSnapshotLeaseStopCh)
+					fullSnapshotLeaseUpdateInterval := 2 * time.Second
+					go ssr.RenewFullSnapshotLeasePeriodically(FullSnapshotLeaseStopCh, fullSnapshotLeaseUpdateInterval)
 					time.Sleep(2 * time.Second)
 					close(FullSnapshotLeaseStopCh)
 
@@ -974,8 +974,8 @@ var _ = Describe("Snapshotter", func() {
 					err := ssr.K8sClientset.Create(ctx, lease)
 					Expect(err).ShouldNot(HaveOccurred())
 
-					ssr.HealthConfig.FullSnapshotLeaseUpdateInterval.Duration = time.Second
-					go ssr.RenewFullSnapshotLeasePeriodically(FullSnapshotLeaseStopCh)
+					fullSnapshotLeaseUpdateInterval := time.Second
+					go ssr.RenewFullSnapshotLeasePeriodically(FullSnapshotLeaseStopCh, fullSnapshotLeaseUpdateInterval)
 					time.Sleep(2 * time.Second)
 					close(FullSnapshotLeaseStopCh)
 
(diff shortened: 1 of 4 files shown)


shreyas-s-rao (approved):
/lgtm

pkg/types/health.go
```diff
@@ -49,7 +49,7 @@ type HealthConfig struct {
 func NewHealthConfig() *HealthConfig {
 	return &HealthConfig{
 		SnapshotLeaseRenewalEnabled:     DefaultSnapshotLeaseRenewalEnabled,
-		FullSnapshotLeaseUpdateInterval: wrappers.Duration{Duration: DefaultFullSnapshotLeaseUpdateInterval},
+		FullSnapshotLeaseUpdateInterval: wrappers.Duration{Duration: FullSnapshotLeaseUpdateInterval},
```
ishan16696, line 52: If now `FullSnapshotLeaseUpdateInterval` is no longer configurable via backup-restore then why it has to be a part of `HealthConfig` ?

anveshreddy18: You're right. I'll remove it from `HealthConfig` and make the `const` accessible directly from `types` pkg. Is this good?

ishan16696: yup, that will be better

anveshreddy18: Done. Pls see the latest [commit](https://github.com/gardener/etcd-backup-restore/pull/744/commits/8143434d469c1120126340da638c88a464538b3a)

pkg/types/health.go
```diff
@@ -15,8 +15,8 @@ import (
 const (
 	// DefaultSnapshotLeaseRenewalEnabled is a default value for enabling the snapshot lease renewal feature
 	DefaultSnapshotLeaseRenewalEnabled = false
-	// DefaultFullSnapshotLeaseUpdateInterval is a default interval for updating full snapshot lease
-	DefaultFullSnapshotLeaseUpdateInterval = 3 * time.Minute
+	// FullSnapshotLeaseUpdateInterval is a default interval for updating full snapshot lease
```
ishan16696, line 18:
```suggestion
	// FullSnapshotLeaseUpdateInterval is a interval for updating full snapshot lease
```
It is no longer a "default" value as it is no longer configurable.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -945,8 +945,8 @@ var _ = Describe("Snapshotter", func() {
 					err := ssr.K8sClientset.Create(ctx, lease)
 					Expect(err).ShouldNot(HaveOccurred())
 
-					ssr.HealthConfig.FullSnapshotLeaseUpdateInterval.Duration = 2 * time.Second
-					go ssr.RenewFullSnapshotLeasePeriodically(FullSnapshotLeaseStopCh)
+					fullSnapshotLeaseUpdateInterval := 2 * time.Second
```
ishan16696, line 948: you can declare this once and re-use it in all tests as I can see value of `fullSnapshotLeaseUpdateInterval ` is not much varying b/w each tests. but I leave it you

ishan16696 (commented):
Overall looks good. few nits

pkg/types/health.go
anveshreddy18: Done

pkg/snapshot/snapshotter/snapshotter_test.go
anveshreddy18: Yeah I declared it once and reused. Thanks

PR #753, opened 2024-07-23
Update the full snapshot lease with `renewTime` set to the time FullSnapshot is taken

### LINKED ISSUE #752: [Enhancement] Set Full Snap `renewTime` to the time full snapshot was taken instead of full snap lease update time.

During update of full snapshot lease, set `renewTime` to the time the full snapshot was taken, instead of setting it to the time when the lease was updated as part of retry mechanism.

Currently if a full snapshot lease is not updated upon taking a full snapshot, PR #711 has introduced a retry mechanism where in the method `heartbeat.FullSnapshotCaseLeaseUpdate` that takes care of updating a full snapshot is repeatedly called until the lease is updated. This works. But with this, the `renewTime` in the lease is set to the time when the lease is updated, ideally this makes sense but `druid` consumes the `renewTime` to check the health of backups. So instead of just retrying the method to update the lease, we should carry forward the snapshot time to be set as `renewTime` even if the lease is updated `x` hours after the snapshot is taken. This way druid knows only about the snapshot time and need not care about whether it was updated instantly or as part of retry method. 

anveshreddy18:

During update of full snapshot lease, set renewTime to the time the full snapshot was taken, instead of setting it to the time when the lease was updated as part of retry mechanism.

Fixes #752

**Release note**:

```improvement operator
improve the `renewTime` of full snapshot lease when the lease is updated as part of retry mechanism
```

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -182,7 +182,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 	if opts.EnabledLeaseRenewal {
 		// Update revisions in holder identity of full snapshot lease.
 		ctx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
-		if err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, cp.logger, snapshot, cp.k8sClientset, opts.FullSnapshotLeaseName); err != nil {
+		if err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, cp.logger, snapshot, cp.k8sClientset, opts.FullSnapshotLeaseName, time.Now()); err != nil {
 			cp.logger.Warnf("Snapshot lease update failed : %v", err)
 		}
 		cancel()
--- a/pkg/health/heartbeat/heartbeat.go
+++ b/pkg/health/heartbeat/heartbeat.go
@@ -144,7 +144,7 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 }
 
 // UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot.
-func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
+func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, fullSnapshotTime time.Time) error {
 	if k8sClientset == nil {
 		return &errors.EtcdError{
 			Message: "nil clientset passed",
@@ -191,7 +191,7 @@ func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnap
 		}
 
 		renewedLease := fullSnapLease.DeepCopy()
-		renewedTime := time.Now()
+		renewedTime := fullSnapshotTime
 		renewedLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
 		// Update revisions in fullSnapLease.Spec.HolderIdentity only when its value is less than latest fullSnap.LastRevision
 		if fullSnapLease.Spec.HolderIdentity == nil || rev < fullSnapshot.LastRevision {
@@ -278,8 +278,8 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 }
 
 // FullSnapshotCaseLeaseUpdate Updates the fullsnapshot lease as needed when a full snapshot is taken
-func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
-	if err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName); err != nil {
+func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, fullSnapshotTime time.Time) error {
+	if err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName, fullSnapshotTime); err != nil {
 		return &errors.EtcdError{
 			Message: fmt.Sprintf("Failed to update full snapshot lease: %v", err),
 		}
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -423,7 +423,7 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 				if b.config.HealthConfig.SnapshotLeaseRenewalEnabled {
 					leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
 					defer cancel()
-					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName); err != nil {
+					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName, time.Now()); err != nil {
 						b.logger.Warnf("Snapshot lease update failed : %v", err)
 					}
 				}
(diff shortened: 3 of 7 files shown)


pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -643,6 +644,7 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
 			if ssr.HealthConfig.SnapshotLeaseRenewalEnabled {
 				ssr.FullSnapshotLeaseUpdateTimer.Stop()
 				ssr.FullSnapshotLeaseUpdateTimer.Reset(time.Nanosecond)
+				FullSnapshotTriggeredTimeCh <- time.Now()
```
ishan16696, line 647: instead of passing `time.Now()` from snapshotter, why not `RenewFullSnapshotLeasePeriodically()` func will itself calculated this as `RenewFullSnapshotLeasePeriodically` will only start updating the lease when snapshot is taken then why not use this time to calculate `FullSnapshotTriggeredTime`.  
I'm suggesting this to avoid creating more channels for this small improvement and not to make code more complicated.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -670,6 +672,7 @@ func (ssr *Snapshotter) snapshotEventHandler(stopCh <-chan struct{}) error {
...
```
ishan16696, line 675: same as above

ishan16696 (changes requested):

pkg/snapshot/snapshotter/snapshotter.go
anveshreddy18: The goroutine `RenewFullSnapshotLeasePeriodically` follows the lifecycle of the snapshotter. Now if the ask is to rely on the call from `case <-ssr.FullSnapshotLeaseUpdateTimer.C`, then it can happen in two cases 1. Fresh full snapshot triggered 2. Previous full snapshot lease update is going on failing, so retry calls also comes to this case itself. So we need a way to distinguish between 1 & 2, that's why I went with having a channel which does this differentiation because the information about the new snapshot has to come from the `snapshotEventHandler`. Do you have any other easy alternative in mind that I can go for?

ishan16696: > 1. Fresh full snapshot triggered

yes, when a fresh full snapshot is triggered at that time, you should capture that timestamp to update the in full-snapshot lease.

> 2. Previous full snapshot lease update is going on failing, so retry calls also comes to this case itself. So we need a way to distinguish between 1 & 2

you can easily solve this by check whether timestamp variable is empty or not .
If it's empty then initialize it with current timestamp else don't re-initialize it keep retry with same timestamp.
```go
case <-ssr.FullSnapshotLeaseUpdateTimer.C
// check timestamp is empty or not
// if empty then initialize it with time.Now()
// else don't initialize it, so that it can keep retry with the timestamp it was 

if ssr.PrevFullSnapshot != nil {
  if err := func() error {
	  // retry logic
  } else {
	  //FullSnapshot lease successfully updated. Stop the timer
	  logger.Infof("Stopping the FullSnapshot lease update")
	  ssr.FullSnapshotLeaseUpdateTimer.Stop()
           //  here you can reset the variable to empty, so that it can get re-initialize
  }
  } else {
  //Skip the FullSnapshot lease update as no full snapshot has been taken yet. Reset the timer to retry after interval
  
  }
```

anveshreddy18: I thought of this initially but this won't solve the issue if the retry logic never succeeds till the next full snapshot also, because then the timestamp will not be nil but we should still be resetting the time in this case, so we can't use this logic. 

There is a way, we will have to check the difference in time between currentTime - recorded timestamp and decide. But relying on time duration like that when we actually can get the exact timestamp of when the snapshot has happened with the help of a channel seems not ideal.

ishan16696: >  this won't solve the issue if the retry logic never succeeds till the next full snapshot also

this has never happened and seems like a hypothetical case  😕 
but you still want to solve this then why not use this `ssr.PrevFullSnapshot.CreatedOn` timestamp ?

anveshreddy18: > this has never happened and seems like a hypothetical case

yes, since current interval is 24h it may not occur. If full snapshot duration is reduced, then this issue may occur frequently, so I think it's better to cover this 😅 

> to cover this then why not use this ssr.PrevFullSnapshot.CreatedOn timestamp

Oh nice, I didn't realise this field exists that's why I'm doing all this. Then in this case I can just pass this time to `FullSnapshotCaseLeaseUpdate` directly. I'll make this change then

pkg/compactor/compactor.go
```diff
@@ -182,7 +182,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions)
 	if opts.EnabledLeaseRenewal {
 		// Update revisions in holder identity of full snapshot lease.
 		ctx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
-		if err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, cp.logger, snapshot, cp.k8sClientset, opts.FullSnapshotLeaseName); err != nil {
+		if err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, cp.logger, snapshot, cp.k8sClientset, opts.FullSnapshotLeaseName, time.Now()); err != nil {
```
ishan16696, line 185:
```suggestion
		if err := heartbeat.FullSnapshotCaseLeaseUpdate(ctx, cp.logger, snapshot, cp.k8sClientset, opts.FullSnapshotLeaseName, snapshot.CreatedOn); err != nil {
```

pkg/server/backuprestoreserver.go
```diff
@@ -423,7 +423,7 @@ func (b *BackupRestoreServer) runEtcdProbeLoopWithSnapshotter(ctx context.Contex
 				if b.config.HealthConfig.SnapshotLeaseRenewalEnabled {
 					leaseUpdatectx, cancel := context.WithTimeout(ctx, brtypes.LeaseUpdateTimeoutDuration)
 					defer cancel()
-					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName); err != nil {
+					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName, time.Now()); err != nil {
```
ishan16696, line 426:
```suggestion
					if err = heartbeat.FullSnapshotCaseLeaseUpdate(leaseUpdatectx, b.logger, snapshot, ssr.K8sClientset, b.config.HealthConfig.FullSnapshotLeaseName, snapshot.CreatedOn); err != nil {
```

ishan16696 (changes requested):

pkg/compactor/compactor.go
anveshreddy18: Done

pkg/health/heartbeat/heartbeat.go
```diff
@@ -144,7 +144,7 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 }
 
 // UpdateFullSnapshotLease renews the full snapshot lease and updates the holderIdentity field with the last revision in the latest full snapshot.
-func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
+func UpdateFullSnapshotLease(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, fullSnapshotTime time.Time) error {
```
ishan16696, line 147: Can we use a better name for `fullSnapshotTime` like `fullSnapshotCreationTime`  or `fullSnapshotCreationTimestamp` ?

pkg/health/heartbeat/heartbeat.go
```diff
@@ -278,8 +279,8 @@ func UpdateDeltaSnapshotLease(ctx context.Context, logger *logrus.Entry, prevDel
 }
 
 // FullSnapshotCaseLeaseUpdate Updates the fullsnapshot lease as needed when a full snapshot is taken
-func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string) error {
-	if err := UpdateFullSnapshotLease(ctx, logger, fullSnapshot, k8sClientset, fullSnapshotLeaseName); err != nil {
+func FullSnapshotCaseLeaseUpdate(ctx context.Context, logger *logrus.Entry, fullSnapshot *brtypes.Snapshot, k8sClientset client.Client, fullSnapshotLeaseName string, fullSnapshotTime time.Time) error {
```
ishan16696, line 282: same as above

ishan16696 (commented):
Overall looks good, 1 nit:

pkg/health/heartbeat/heartbeat.go
anveshreddy18: Done

ishan16696 (approved):
LGTM!!

ishan16696:
@anveshreddy18 please re-trigger the pipeline tests.

Issue #727, opened 2024-04-16
[Enhancement] Do not store Kubernetes secrets as environment variables


AleksandarSavchev:
**Enhancement (What you would like to be added):**
Kubernetes secrets should not be stored as environment variables. Currently `storageAPIEndpoint` is used as an env variable in the `backup-restore` container of `etcd-main` [ref](https://github.com/gardener/etcd-druid/blob/637da10c12b8df72e46532e04a0fd4768ced60d9/pkg/utils/envvar.go#L72).

**Motivation (Why is this needed?):**
Gardener aims to comply with DISA K8s STIGs. This issue is in sync with rule [242415](https://www.stigviewer.com/stig/kubernetes/2023-08-29/finding/V-242415).

**Approach/Hint to the implement solution (optional):**
Specifically for the case of `storageAPIEndpoint` it can be stored in a `ConfigMap`, since it is not sensitive information. If it needs to stay in the `etcd-backup` secret It can be read from a mounted file. The secret is [already mounted](https://github.com/gardener/etcd-druid/blob/637da10c12b8df72e46532e04a0fd4768ced60d9/pkg/component/etcd/statefulset/statefulset.go#L741-L745) in `backup-restore` for the use of `serviceaccount.json` [ref](https://github.com/gardener/etcd-druid/blob/637da10c12b8df72e46532e04a0fd4768ced60d9/pkg/utils/envvar.go#L71).

For other cases a similar approach can be used or if possible a case specific one.

PR #756, opened 2024-07-30
Add support to override Azure Blob Storage Domain

### LINKED ISSUE #757: [Enhancement] Add support to enable the overriding of Azure Blob Storage's endpoint

I would like functionality to be added so that a consumer of etcd-backup-restore that has Azure resources with URIs that do not conform to the Azure Global URI can still make use of etcd-backup-restore.

Currently, etcd-backup-restore only supports `*.blob.core.windows.net` URIs, and has no support whatsoever for consumers who would like to use `*.blob.core.chinacloudapi.cn`.

The files in the directory that etcd-backup-restore consumes for credentials could be extended to have an extra file that specifies the URL of Azure resource.

renormalize:

* Add support to override the domain for ABS in `pkg/snapstore/abs_snapstore.go`.

* Adapt example in `example/storage-provider-secrets/00-azure-blob-storage-secret.yaml` to explain the override feature.

* Enhance documentation to explain the override feature in `docs/deployment/getting_started.md`.

For example if a consumer provides the overriding `domain` as `blob.core.windows.net`, the service URL that is formed will be:

```
https://<storage-account>.blob.core.windows.net
```

Similarly, if the overriding `domain` is provided as `blob.core.chinacloudapi.cn`, then the service URL that is formed will be:

```
https://<storage-account>.blob.core.chinacloudapi.cn
```

Fixes #757

cc @unmarshall

**Release note**:

```improvement user
The Azure Blob Storage domain can now be overridden by providing the overriding domain as a field in the Secret which provides credentials to etcd-backup-restore
```

### DIFF
--- a/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
+++ b/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
@@ -7,6 +7,7 @@ type: Opaque
 data: 
   storageAccount: YWRtaW4= # admin
   storageKey: YWRtaW4= # admin
+  # uri: blob.core.windows.net # override the ABS URI
 
 #### OR ####
 
--- a/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
+++ b/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
@@ -7,7 +7,7 @@ type: Opaque
 data: 
   storageAccount: YWRtaW4= # admin
   storageKey: YWRtaW4= # admin
-  # uri: blob.core.windows.net # override the ABS URI
+  # storageAPIEndpoint: http[s]://<storage-account>.<domain> # override the ABS endpoint
 
 #### OR ####
 
--- a/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
+++ b/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
@@ -7,7 +7,7 @@ type: Opaque
 data: 
   storageAccount: YWRtaW4= # admin
   storageKey: YWRtaW4= # admin
-  # storageAPIEndpoint: http[s]://<storage-account>.<domain> # override the ABS endpoint
+  # domain: blob.core.windows.net # override the ABS domain
 
 #### OR ####
 
--- a/pkg/snapstore/abs_snapstore.go
+++ b/pkg/snapstore/abs_snapstore.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/sirupsen/logrus"
+	"k8s.io/utils/pointer"
 
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
 )
@@ -51,14 +52,15 @@ type ABSSnapStore struct {
 }
 
 type absCredentials struct {
-	BucketName     string `json:"bucketName"`
-	SecretKey      string `json:"storageKey"`
-	StorageAccount string `json:"storageAccount"`
+	BucketName     string  `json:"bucketName"`
+	SecretKey      string  `json:"storageKey"`
+	StorageAccount string  `json:"storageAccount"`
+	URI            *string `json:"uri,omitempty"`
 }
 
 // NewABSSnapStore create new ABSSnapStore from shared configuration with specified bucket
 func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
-	storageAccount, storageKey, err := getCredentials(getEnvPrefixString(config.IsSource))
+	storageAccount, storageKey, overrideURI, err := getCredentials(getEnvPrefixString(config.IsSource))
 	if err != nil {
 		return nil, err
 	}
@@ -72,35 +74,41 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		Retry: azblob.RetryOptions{
 			TryTimeout: downloadTimeout,
 		}})
-	u, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccount, brtypes.AzureBlobStorageHostName))
+
+	absURI := brtypes.AzureBlobStorageGlobalURI
+	// configuration provides an override for Azure's URI
+	if overrideURI != nil {
+		absURI = *overrideURI
+	}
+	u, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccount, absURI))
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse service url: %v", err)
 	}
+
 	serviceURL := azblob.NewServiceURL(*u, p)
 	containerURL := serviceURL.NewContainerURL(config.Container)
 
 	return GetABSSnapstoreFromClient(config.Container, config.Prefix, config.TempDir, config.MaxParallelChunkUploads, config.MinChunkSize, &containerURL)
 }
 
-func getCredentials(prefixString string) (string, string, error) {
-
+func getCredentials(prefixString string) (string, string, *string, error) {
 	if filename, isSet := os.LookupEnv(prefixString + absCredentialJSONFile); isSet {
 		credentials, err := readABSCredentialsJSON(filename)
 		if err != nil {
-			return "", "", fmt.Errorf("error getting credentials using %v file", filename)
+			return "", "", nil, fmt.Errorf("error getting credentials using %v file with error: %w", filename, err)
 		}
-		return credentials.StorageAccount, credentials.SecretKey, nil
+		return credentials.StorageAccount, credentials.SecretKey, credentials.URI, nil
 	}
 
 	if dir, isSet := os.LookupEnv(prefixString + absCredentialDirectory); isSet {
 		credentials, err := readABSCredentialFiles(dir)
 		if err != nil {
-			return "", "", fmt.Errorf("error getting credentials from %v dir", dir)
+			return "", "", nil, fmt.Errorf("error getting credentials from %v dir with error: %w", dir, err)
 		}
-		return credentials.StorageAccount, credentials.SecretKey, nil
+		return credentials.StorageAccount, credentials.SecretKey, credentials.URI, nil
 	}
 
-	return "", "", fmt.Errorf("unable to get credentials")
+	return "", "", nil, fmt.Errorf("unable to get credentials")
 }
 
 func readABSCredentialsJSON(filename string) (*absCredentials, error) {
@@ -132,17 +140,23 @@ func readABSCredentialFiles(dirname string) (*absCredentials, error) {
 
 	for _, file := range files {
 		if file.Name() == "storageAccount" {
-			data, err := os.ReadFile(dirname + "/storageAccount")
+			data, err := os.ReadFile(filepath.Join(dirname, "storageAccount"))
 			if err != nil {
 				return nil, err
 			}
 			absConfig.StorageAccount = string(data)
 		} else if file.Name() == "storageKey" {
-			data, err := os.ReadFile(dirname + "/storageKey")
+			data, err := os.ReadFile(filepath.Join(dirname, "storageKey"))
 			if err != nil {
 				return nil, err
 			}
 			absConfig.SecretKey = string(data)
+		} else if file.Name() == "uri" {
+			data, err := os.ReadFile(filepath.Join(dirname, "uri"))
+			if err != nil {
+				return nil, err
+			}
+			absConfig.URI = pointer.String(string(data))
 		}
 	}
 
--- a/pkg/snapstore/abs_snapstore.go
+++ b/pkg/snapstore/abs_snapstore.go
@@ -52,10 +52,10 @@ type ABSSnapStore struct {
 }
 
 type absCredentials struct {
-	BucketName         string  `json:"bucketName"`
-	AccountKey         string  `json:"storageKey"`
-	AccountName        string  `json:"storageAccount"`
-	StorageAPIEndpoint *string `json:"storageAPIEndpoint,omitempty"`
+	BucketName     string  `json:"bucketName"`
+	StorageKey     string  `json:"storageKey"`
+	StorageAccount string  `json:"storageAccount"`
+	Domain         *string `json:"domain,omitempty"`
 }
 
 // NewABSSnapStore create new ABSSnapStore from shared configuration with specified bucket
@@ -65,7 +65,7 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		return nil, err
 	}
 
-	sharedKeyCredential, err := azblob.NewSharedKeyCredential(absCreds.AccountName, absCreds.AccountKey)
+	sharedKeyCredential, err := azblob.NewSharedKeyCredential(absCreds.StorageAccount, absCreds.StorageKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create shared key credentials: %v", err)
 	}
@@ -75,13 +75,12 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 			TryTimeout: downloadTimeout,
 		}})
 
-	storageEndpoint := fmt.Sprintf("https://%s.%s", absCreds.AccountName, brtypes.AzureBlobStorageHostName)
-	// Consumer provided endpoint overrides the default endpoint for Azure Blob Storage
-	if absCreds.StorageAPIEndpoint != nil {
-		storageEndpoint = *absCreds.StorageAPIEndpoint
+	domain := brtypes.AzureBlobStorageGlobalDomain
+	if absCreds.Domain != nil {
+		domain = *absCreds.Domain
 	}
 
-	u, err := url.Parse(storageEndpoint)
+	u, err := url.Parse(fmt.Sprintf("https://%s.%s", absCreds.StorageAccount, domain))
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse service url: %v", err)
 	}
@@ -145,19 +144,19 @@ func readABSCredentialFiles(dirname string) (*absCredentials, error) {
 			if err != nil {
 				return nil, err
 			}
-			absConfig.AccountName = string(data)
+			absConfig.StorageAccount = string(data)
 		} else if file.Name() == "storageKey" {
 			data, err := os.ReadFile(filepath.Join(dirname, "storageKey"))
 			if err != nil {
 				return nil, err
 			}
-			absConfig.AccountKey = string(data)
-		} else if file.Name() == "storageAPIEndpoint" {
-			data, err := os.ReadFile(filepath.Join(dirname, "storageAPIEndpoint"))
+			absConfig.StorageKey = string(data)
+		} else if file.Name() == "domain" {
+			data, err := os.ReadFile(filepath.Join(dirname, "domain"))
 			if err != nil {
 				return nil, err
 			}
-			absConfig.StorageAPIEndpoint = pointer.String(string(data))
+			absConfig.Domain = pointer.String(string(data))
 		}
 	}
 
@@ -392,7 +391,7 @@ func GetABSCredentialsLastModifiedTime() (time.Time, error) {
 }
 
 func isABSConfigEmpty(config *absCredentials) error {
-	if len(config.AccountKey) != 0 && len(config.AccountName) != 0 {
+	if len(config.StorageKey) != 0 && len(config.StorageAccount) != 0 {
 		return nil
 	}
 	return fmt.Errorf("azure object storage credentials: storageKey or storageAccount is missing")
--- a/pkg/snapstore/abs_snapstore.go
+++ b/pkg/snapstore/abs_snapstore.go
@@ -52,35 +52,36 @@ type ABSSnapStore struct {
 }
 
 type absCredentials struct {
-	BucketName     string  `json:"bucketName"`
-	SecretKey      string  `json:"storageKey"`
-	StorageAccount string  `json:"storageAccount"`
-	URI            *string `json:"uri,omitempty"`
+	BucketName         string  `json:"bucketName"`
+	AccountKey         string  `json:"storageKey"`
+	AccountName        string  `json:"storageAccount"`
+	StorageAPIEndpoint *string `json:"storageAPIEndpoint,omitempty"`
 }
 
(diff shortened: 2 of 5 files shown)


shreyas-s-rao (approved):
@renormalize thanks for the quick PR!
/lgtm

pkg/snapstore/abs_snapstore.go
```diff
@@ -72,35 +74,41 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		Retry: azblob.RetryOptions{
 			TryTimeout: downloadTimeout,
 		}})
-	u, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccount, brtypes.AzureBlobStorageHostName))
+
+	absURI := brtypes.AzureBlobStorageGlobalURI
+	// configuration provides an override for Azure's URI
+	if overrideURI != nil {
+		absURI = *overrideURI
+	}
```
anveshreddy18, line 82:
```suggestion
	absURI := brtypes.AzureBlobStorageGlobalURI
	// configuration provides an override for Azure's URI
	if overrideURI != nil && !config.IsSource {
		absURI = *overrideURI
	}
```

For copy operation, we should not set the endpoint override as the buckets may reside on different regions. Refer the way it's done for gcs [here](https://github.com/gardener/etcd-backup-restore/blob/b311db7f4c9692b4f887319a4fa115fcba5c7e02/pkg/snapstore/gcs_snapstore.go#L65-L70)
cc @shreyas-s-rao wdyt?

anveshreddy18 (commented):
Thanks for the fix @renormalize

I just have couple of minor comments. PTAL.

pkg/snapstore/abs_snapstore.go
```diff
@@ -51,14 +52,15 @@ type ABSSnapStore struct {
 }
 
 type absCredentials struct {
-	BucketName     string `json:"bucketName"`
-	SecretKey      string `json:"storageKey"`
-	StorageAccount string `json:"storageAccount"`
+	BucketName     string  `json:"bucketName"`
```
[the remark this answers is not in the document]
renormalize: A similar unused field is present here: as well https://github.com/gardener/etcd-backup-restore/blob/b311db7f4c9692b4f887319a4fa115fcba5c7e02/pkg/snapstore/s3_snapstore.go#L53-L62

We can get rid of such unused fields in another PR, since I feel touching this part of the code is out of this PR's scope.

pkg/snapstore/abs_snapstore.go
renormalize: Great find @anveshreddy18!
I unfortunately forgot checking the implementation of GCS for this.

However, I see one issue with the suggestion.
If a copy operation is to be performed between two buckets in the *same* region which both need an overridden URI for ABS, then this would fail. Should it be etcd-backup-restore's responsibility to disallow copy of contents in buckets, @shreyas-s-rao?

pkg/snapstore/abs_snapstore.go
```diff
@@ -51,14 +52,15 @@ type ABSSnapStore struct {
...
+	SecretKey      string  `json:"storageKey"`
+	StorageAccount string  `json:"storageAccount"`
+	URI            *string `json:"uri,omitempty"`
 }
 
 // NewABSSnapStore create new ABSSnapStore from shared configuration with specified bucket
 func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
-	storageAccount, storageKey, err := getCredentials(getEnvPrefixString(config.IsSource))
+	storageAccount, storageKey, overrideURI, err := getCredentials(getEnvPrefixString(config.IsSource))
```
unmarshall, line 63: why can't `getCredentials` just return the struct instead constituents of the struct?

pkg/snapstore/abs_snapstore.go
```diff
@@ -72,35 +74,41 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		Retry: azblob.RetryOptions{
 			TryTimeout: downloadTimeout,
 		}})
-	u, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccount, brtypes.AzureBlobStorageHostName))
+
+	absURI := brtypes.AzureBlobStorageGlobalURI
```
unmarshall, line 78: what is `absURI`? Should i interpret it as absolute URI or azure blob store URI?

pkg/snapstore/abs_snapstore.go
```diff
@@ -72,35 +74,41 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
...
+	// configuration provides an override for Azure's URI
+	if overrideURI != nil {
+		absURI = *overrideURI
+	}
+	u, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccount, absURI))
```
unmarshall, line 83: Using `URI` is semantically incorrect as URI identifies a resource and URL locates a resource. So here you call a domain a URI and then use that to create a URL. Why i call it a domain is because originally we wanted the consumers to specify the domain and not the URL. Therefore what you do is a bit weird. Are you expecting the end user to provide a fully formed URL in the secret if not then are you validating that? If not then who is validating it?

pkg/snapstore/abs_snapstore.go
unmarshall: @anveshreddy18 
>For copy operation, we should not set the endpoint override as the buckets may reside on different regions.

Can you explain this further? I also had a look at the GCS code along with @renormalize  and the implementation does not prevent any such thing. Also in case we have missed the code that actually does it, i agree with @renormalize  that it is not a responsibility of backup-restore to prevent such copies. Given 2 stores it copies from source to target and that is it.

unmarshall (changes requested):

pkg/snapstore/abs_snapstore.go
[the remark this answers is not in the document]
anveshreddy18: right. we can do it later then. 👍

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
```diff
@@ -7,6 +7,7 @@ type: Opaque
 data: 
   storageAccount: YWRtaW4= # admin
   storageKey: YWRtaW4= # admin
+  # uri: blob.core.windows.net # override the ABS URI
```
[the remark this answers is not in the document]
renormalize: I will make it consistent and call it `storageAPIEndpoint`. Thanks for the feedback.

pkg/snapstore/abs_snapstore.go
renormalize: Noted. Making the change.

pkg/snapstore/abs_snapstore.go
renormalize: I understand your point of view, will be removing this altogether - and will make use of the `storageAPIEndpoint` throughout.

pkg/snapstore/abs_snapstore.go
anveshreddy18: >it is not a responsibility of backup-restore to prevent such copies

@unmarshall  I agree with your latter point. For explaining the first part, what I meant was that the functionality to disable copy between buckets of different region is implemented in the `etcdbr` even though that is not being consumed by druid now. I guess this was not coded in druid after this [PR](https://github.com/gardener/etcd-backup-restore/pull/691) in backup-restore. 

I believe there are both pros and cons to that approach ( gcs way ) as it will prevent the the copy if one region is override region, and other is normal. Because with the current way ( azure way ), to do that kind of copy between different subdomains, we would probably need an additional field in the secret like `destEndpoint`. I was thinking in this point of view. 

Overall, I am fine with not having this, I just quoted what Shreyas has mentioned in that PR thinking it would be better to have it same across providers.

pkg/snapstore/abs_snapstore.go
renormalize: As per our discussions, we have decided that the consumer must provide the fully qualified URL for the storage account, where the URL will be of the form `http[s]://<storage-account>.<domain>/`.

Thanks for your suggestion.

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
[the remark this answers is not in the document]
shreyas-s-rao: I would differ here. I thought about this point, and realised that there's no real benefit in harmonizing the field name across providers, since we anyway already have different fields in the provider secret for the credentials. I would recommend using `uri` for Azure, since Azure uses this terminology, and leave it `storageAPIEndpoint` for Google. Any user who's trying to set this field will already have an idea about the provider and what the field itself means.

pkg/snapstore/abs_snapstore.go
shreyas-s-rao: In the case of GCS, there is a restriction on operations performed on a regional endpoint, mentioned here - https://cloud.google.com/storage/docs/regional-endpoints#supported-operations. IIRC, we decided that copy operations could not be supported by etcdbr when a regional endpoint is used, because the buckets could be residing in a region outside the "restricted" region, and the copy would fail.

But you're right, etcd-backup-restore should not bother about this, and simply try to perform the operation the user has specified. If the copy fails due to the restriction, it's the user's duty to check and correct it. I will make this change when fixing https://github.com/gardener/etcd-backup-restore/issues/727

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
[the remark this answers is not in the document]
shreyas-s-rao: Ok, thinking on this again, I realise that Google calls these endpoints `regional endpoints` ([ref](https://cloud.google.com/storage/docs/regional-endpoints)). And `storageAPIEndpoint` was a term we came up with in etcdbr. So maybe we need to re-look at this field name as well, or else harmonize the existing name across providers.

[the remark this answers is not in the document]
renormalize: Azure uses the terminology URI for `*.blob.core.windows.net` and `*.blob.core.chinacloudapi.cn` (the `*`, being key here). As @unmarshall mentions in this https://github.com/gardener/etcd-backup-restore/pull/756#discussion_r1696619193, this would be semantically correct since these are generic descriptors to URIs that can be formed for Azure resources.

However, using a field called `uri` for only the `blob.core.windows.net` part of the domain would be confusing since it is not a URI to the blob Azure resource/storage account.

unmarshall (approved):
/lgtm

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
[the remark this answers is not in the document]
unmarshall: Yes calling only `blob.core.windows.net`  as URI is semantically and completely incorrect as per the definition of URI  as per [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.3).  This is a part of the subdomain. Therefore asking the user for a complete URI and then calling that is fine but asking the user to only specify a part of a subdomain and calling it a URI is not correct.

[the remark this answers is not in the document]
shreyas-s-rao: Ok, makes sense. But would calling it `storageAPIEndpoint` be semantically correct, because that still refers to the entire endpoint, and not just part of it?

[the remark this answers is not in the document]
renormalize: The code has been changed to expect the full endpoint for the blob storage service, `http[s]://<storage-account-name>.<domain>` which is also the case in #691.

Is that what you mean?

pkg/snapstore/abs_snapstore.go
```diff
@@ -51,56 +52,64 @@ type ABSSnapStore struct {
 }
 
 type absCredentials struct {
-	BucketName     string `json:"bucketName"`
-	SecretKey      string `json:"storageKey"`
-	StorageAccount string `json:"storageAccount"`
+	BucketName         string  `json:"bucketName"`
+	AccountKey         string  `json:"storageKey"`
+	AccountName        string  `json:"storageAccount"`
```
kon-angelo, line 57: I would say that the mismatch between the struct field and the Json would lead to some confusion in the future

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
```diff
@@ -7,6 +7,7 @@ type: Opaque
 data: 
   storageAccount: YWRtaW4= # admin
   storageKey: YWRtaW4= # admin
+  # storageAPIEndpoint: http[s]://<storage-account>.<domain> # override the ABS endpoint
```
kon-angelo, line 10: Voting for `storageEndpoint`

pkg/snapstore/abs_snapstore.go
```diff
@@ -51,56 +52,64 @@ type ABSSnapStore struct {
...
 		Retry: azblob.RetryOptions{
 			TryTimeout: downloadTimeout,
 		}})
-	u, err := url.Parse(fmt.Sprintf("https://%s.%s", storageAccount, brtypes.AzureBlobStorageHostName))
+
+	storageEndpoint := fmt.Sprintf("https://%s.%s", absCreds.AccountName, brtypes.AzureBlobStorageHostName)
+	// Consumer provided endpoint overrides the default endpoint for Azure Blob Storage
+	if absCreds.StorageAPIEndpoint != nil {
+		storageEndpoint = *absCreds.StorageAPIEndpoint
+	}
```
kon-angelo, line 82: If I understand correctly this change:
- you expect the fully resolved URL and not just the storage domain. That means, you will no longer do the "stitching" between the domain and the storageAccount

Should you consider the `storageAccount` field then somewhat deprecated and mark it as such. The behavior with the endpoint is way too similar and it would make sense to have a "unified" way to do that in the future.

pkg/snapstore/abs_snapstore.go
renormalize: Alright, I don't particularly have a strong opinion here - I don't mind reverting the rename; however, the fields are somewhat self explanatory so I don't think it'd be an issue.

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
renormalize: We'd decided to name it `storageAPIEndpoint` since that is what the override is named currently for GCS.

Would you suggest we break away from consistency and go with this?

kon-angelo: I did not know about GCS approach, but on a local level next to `storageKey` it's less "locally consistent"

pkg/snapstore/abs_snapstore.go
renormalize: Thanks for your feedback @kon-angelo.

Yes, you are right that the override, which is stored in `absCreds.StorageAPIEndpoint` would have the full qualified URL for the Azure resource, for example:

`https://storeageaccount.blob.core.windows.net` or
`https://storeageaccount.blob.core.chinacloudapi.cn`

"Stitching" between the domain and storage account is not performed, *only* for the override. The stitch is performed for storage accounts which have the domain as `blob.core.windows.net`.

@unmarshall @shreyas-s-rao would you like to chime in on whether we should deprecate the `storageAccount` field? I agree that it would be more "unified" but puts the burden on the consumers and operators of etcd-backup-restore to now fully form the URL, and this would make the API somewhat different just for Azure, when compared to S3 and GCS.

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
renormalize: Yes, you are right. It doesn't look super coherent in that sense.

I'll wait for @unmarshall to chime in with his opinions before we decide on what to go ahead with.

kon-angelo: If it's used by the Azurite emulator with the same name already, then it's okay - no need to make this harder.

renormalize: This isn't making it harder really, the emulator field was only intended for developers so we aren't breaking any public APIs.

It's better to have a more coherent scheme now like you mentioned, and we shouldn't refrain from changing this up now in my opinion.

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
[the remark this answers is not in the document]
shreyas-s-rao: Ah ok, then perfect. We can go ahead and use any name that mentions it's an `endpoint` or `URL`.

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
shreyas-s-rao: Well, we have two choices - either use the same field name `storageAPIEndpoint` for all providers, since that provides consistency for users who use EtcdBR for multiple providers. But on the other hand, since each provider specifies its own fields already, we are already "inconsistent", in the sense that each provider already has its own set of credential fields. So using a local name for the endpoint override does make sense in this context. If this is the case, then we need to make the change for both Azure as well as GCS (since GCS uses the term `regionalEndpoint` rather than `storageAPIEndpoint`). And since v0.29.0 is already released and the field name `storageAPIEndpoint` in the GCS secret is already being used by users, it would take time and effort to roll out this change and eventually enforce it.

pkg/snapstore/abs_snapstore.go
kon-angelo: @shreyas-s-rao I understand that you can't just go ahead and delete the field. I was only hinting that you mark it as deprecated (and remove it sometime in the future potentially). From the provider-extension standpoint, it makes sense that we use one option, so I believe that the storageAccount usage will naturally fade with gardener. This is just an optimization I guess

kon-angelo: Actually, scratch that. You always need the `accountName` now because of the way the credentials are constructed:
```
storageEndpoint := fmt.Sprintf("https://%s.%s", absCreds.AccountName, brtypes.AzureBlobStorageHostName)
```

Hmm, not sure if that makes me more inclined to think that passing just the domain part is better but 🤷

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
unmarshall: @kon-angelo in any provider specific configuration we have used names which are also similar/same to the ones that are used by the respective provider in the APIs that they expose. This establishes a 1:1 correspondence with what the infra provider then provides.  

Since these secrets are already very specific to a provider. E.g. for AWS you will have accessKeyID and secretAccessKey whereas for azure you will have storageAccount and storageKey. These names are completely in line with what the provider wants. One could extend this argument and say that the endpoint that is required to connect to the object store should also follow the naming convention as defined by the provider in order to be consistent. 

From this argument using [serviceURL](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/service/client.go#L70C8-L70C17) should be sufficient. Unfortunately this is just the name of an argument to a constructor function and cannot be construed strictly as an API field name.

Similarly in GCS they just call it [Endpoint](https://github.com/googleapis/google-api-go-client/blob/v0.189.0/option/option.go#L74).

If you have already released it in v0.29.0 then we are now stuck with `storageAPIEndpoint` - i agree with @kon-angelo  that `API` is not really required in the name. Then lets stick to using one name for now and when we refactor backup-restore then we can break the API then.

pkg/snapstore/abs_snapstore.go
unmarshall: >Hmm, not sure if that makes me more inclined to think that passing just the domain part is better but 🤷

In general you are right. The consumer should not be bothered with how the service URL is constructed and that should be responsibility of the provider implementation inside etcd-backup-restore. Only the well-defined and known constituents which are required to construct the serviceURL should be part of the API.

Ideally the service endpoint URL should be the same across using an emulator and actual blob storage. If you see [this](https://github.com/Azure/Azurite?tab=readme-ov-file#differences-between-azurite-and-azure-storage) then you will realize that azurite supports 2 different serviceURL formats. The easy one was chosen when support for an emulator was introduced but then the cost of using this option was not very clearly understood. This is now clear - it causes inconsistency in the API if both needs to be supported.

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
unmarshall: Currently v0.29.0 is not being used by gardener. In addition if you have a look at [ConstructBlobServiceURL](https://github.com/gardener/etcd-backup-restore/blob/v0.29.0/pkg/snapstore/abs_snapstore.go#L85-L107) then only to support emulator we use an environment variable [AZURE_STORAGE_API_ENDPOINT](https://github.com/gardener/etcd-backup-restore/blob/v0.29.0/pkg/snapstore/abs_snapstore.go#L33). For productive use cases we do not use it.

So we have an option to change it in v0.28.x and then adapt the changes in v0.29.x without too much issues w.r.t backward compatibility since no productive use cases are impacted.

pkg/snapstore/abs_snapstore.go
renormalize: Please check [b154104](https://github.com/gardener/etcd-backup-restore/pull/756/commits/b15410461787025718d22ad3ca4823a6ca012970).

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
renormalize: Please check [b154104](https://github.com/gardener/etcd-backup-restore/pull/756/commits/b15410461787025718d22ad3ca4823a6ca012970).

pkg/snapstore/abs_snapstore.go
renormalize: Reverted.

unmarshall (approved):
/lgtm

example/storage-provider-secrets/00-azure-blob-storage-secret.yaml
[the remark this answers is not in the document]
unmarshall: now its been changed to `domain`.

kon-angelo:
/lgtm

I also opened a pull request for the extension in https://github.com/gardener/gardener-extension-provider-azure/pull/931. Feel free to check it out and see that the changes are in order.

PR #764, opened 2024-08-26
WIP: some fixes for non-HA autonomous cluster.

ishan16696:
This PR moves the creation of kubernetes clientSet to multi-node cluster only. kubernetes clientSet won't be created for nonHA etcd cluster

Fixes #

**Release note**:

```improvement operator
None
```

### DIFF
--- a/pkg/server/httpAPI.go
+++ b/pkg/server/httpAPI.go
@@ -35,7 +35,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/clientv3"
 	"k8s.io/client-go/util/retry"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -419,13 +418,6 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 	podNS := os.Getenv("POD_NAMESPACE")
 	podName := os.Getenv("POD_NAME")
 
-	clientSet, err := miscellaneous.GetKubernetesClientSetOrError()
-	if err != nil {
-		h.Logger.Warnf("Failed to create clientset: %v", err)
-		rw.WriteHeader(http.StatusInternalServerError)
-		return
-	}
-
 	config["name"] = podName
 
 	initAdPeerURL := config["initial-advertise-peer-urls"]
@@ -456,7 +448,7 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 		rw.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	state, err := h.GetClusterState(req.Context(), clusterSize, clientSet, podName, podNS)
+	state, err := h.GetClusterState(req.Context(), clusterSize, podName, podNS)
 	if err != nil {
 		h.Logger.Warnf("failed to get cluster state %v", err)
 		rw.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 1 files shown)


pkg/server/httpAPI.go
```diff
@@ -472,7 +464,7 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
 		rw.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	state, err := h.GetClusterState(req.Context(), clusterSize, clientSet, podName, podNS)
+	state, err := h.GetClusterState(req.Context(), clusterSize, podName, podNS)
 	if err != nil {
```
shreyas-s-rao, line 468: Since `GetClusterState` now returns error `failed to get clusterState` upon not getting a kubernetes client, we will return `http.StatusInternalServerError` in this call. Etcd-wrapper [handles this by returning error](https://github.com/gardener/etcd-wrapper/blob/190a4194e88e66e5f8bf240a47a68df08a466692/internal/brclient/brclient.go#L140-L142) and subsequently crashes, as it requires the config from etcdbr. 

Instead, for the moment, looks like we might need to default to some value like `new` upon error, with a limitation that scale-up is not supported when k8s client is not provided (autonomous cluster case). 

The proper way to solve this is for druid to correctly specify the intent to the etcdbr, saying whether this is a new cluster or an existing cluster, and not have etcdbr try to deduce this by checking the statefulset which is actually hacky. We can do this as part of [EtcdMember resource](https://github.com/gardener/etcd-druid/issues/702).

shreyas-s-rao (changes requested):
@ishan16696 just one comment. PTAL.

pkg/server/httpAPI.go
ishan16696: > Since GetClusterState now returns error failed to get clusterState upon not getting a kubernetes client

@shreyas-s-rao it's not a new change, in a existing code also it's return error
https://github.com/gardener/etcd-backup-restore/blob/c2180ca97585dd9a7904b599a0552607f710a2c5/pkg/server/httpAPI.go#L475
https://github.com/gardener/etcd-backup-restore/blob/c2180ca97585dd9a7904b599a0552607f710a2c5/pkg/server/httpAPI.go#L440

> Etcd-wrapper [handles this by returning error](https://github.com/gardener/etcd-wrapper/blob/190a4194e88e66e5f8bf240a47a68df08a466692/internal/brclient/brclient.go#L140-L142) and subsequently crashes, as it requires the config from etcdbr.

yes, logic behind this is to fail fast as etcd-wrapper container will restart and request the `config`.

> Instead, for the moment, looks like we might need to default to some value like new upon error, with a limitation that scale-up is not supported when k8s client is not provided (autonomous cluster case).

nopes, I don't agree with the default value as we end up taking wrong path or making wrong decision. It's better to re-try then start etcd with wrong config which can leads to problem like split brain.

> The proper way to solve this is for druid to correctly specify the intent to the etcdbr, saying whether this is a new cluster or an existing cluster, and not have etcdbr try to deduce this by checking the statefulset which is actually hacky. We can do this as part of https://github.com/gardener/etcd-druid/issues/702.

in a single member restoration case, backup-restore itself has to deduce the `ClusterState` of member while providing config and I don't think druid should involve in that kind of scenario.

shreyas-s-rao: > It's better to re-try then start etcd with wrong config which can leads to problem like split brain.

That works for regular, non-autonomous cluster, because upon a retry etcd-backup-restore might be able to get the k8s config and client again. But for autonomous cluster case, it is guaranteed that there is no k8s config, so GetClusterState will always return error. How do we handle this? In that case, druid needs to explicitly tell etcdbr to not look for the k8s client. But if there's no k8s client, how does etcdbr know what cluster state to return? There are still many ambiguities.

ishan16696: > But for autonomous cluster case, it is guaranteed that there is no k8s config, so GetClusterState will always return error.

Currently, backup-restore don't support  autonomous cluster case for HA, so there is no question of creation of k8s client, this PR is for non-HA autonomous cluster.

> There are still many ambiguities.

yes, there are as currently we haven't looked backup-restore from autonomous cluster pov. That will require separate discussion.

shreyas-s-rao: After an out-of-band discussion with @ishan16696 , I realised that if it's a non-HA etcd cluster, then this code is run before creating clientset:
```go
if clusterSize == 1 {
    return miscellaneous.ClusterStateNew, nil
}
```

This means, for a non-HA cluster without a KUBECONFIG configured, the cluster state will be returned as New, and etcd will be able to start. If an HA cluster is configured without KUBECONFIG, then etcd will not be able to start and will retry, which is the current and expected behavior.

shreyas-s-rao (approved):
/lgtm

PR #779, opened 2024-09-23
Ensuring the integrity of full snapshot before uploading it to the object store.

### LINKED ISSUE #778: [Enhancement] Restoration failed due to full-snapshot taken by backup-restore is found to be corrupted or not having a hash
It has been observed that while doing the restoration from full snapshot, backup-restore failed to restore the etcd from full-snapshot. This failure appears to be due to some full snapshots being corrupted or missing a hash.

Logs of backup-restore during restoration.
```
FATA[0002] Failed to restore snapshot: failed to restore from the base snapshot: snapshot missing hash but --skip-hash-check=false 
```

1. One of such case is reported here: https://github.com/gardener/etcd-backup-restore/issues/749 by community user.
2. We have also observed such cases in our internal clusters while doing the recovery from permanent quorum loss for etcd cluster.

Screenshot 2024-09-23 at 10 10 45 AM

ishan16696:
It has been observed that while doing the restoration from full snapshot sometimes backup-restore failed to restore the etcd due to sometimes full snapshots got corrupted or missing a hash.
This PR is try to minimise the occurrence of such scenarios by verifying the integrity of full snapshot before uploading it to the object store.

Fixes #778 

This is how it's done:

Screenshot 2024-09-23 at 10 10 45 AM

**Release note**:

```noteworthy operator
Ensure the integrity of the full snapshot prior to uploading it to the object store, thereby reducing the potential restoration failures.
```

### DIFF
--- a/pkg/compactor/compactor.go
+++ b/pkg/compactor/compactor.go
@@ -173,7 +173,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) 
 	isFinal := compactorRestoreOptions.BaseSnapshot.IsFinal
 
 	cc := &compressor.CompressionConfig{Enabled: isCompressed, CompressionPolicy: compressionPolicy}
-	snapshot, err := etcdutil.TakeAndSaveFullSnapshot(snapshotReqCtx, clientMaintenance, cp.store, etcdRevision, cc, suffix, isFinal, cp.logger)
+	snapshot, err := etcdutil.TakeAndSaveFullSnapshot(snapshotReqCtx, clientMaintenance, cp.store, opts.Config.TempSnapshotsDir, etcdRevision, cc, suffix, isFinal, cp.logger)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/etcdutil/etcdutil.go
+++ b/pkg/etcdutil/etcdutil.go
@@ -6,8 +6,13 @@ package etcdutil
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/tls"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"reflect"
 	"time"
 
 	"github.com/gardener/etcd-backup-restore/pkg/compressor"
@@ -21,6 +26,10 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/pkg/transport"
+)
+
+const (
+	bufferSize = 4 * 1024 * 1024 // 4 MB
 )
 
 // NewFactory returns a Factory that constructs new clients using the supplied ETCD client configuration.
@@ -232,7 +241,7 @@ func GetEtcdEndPointsSorted(ctx context.Context, clientMaintenance client.Mainte
 		endPoint = etcdEndpoints[0]
 	} else {
 		return nil, nil, &errors.EtcdError{
-			Message: fmt.Sprintf("etcd endpoints are not passed correctly"),
+			Message: "etcd endpoints are not passed correctly",
 		}
 	}
 
@@ -254,7 +263,7 @@ func GetEtcdEndPointsSorted(ctx context.Context, clientMaintenance client.Mainte
 }
 
 // TakeAndSaveFullSnapshot takes full snapshot and save it to store
-func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceCloser, store brtypes.SnapStore, lastRevision int64, cc *compressor.CompressionConfig, suffix string, isFinal bool, logger *logrus.Entry) (*brtypes.Snapshot, error) {
+func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceCloser, store brtypes.SnapStore, tempDir string, lastRevision int64, cc *compressor.CompressionConfig, suffix string, isFinal bool, logger *logrus.Entry) (*brtypes.Snapshot, error) {
 	startTime := time.Now()
 	rc, err := client.Snapshot(ctx)
 	if err != nil {
@@ -265,22 +274,29 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
 	timeTaken := time.Since(startTime)
 	logger.Infof("Total time taken by Snapshot API: %f seconds.", timeTaken.Seconds())
 
+	var snapshotData io.ReadCloser
+	snapshotTempDBPath := filepath.Join(tempDir, "db")
+	if snapshotData, err = checkFullSnapshotIntegrity(rc, snapshotTempDBPath, logger); err != nil {
+		logger.Errorf("verification of full snapshot SHA256 hash has been failed: %v", err)
+		return nil, err
+	}
+	logger.Info("full snapshot SHA256 hash has been successfully verified.")
+
 	if cc.Enabled {
 		startTimeCompression := time.Now()
-		rc, err = compressor.CompressSnapshot(rc, cc.CompressionPolicy)
+		snapshotData, err = compressor.CompressSnapshot(snapshotData, cc.CompressionPolicy)
 		if err != nil {
 			return nil, fmt.Errorf("unable to obtain reader for compressed file: %v", err)
 		}
 		timeTakenCompression := time.Since(startTimeCompression)
 		logger.Infof("Total time taken in full snapshot compression: %f seconds.", timeTakenCompression.Seconds())
 	}
-	defer rc.Close()
 
 	logger.Infof("Successfully opened snapshot reader on etcd")
 
 	// Then save the snapshot to the store.
 	snapshot := snapstore.NewSnapshot(brtypes.SnapshotKindFull, 0, lastRevision, suffix, isFinal)
-	if err := store.Save(*snapshot, rc); err != nil {
+	if err := store.Save(*snapshot, snapshotData); err != nil {
 		timeTaken := time.Since(startTime)
 		metrics.SnapshotDurationSeconds.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull, metrics.LabelSucceeded: metrics.ValueSucceededFalse}).Observe(timeTaken.Seconds())
 		return nil, &errors.SnapstoreError{
@@ -292,5 +308,102 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
 	metrics.SnapshotDurationSeconds.With(prometheus.Labels{metrics.LabelKind: brtypes.SnapshotKindFull, metrics.LabelSucceeded: metrics.ValueSucceededTrue}).Observe(timeTaken.Seconds())
 	logger.Infof("Total time to save full snapshot: %f seconds.", timeTaken.Seconds())
 
+	defer func() {
+		if err := os.Remove(snapshotTempDBPath); err != nil {
+			logger.Warnf("failed to remove temporary full snapshot file: %v", err)
+		}
+		if err := snapshotData.Close(); err != nil {
+			logger.Warnf("failed to close snapshot data file: %v", err)
+		}
+	}()
 	return snapshot, nil
+}
+
+// checkFullSnapshotIntegrity verifies the integrity of the full snapshot by comparing
+// the appended SHA256 hash of the full snapshot with the calculated SHA256 hash of the full snapshot data.
+func checkFullSnapshotIntegrity(snapshotData io.ReadCloser, snapTempDBFilePath string, logger *logrus.Entry) (io.ReadCloser, error) {
+	logger.Info("checking the full snapshot integrity with the help of SHA256")
+
+	// If previous temp db file already exist then remove it.
+	if _, err := os.Stat(snapTempDBFilePath); err == nil {
+		if err := os.Remove(snapTempDBFilePath); err != nil {
+			return nil, err
+		}
+	} else if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	db, err := os.OpenFile(snapTempDBFilePath, os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := io.Copy(db, snapshotData); err != nil {
+		return nil, err
+	}
+
+	off, err := db.Seek(0, io.SeekEnd)
+	if err != nil {
+		return nil, err
+	}
+	// 512 is chosen because it's a minimum disk sector size in most systems.
+	hasHash := (off % 512) == sha256.Size
+	if !hasHash {
+		return nil, fmt.Errorf("SHA256 hash seems to be missing from snapshot data")
+	}
+
+	var snapshotLastOffset int64
+	if snapshotLastOffset, err = db.Seek(-sha256.Size, io.SeekEnd); err != nil {
+		return nil, err
+	}
+
+	// get snapshot SHA256 hash
+	sha := make([]byte, sha256.Size)
+	if _, err := db.Read(sha); err != nil {
+		return nil, fmt.Errorf("failed to read SHA256 from snapshot data %v", err)
+	}
+
+	buf := make([]byte, bufferSize)
+	hash := sha256.New()
+
+	// reset the file pointer back to starting
+	currentOffset, err := db.Seek(0, io.SeekStart)
+	if err != nil {
+		return nil, err
+	}
+
+	for currentOffset+bufferSize < snapshotLastOffset {
+		offset, err := db.Read(buf)
+		if err != nil {
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
+			break
+		}
+
+		hash.Write(buf[:offset])
+		if currentOffset, err = db.Seek(0, io.SeekCurrent); err != nil {
+			logger.Warnf("unable to get currentOffset using seek: %v", err)
+			currentOffset += int64(offset)
+		}
+	}
+
+	if currentOffset < snapshotLastOffset {
+		if _, err := db.Read(buf); err != nil {
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
+		}
+
+		hash.Write(buf[:snapshotLastOffset-currentOffset])
+	}
+
+	dbSha := hash.Sum(nil)
+	if !reflect.DeepEqual(sha, dbSha) {
+		return nil, fmt.Errorf("expected SHA256 for full snapshot: %v, got %v", sha, dbSha)
+	}
+
+	// reset the file pointer back to starting
+	if _, err := db.Seek(0, io.SeekStart); err != nil {
+		return nil, err
+	}
+
+	// full-snapshot of database has been successfully verified.
+	return db, nil
 }
--- a/pkg/snapshot/restorer/restorer.go
+++ b/pkg/snapshot/restorer/restorer.go
@@ -233,8 +233,7 @@ func (r *Restorer) makeDB(snapDir string, snap *brtypes.Snapshot, commit int, sk
 	}
 	hasHash := (off % 512) == sha256.Size
 	if !hasHash && !skipHashCheck {
-		err := fmt.Errorf("snapshot missing hash but --skip-hash-check=false")
-		return err
+		return fmt.Errorf("snapshot missing hash but --skip-hash-check=false")
 	}
 
 	if hasHash {
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -362,7 +362,7 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
 		}
 		defer clientMaintenance.Close()
 
-		s, err := etcdutil.TakeAndSaveFullSnapshot(ctx, clientMaintenance, ssr.store, lastRevision, ssr.compressionConfig, compressionSuffix, isFinal, ssr.logger)
+		s, err := etcdutil.TakeAndSaveFullSnapshot(ctx, clientMaintenance, ssr.store, ssr.snapstoreConfig.TempDir, lastRevision, ssr.compressionConfig, compressionSuffix, isFinal, ssr.logger)
 		if err != nil {
 			return nil, err
 		}


ishan16696:
Performance tests have been performed with etcd of `7GB` database size:
```
~ > etcdctl endpoint status --endpoints=http://127.0.0.1:2379 --cluster -w table
+-----------------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
|       ENDPOINT        |        ID        | VERSION | DB SIZE | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS |
+-----------------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
| http://localhost:2379 | 8e9e05c52164694d |   3.5.9 |  7.0 GB |      true |      false |         3 |      35952 |              35952 |        |
+-----------------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
```

Backups taken by backup-restore:
```
25M	Full-00000000-02000001-1726772100.gz
79M	Full-00000000-05532798-1726864342.gz
140M	Full-00000000-09083821-1726946137.gz
185M	Full-00000000-11680572-1727038676.gz
200K	Incr-11680573-11691572-1727038777.gz
200K	Incr-11691573-11702572-1727038778.gz
200K	Incr-11702573-11712987-1727038807.gz
68K	Incr-11712988-11716517-1727068649.gz
```


Then a big full-snapshot is triggered by backup-restore:
```
INFO[0536] Taking scheduled full snapshot for time: 2024-09-23 11:00:00.014743 +0530 IST  actor=snapshotter
{"level":"info","ts":"2024-09-23T11:00:00.038+0530","caller":"clientv3/maintenance.go:200","msg":"opened snapshot stream; downloading"}
INFO[0536] Total time taken by Snapshot API: 0.002619 seconds.  actor=snapshotter
INFO[0536] checking the full snapshot integrity with the help of SHA256  actor=snapshotter
{"level":"info","ts":"2024-09-23T11:00:06.902+0530","caller":"clientv3/maintenance.go:208","msg":"completed snapshot read; closing"}
INFO[0547] full snapshot SHA256 hash has been successfully verified.  actor=snapshotter
INFO[0547] start compressing the snapshot using gzip Compression Policy  actor=compressor
INFO[0547] Total time taken in full snapshot compression: 0.001206 seconds.  actor=snapshotter
INFO[0547] Successfully opened snapshot reader on etcd   actor=snapshotter
INFO[0578] Total written bytes: 6982844448               actor=compressor
INFO[0578] Total time to save full snapshot: 41.202852 seconds.  actor=snapshotter
INFO[0578] Successfully saved full snapshot at: Full-00000000-11716517-1727069410.gz  actor=snapshotter
INFO[0578] Applied watch on etcd from revision: 11716518  actor=snapshotter
```

Then restoration is triggered:
```
INFO[0007] Received start initialization request.        actor=backup-restore-server
INFO[0007] Updating status from New to Progress          actor=backup-restore-server
INFO[0007] Setting status to : 503                       actor=backup-restore-server
INFO[0007] Validation failBelowRevision:                 actor=backup-restore-server
INFO[0007] Validation mode: full                         actor=backup-restore-server
INFO[0007] Checking for data directory structure validity...
INFO[0007] Data directory structure invalid.
INFO[0007] Finding latest set of snapshot to recover from...
INFO[0007] Removing directory(default.etcd.part).
INFO[0007] Restoring from base snapshot: Full-00000000-11716517-1727069410.gz  actor=restorer
INFO[0007] start decompressing the snapshot with gzip compressionPolicy  actor=de-compressor
INFO[0009] Responding to status request with: Progress   actor=backup-restore-server
{"level":"warn","ts":"2024-09-23T13:04:23.060+0530","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"passthrough:///http://localhost:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = \"transport: Error while dialing dial tcp [::1]:2379: connect: connection refused\""}
ERRO[0010] failed to get status of etcd endPoint: http://localhost:2379 with error: context deadline exceeded
INFO[0012] successfully fetched data of base snapshot in 4.938676375 seconds [CompressionPolicy:gzip]  actor=restorer
{"level":"warn","ts":"2024-09-23T13:04:27.060+0530","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"passthrough:///http://localhost:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = \"transport: Error while dialing dial tcp [::1]:2379: connect: connection refused\""}
ERRO[0014] failed to get status of etcd endPoint: http://localhost:2379 with error: context deadline exceeded
INFO[0016] Responding to status request with: Progress   actor=backup-restore-server
{"level":"warn","ts":"2024-09-23T13:04:31.061+0530","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"passthrough:///http://localhost:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = \"transport: Error while dialing dial tcp [::1]:2379: connect: connection refused\""}
ERRO[0018] failed to get status of etcd endPoint: http://localhost:2379 with error: context deadline exceeded
{"level":"info","ts":1727076872.963309,"caller":"membership/cluster.go:392","msg":"added member","cluster-id":"cdf818194e3a8c32","local-member-id":"0","added-peer-id":"8e9e05c52164694d","added-peer-peer-urls":["http://localhost:2380"]}
INFO[0020] No delta snapshots present over base snapshot.  actor=restorer
INFO[0020] Removing directory(default.etcd).
INFO[0020] Successfully restored the etcd data directory.
INFO[0020] Successfully initialized data directory for etcd.  actor=backup-restore-server
```

- [x] Full snapshot is successfully taken with integrity check. ✅ 
- [x] Restoration is successful from full-snapshot ✅ 
- [ ] Snapshot-Compaction is also been successfully tested

pkg/etcdutil/etcdutil.go
```diff
@@ -23,6 +28,10 @@ import (
 	"go.etcd.io/etcd/pkg/transport"
 )
 
+const (
+	bufferSize = 4 * 1024 * 1024 // 4 MB
```
anveshreddy18, line 32:
```suggestion
	hashBufferSize = 4 * 1024 * 1024 // 4 MB
```
Can we name it like something like this so that it's easy to understand what this is used for. Also a comment here explaining something like "buffer size for streaming hash" would be better.

anveshreddy18 (commented):
Thanks for the PR @ishan16696. 

PR looks good, I just have 2 small suggestions. PTAL

pkg/etcdutil/etcdutil.go
```diff
@@ -292,5 +308,102 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+
+	// If previous temp db file already exist then remove it.
+	if _, err := os.Stat(snapTempDBFilePath); err == nil {
+		if err := os.Remove(snapTempDBFilePath); err != nil {
+			return nil, err
+		}
+	} else if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
ishan16696: I have addressed the review comments. PTAL

pkg/etcdutil/etcdutil.go
ishan16696: I have addressed the review comments. PTAL

anveshreddy18 (approved):
I tested it and things are working good. 
LGTM

pkg/etcdutil/etcdutil.go
```diff
@@ -292,5 +308,98 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+		}
+	}
+
+	if currentOffset < snapshotLastOffset {
+		if _, err := db.Read(buf); err != nil {
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
+		}
+
+		hash.Write(buf[:snapshotLastOffset-currentOffset])
+	}
```
renormalize, line 391: If an error does occur in line 386 while reading the file `db` to the buffer `buf`, we're logging an error has occurred (not a warning) - but continuing to write to the hash.

The value `buf` will not have the correct contents for us to do this.

It would be correct to return from here with an error along with an error log. Not continue.

Another suggestion, instead of calculating the number of bytes written into the buffer with `snapshotLastOffset-currentOffset`, we can not ignore the value returned by `db.Read(buf)` - which is exactly the value we want, i.e. the number of byte read. This would be easier to read.
```suggestion
	if currentOffset < snapshotLastOffset {
		if bytesRead, err := db.Read(buf); err != nil {
			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
			return nil, err // expand on this error
		} else {
		    hash.Write(buf[:bytesRead])
		}
	}
```

pkg/etcdutil/etcdutil.go
```diff
@@ -292,5 +308,98 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
+			break
+		}
+
+		hash.Write(buf[:offset])
+		if currentOffset, err = db.Seek(0, io.SeekCurrent); err != nil {
+			logger.Warnf("unable to get currentOffset using seek: %v", err)
+			currentOffset += int64(offset)
+		}
+	}
```
renormalize, line 383: We break out of the loop in line 375 when an error occurs, but this error is not checked for anywhere else in the rest of the function. `hash.Sum()` is invoked after this error occurs. Wouldn't this lead to a miscalculated hash?

The error this function returns finally would be "mismatching hashes" - and the invoker would not be aware that the error actually occurred while reading the database file.

It would be better if we returned out of this loop instead of breaking.

```suggestion
	for currentOffset+hashBufferSize < snapshotLastOffset {
		offset, err := db.Read(buf)
		if err != nil {
			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
			break // return here instead
		}

		hash.Write(buf[:offset])
		if currentOffset, err = db.Seek(0, io.SeekCurrent); err != nil {
			logger.Warnf("unable to get currentOffset using seek: %v", err)
			currentOffset += int64(offset)
		}
	}
```

renormalize (changes requested):
Thanks for this much needed PR @ishan16696!

Just a few comments from my side.

Thanks.

pkg/etcdutil/etcdutil.go
```diff
@@ -292,5 +308,98 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
 	logger.Infof("Total time to save full snapshot: %f seconds.", timeTaken.Seconds())
 
+	defer func() {
+		if err := os.Remove(snapshotTempDBPath); err != nil {
+			logger.Warnf("failed to remove temporary full snapshot file: %v", err)
+		}
+		if err := snapshotData.Close(); err != nil {
+			logger.Warnf("failed to close snapshot data file: %v", err)
+		}
+	}()
```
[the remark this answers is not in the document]
ishan16696: as discussed offline, deferred func is being used here to close the open resources, that's why it's important to put them in deferred func, so that resources get closed in case any failure in taking snapshot.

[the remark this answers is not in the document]
renormalize: Thanks. I'd missed seeing the error checks being done before. Sorry for bothering with this trivial comment.

renormalize, line 318: The `defer func()` as it is at line 311 might not ever get invoked causing resource leakage in case an error occurs anywhere before line 311. So it would be better to break it up into multiple defer calls and invoke as soon as possible.

The defer of `os.Remove()` should be at line 283, since it is established that the file is successfully created after the error check before that.

The defer of `snapshotData.Close()` should be at line 294 since we are writing to `snapshotData` in line 287, and we need to check for an error with the call at that line to ensure that `snapshotData` is not a `nil` and hasn't errored.

renormalize (changes requested):
Another suggestion for better error handling, as discussed.

cmd/compact.go
```diff
@@ -57,7 +57,7 @@ func NewCompactCommand(ctx context.Context) *cobra.Command {
 				CompactorConfig: opts.compactorConfig,
 			}
 
-			snapshot, err := cp.Compact(ctx, compactOptions)
+			snapshot, err := cp.Compact(ctx, compactOptions, opts.snapstoreConfig.TempDir)
```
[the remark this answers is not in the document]
ishan16696: address your review comments. PTAL

pkg/etcdutil/etcdutil.go
ishan16696: address your review comments. PTAL

pkg/etcdutil/etcdutil.go
ishan16696: address your review comments. PTAL

cmd/compact.go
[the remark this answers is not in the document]
renormalize: Thanks.

pkg/compactor/compactor.go
```diff
@@ -55,7 +55,7 @@ func NewCompactor(store brtypes.SnapStore, logger *logrus.Entry, clientSet clien
 }
 
 // Compact is mainly responsible for applying snapshots (full + delta), compacting, drefragmenting, taking the snapshot and saving it sequentially.
-func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions) (*brtypes.Snapshot, error) {
+func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions, snapTempDir string) (*brtypes.Snapshot, error) {
```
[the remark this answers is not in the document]
renormalize: Thanks.

pkg/etcdutil/etcdutil.go
```diff
@@ -265,22 +274,38 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+
 	if cc.Enabled {
 		startTimeCompression := time.Now()
-		rc, err = compressor.CompressSnapshot(rc, cc.CompressionPolicy)
+		snapshotData, err = compressor.CompressSnapshot(snapshotData, cc.CompressionPolicy)
```
renormalize, line 296: If an error occurs here, `snapshotData` will be `nil`. The deferred `snapshotData.Close()` will thus cause a panic.

Although the code says the following error returning block is unreachable, it's better to be be on the safer side.

https://github.com/gardener/etcd-backup-restore/blob/8a3bc2c87cc32dcd4dff9d1d347531cc332638a3/pkg/compressor/compressor.go#L36-L41

I've verified that calling a `Close()` on a `nil` will panic.

```
func main() {
	a := returnNil()
	a.Close()
}

func returnNil() io.ReadCloser {
	return nil
}
```

```
➜  temp go run .
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10269b0bc]

goroutine 1 [running]:
main.main()
        /Users/renormalize/go/src/github.com/gardener/temp/temp/main.go:9 +0x1c
exit status 2
```

renormalize (commented):
Tiny nit, but otherwise the PR looks in great condition.
Thanks.

pkg/etcdutil/etcdutil.go
renormalize: Resolved by [002c481](https://github.com/gardener/etcd-backup-restore/pull/779/commits/002c48192fd0eb97fef1ce46fb0344d4f8acbf27).

renormalize (approved):
Thanks @ishan16696 for filling the gaps in my understanding in validation, and addressing all review comments!

renormalize:
@ishan16696 could you rebase your PR on master? This will fix the failing integration tests.
Thanks.

ishan16696:
> could you rebase your PR on master? This will fix the failing integration tests.

done

seshachalam-yv:
**Benchmark Test Results Comparison Between PR Branch and Master Branch**

I conducted benchmark tests to compare the performance of the `TakeAndSaveFullSnapshot` function between the current PR branch and the master branch.

**Master Branch (`old.txt`):**
```
goos: darwin
goarch: arm64
pkg: github.com/gardener/etcd-backup-restore/pkg/etcdutil
cpu: Apple M1 Max
BenchmarkTakeAndSaveFullSnapshot-10          100     466519650 ns/op    252784125 B/op    29873 allocs/op
PASS
ok      github.com/gardener/etcd-backup-restore/pkg/etcdutil   47.593s
```

**PR Branch (`new.txt`):**
```
goos: darwin
goarch: arm64
pkg: github.com/gardener/etcd-backup-restore/pkg/etcdutil
cpu: Apple M1 Max
BenchmarkTakeAndSaveFullSnapshot-10          100     571449465 ns/op    256778608 B/op    30696 allocs/op
PASS
ok      github.com/gardener/etcd-backup-restore/pkg/etcdutil   58.196s
```

```
                         │     Old      │     New      │    Delta
-------------------------│--------------│--------------│-----------------
Execution Time (sec/op)  │ 466.5ms ± 0% │ 571.4ms ± 0% │ +22.5% (↑)
Memory Usage (B/op)      │ 241.1MiB ± 0%│ 244.9MiB ± 0%│  +1.57% (↑)
Allocations (allocs/op)  │ 29,873 ± 0%  │ 30,696 ± 0%  │  +2.78% (↑)
```
**Conclusion**

There is a 22.5% increase in execution time in the new version. However, since this increase amounts to approximately `105 milliseconds` per operation, it may not have a significant impact on overall performance.

WDYT @ishan16696

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+		hash.Write(buf[:offset])
+		if currentOffset, err = db.Seek(0, io.SeekCurrent); err != nil {
+			logger.Warnf("unable to get currentOffset using seek: %v", err)
+			currentOffset += int64(offset)
+		}
+	}
+
+	if currentOffset < snapshotLastOffset {
+		if _, err := db.Read(buf); err != nil {
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
```
seshachalam-yv, line 390: The error handling for db.Read(buf) does not return the error if it occurs. This could lead to silent failures. Consider returning the error if it is not nil.

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+
+	// If previous temp db file already exist then remove it.
+	if err := os.Remove(snapTempDBFilePath); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	db, err := os.OpenFile(snapTempDBFilePath, os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return nil, err
+	}
```
seshachalam-yv, line 338: Use defer db.Close()

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+	buf := make([]byte, hashBufferSize)
+	hash := sha256.New()
+
+	// reset the file pointer back to starting
+	currentOffset, err := db.Seek(0, io.SeekStart)
+	if err != nil {
+		return nil, err
+	}
+
+	for currentOffset+hashBufferSize < snapshotLastOffset {
```
seshachalam-yv, line 374: Consider checking the condition` currentOffset+hashBufferSize <= snapshotLastOffset` instead of `<` to include the scenario where the remaining data exactly matches the buffer size.

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+		offset, err := db.Read(buf)
+		if err != nil {
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
```
seshachalam-yv, line 377: Instead of logging the error message directly, wrap the error and return it with additional context. This allows the caller to handle and log it appropriately.

ishan16696:
Hi @seshachalam-yv , thanks for performing the benchmark comparison tests.  IMO, `22.5%` is not a very significant increase because of the following reasons:
1. As you have already mentioned that `x milliseconds ` per operation won't have significant impact on overall performance, and moreover etcd's pods usually run on powerful VMs, so I don't think it's impactful.
2. As it ensures the integrity of our full snapshots, so the checks performed will overshadow the performance impact. This is because a full snapshot is typically taken once a day. While community users may configure full snapshots to occur two or three times a day, but still these are not as frequent as delta snapshots. Therefore, a slight performance degradation could be tolerable, given it's not overly frequent.

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+			return nil, err
+		}
+
+		hash.Write(buf[:offset])
+		if currentOffset, err = db.Seek(0, io.SeekCurrent); err != nil {
+			logger.Warnf("unable to get currentOffset using seek: %v", err)
```
seshachalam-yv, line 383: Why do we need to read the currentOffset using `db.Seek`? If an error occurs, we're defaulting to using the offset, which I understand. Could we always use the offset instead? @ishan16696, is there a specific reason for this approach?

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+	if currentOffset < snapshotLastOffset {
+		if _, err := db.Read(buf); err != nil {
+			logger.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
+		}
+
+		hash.Write(buf[:snapshotLastOffset-currentOffset])
+	}
+
+	dbSha := hash.Sum(nil)
+	if !reflect.DeepEqual(sha, dbSha) {
```
seshachalam-yv, line 397: Use `bytes.Equal(sha, dbSha)` instead of `reflect.DeepEqual` for comparing byte slices as it is more idiomatic and potentially more efficient.

pkg/etcdutil/etcdutil.go
```diff
@@ -265,22 +274,40 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
 	timeTaken := time.Since(startTime)
 	logger.Infof("Total time taken by Snapshot API: %f seconds.", timeTaken.Seconds())
 
+	var snapshotData io.ReadCloser
+	snapshotTempDBPath := filepath.Join(tempDir, "db")
```
seshachalam-yv, line 278: Use `os.MkdirTemp` and add `defer os.RemoveAll(snapshotTempDBPath)` next line.

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,88 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
 
 	return snapshot, nil
 }
+
+// checkFullSnapshotIntegrity verifies the integrity of the full snapshot by comparing
+// the appended SHA256 hash of the full snapshot with the calculated SHA256 hash of the full snapshot data.
+func checkFullSnapshotIntegrity(snapshotData io.ReadCloser, snapTempDBFilePath string, logger *logrus.Entry) (io.ReadCloser, error) {
```
seshachalam-yv, line 327: Can you add Unit tests

pkg/etcdutil/etcdutil.go
renormalize: We're returning the `db` so we can't really `Close` it 🤔

ishan16696: yes, we are returning the db, hence we can't close, we are closing the stream in caller.

pkg/etcdutil/etcdutil.go
ishan16696: > This could lead to silent failures.

This won't result in silent failure as we log the error and hash won't match. Anyway I have now change this as requested to return error.

pkg/etcdutil/etcdutil.go
ishan16696: done

pkg/etcdutil/etcdutil.go
ishan16696: yes, we can use the offset also, I just thought to be safe side better to use `db.Seek(0, io.SeekCurrent)` but both are same, I don't have strong opinion on this.

pkg/etcdutil/etcdutil.go
ishan16696: I will check this

pkg/etcdutil/etcdutil.go
ishan16696: It will be difficult to write a unit test for this function:
1. It's taking data stream as input which comes from full snapshot data of snapshot api call which was called on etcd.
2. Actually this function is already getting tested when integration tests triggers a full snapshot which internally calls this function to verify full snapshot integrity.

pkg/etcdutil/etcdutil.go
ishan16696: the `temp` dir already exist, no need to re-create them, we just need to create a db file under this tempDir which function is doing `checkFullSnapshotIntegrity()`

pkg/etcdutil/etcdutil.go
ishan16696: yeah, I have verified, it make sense to put `=`. I have made the change.

pkg/etcdutil/etcdutil.go
seshachalam-yv: I see your point about integration tests, but they might not cover all scenarios for the checkFullSnapshotIntegrity function. 

We could use a dummy snapshot in unit tests to check these scenarios more thoroughly.

pkg/etcdutil/etcdutil.go
ishan16696: I have re-analyse this comment for `urrentOffset+hashBufferSize <= snapshotLastOffset` but this case will never arise as we have taken `hashBufferSize` of `4MB` which is not multiple of `512`.
This case can only be arise incase of hashBufferSize is multiple of `512` as snapshot data is always multiple of `512`.

ishan16696: so, keep the `=` or without `=` won't give us any edge case or won't harm us. both are fine.
Even if we switches the `hashBufferSize` of `512B`, it still works as `if` condition will handle that

seshachalam-yv (changes requested):
The original `TakeAndSaveFullSnapshot` function is broken down into smaller, more focused functions:
	- `createEtcdSnapshot`: Initiates the etcd snapshot process.
	- `verifyFullSnapshotIntegrity`: Verifies the integrity of the full snapshot using SHA256.
	- `validateSnapshotSHA256`: Checks the SHA256 hash appended to the snapshot.
	- `compressSnapshotData`: Compresses the snapshot data if compression is enabled.
	- `saveSnapshotToStore`: Saves the snapshot to the SnapStore.
	- `resetFilePointer`: Resets the file pointer to the beginning of the file.
	- `cleanUpTempFile`: Removes the temporary file used during snapshot verification.


Utilized `bufio.Reader.ReadFull` for reading data and `io.Copy` with `io.LimitReader` for copying data to improve efficiency and handle large files more effectively.

These changes aim to make the code more modular, easier to test, and maintain.

pkg/etcdutil/etcdutil.go
```diff
@@ -254,7 +263,7 @@ func GetEtcdEndPointsSorted(ctx context.Context, clientMaintenance client.Mainte
 }
 
 // TakeAndSaveFullSnapshot takes full snapshot and save it to store
-func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceCloser, store brtypes.SnapStore, lastRevision int64, cc *compressor.CompressionConfig, suffix string, isFinal bool, logger *logrus.Entry) (*brtypes.Snapshot, error) {
+func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceCloser, store brtypes.SnapStore, tempDir string, lastRevision int64, cc *compressor.CompressionConfig, suffix string, isFinal bool, logger *logrus.Entry) (*brtypes.Snapshot, error) {
```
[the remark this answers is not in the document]
ishan16696: I would not like to change the function `TakeAndSaveFullSnapshot()` as this is not a refactoring PR. 
Feel free to open a separate issue if you want to refactor `TakeAndSaveFullSnapshot()`.

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,87 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+	}
+
+	// reset the file pointer back to starting
+	if _, err := db.Seek(0, io.SeekStart); err != nil {
+		return nil, err
+	}
+
+	// full-snapshot of database has been successfully verified.
+	return db, nil
+}
```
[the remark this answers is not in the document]
ishan16696: > // resetFilePointer resets the file pointer to the beginning of the file.
func resetFilePointer(db *os.File) error {
	if _, err := db.Seek(0, io.SeekStart); err != nil {
		return fmt.Errorf("failed to reset file pointer: %w", err)
	}
	return nil
}
// cleanUpTempFile removes the temporary file used during snapshot verification.
func cleanUpTempFile(filePath string, logger *logrus.Entry) {
	if err := os.Remove(filePath); err != nil && !os.IsNotExist(err) {
		logger.Warnf("Failed to remove temporary full snapshot file: %v", err)
	}
}

I wouldn't like to create a separate function for one liner.

pkg/etcdutil/etcdutil.go
ishan16696: there is already a code is present to remove the `temporary: db` file which we created inside `temp` dir.
```
	defer func() {
		if err := os.Remove(snapshotTempDBPath); err != nil {
			logger.Warnf("failed to remove temporary full snapshot file: %v", err)
		}
	}()
```
we don't want to remove the `temp` dir, we just want to remove the `temporary: db` file which we created, hence `os.Remove()` is used.

pkg/etcdutil/etcdutil.go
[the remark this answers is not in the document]
ishan16696: > 	buf := make([]byte, hashBufferSize)
	if _, err := io.CopyBuffer(db, snapshotData, buf); err != nil {
		return nil, fmt.Errorf("failed to copy snapshot data to temporary DB file %s: %v", snapTempDBFilePath, err)
	}
	// Verify SHA256 hash
	if err := validateSnapshotSHA256(db, logger); err != nil {
		return nil, err
	}
	// Reset the file pointer back to the beginning
	if err := resetFilePointer(db); err != nil {
		return nil, err
	}


this is wrong code, this is not how it's working.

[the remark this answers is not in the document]
seshachalam-yv: May I know what you mean by wrong.

pkg/etcdutil/etcdutil.go
seshachalam-yv: Sure, I'm fine with deleting the DB file.

pkg/etcdutil/etcdutil.go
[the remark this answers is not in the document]
ishan16696: as we have to read the data of snapshots in buffer and keep calculating the hash

```go
	buf := make([]byte, hashBufferSize)
	hash := sha256.New()

	// reset the file pointer back to starting
	currentOffset, err := db.Seek(0, io.SeekStart)
	if err != nil {
		return nil, err
	}

	for currentOffset+hashBufferSize <= totalSnapshotBytes {
		offset, err := db.Read(buf)
		if err != nil {
			return nil, fmt.Errorf("unable to read snapshot data into buffer to calculate SHA256: %v", err)
		}

		hash.Write(buf[:offset])
		currentOffset += int64(offset)
	}
```

I see you are using `io.LimitReader` but IMO we want to avoid that as we don't want to load all db data (~8Gi in worst case) into memory, that's why we went with this approach of reading the data in small chunk, so that there won't be any memory spike as you have yourself has verified this: https://github.com/gardener/etcd-backup-restore/pull/779#issuecomment-2425647408

pkg/etcdutil/etcdutil.go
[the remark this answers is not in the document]
seshachalam-yv: Since you are touching this function and adding more 15+ lines, I don't want to grow this function. This won't help readability of code. Thats why I suggest to move the code into smaller functions, each function name clearly conveys the functionality

pkg/etcdutil/etcdutil.go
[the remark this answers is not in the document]
seshachalam-yv: `io.LimitReader` does not load all the data (~8Gi in the worst case) into memory at once. Instead, it ensures that at most `totalSnapshotBytes` are read from the underlying `db`, and the actual memory usage is determined by the buffer size provided to `io.CopyBuffer` (in this case, `hashBufferSize`). This keeps the memory footprint small and controlled while simplifying the code by removing manual offset tracking and loop management.

If there are still concerns about the behavior or specific cases where `io.LimitReader` might not meet expectations, I’d be happy to investigate further. Let me know if you’d like me to profile this or make adjustments!

pkg/etcdutil/etcdutil.go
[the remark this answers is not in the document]
ishan16696: make sense, but I would not like to do that in the same PR for following reasons:
1. That will change the scope of PR from adding feature PR to refactoring PR which will be hard to track after 1year (say) if some bug is found in some part of code.
2. With this refactoring change, it will create a huge overhead of manual testing, performance testing etc like taking snapshots, compression of snapshots, restoration of compressed and uncompressed snapshots + verification of snapshots (feature is being added) all needs to be fully tested.
3. It will completely change the scope of the PR, which will also invalidate the `LGTM'ed` of reviewers those who has already reviewed the PR as it will become a new PR.

That's I don't want to do this in the same PR.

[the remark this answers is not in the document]
seshachalam-yv: **Point 1:** Do you see this as a major refactor? It’s essentially just restructuring a few existing lines into two functions for better readability.

**Point 2:** This point isn’t entirely valid, as we are already planning to test this functionality as part of the release process. This includes verifying the new integrity checks that have been added.

**Point 3:** Whenever we introduce new changes or code, it’s common to make minor adjustments to the surrounding logic for better structure. In this case, it’s simply wrapping the logic into smaller, reusable functions.

That said, this isn’t a significant code change. Let’s get some thoughts from other reviewers as well. 

WDYT, @anveshreddy18 and @renormalize?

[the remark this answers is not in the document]
ishan16696: >Do you see this as a major refactor? I

yes, I do because of point 2
I have mentioned the my reasoning in this comment https://github.com/gardener/etcd-backup-restore/pull/779#discussion_r1865254302, I'm not making this PR as refactoring PR. Please feel free to open a separate issue later if somebody have time they can pick it up.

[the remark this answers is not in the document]
ishan16696: > This point isn’t entirely valid, as we are already planning to test this functionality as part of the release process. This includes verifying the new integrity checks that have been added.

point 2 is totally valid, Do currently we have e2e tests which will do what I have mentioned in point 2 or not ? the answer is no, then it will be create a huge overhead. 

> we are already planning to test this functionality as part of the release process.

Currently in release process nobody do the performance testing ... question is not about future, question is about do we have it now or not ? I don't think so, we have it snf I don't see it coming in near by weeks.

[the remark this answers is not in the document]
seshachalam-yv: It seems your concern is about not adding the functions `createEtcdSnapshot`, `saveSnapshotToStore`, and `compressSnapshotDataIfNeeded` to encapsulate the existing logic. 

I’m okay with that approach. Are you aligned with the other review comments? Let me know if there’s anything else to discuss!

pkg/etcdutil/etcdutil.go
```diff
@@ -294,3 +321,87 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+// the appended SHA256 hash of the full snapshot with the calculated SHA256 hash of the full snapshot data.
+func checkFullSnapshotIntegrity(snapshotData io.ReadCloser, snapTempDBFilePath string, logger *logrus.Entry) (io.ReadCloser, error) {
+	logger.Info("checking the full snapshot integrity with the help of SHA256")
+
+	// If previous temp db file already exist then remove it.
+	if err := os.Remove(snapTempDBFilePath); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	db, err := os.OpenFile(snapTempDBFilePath, os.O_RDWR|os.O_CREATE, 0600)
```
shreyas-s-rao, line 335: Missing `db.Close()` after opening the file. I see that the `db` file is returned as is at the end of the function, so can you please add a note in the docstring for this function for the caller to take care of closing the file once they read it? I don't see a `snapshotData.Close()` in the caller function as well, but just a `os.Remove()` of the file, which is not sufficient, since that can lead to memory leaks.

pkg/etcdutil/etcdutil.go
```diff
@@ -265,22 +274,40 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
-		rc, err = compressor.CompressSnapshot(rc, cc.CompressionPolicy)
+		snapshotData, err = compressor.CompressSnapshot(snapshotData, cc.CompressionPolicy)
 		if err != nil {
 			return nil, fmt.Errorf("unable to obtain reader for compressed file: %v", err)
 		}
 		timeTakenCompression := time.Since(startTimeCompression)
 		logger.Infof("Total time taken in full snapshot compression: %f seconds.", timeTakenCompression.Seconds())
 	}
-	defer rc.Close()
+	defer func() {
```
shreyas-s-rao, line 300: Please move the `defer snapshotData.Close()` towards the top, to be declared as soon as the `snapshotData` object is created.

pkg/etcdutil/etcdutil.go
```diff
@@ -265,22 +274,40 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
 		startTimeCompression := time.Now()
-		rc, err = compressor.CompressSnapshot(rc, cc.CompressionPolicy)
+		snapshotData, err = compressor.CompressSnapshot(snapshotData, cc.CompressionPolicy)
 		if err != nil {
 			return nil, fmt.Errorf("unable to obtain reader for compressed file: %v", err)
 		}
 		timeTakenCompression := time.Since(startTimeCompression)
 		logger.Infof("Total time taken in full snapshot compression: %f seconds.", timeTakenCompression.Seconds())
 	}
-	defer rc.Close()
```
shreyas-s-rao, line 277: I see that `rc` is no longer being closed anywhere in the new code, which can cause memory leak. I would recommend you re-use the `rc` variable rather than creating a new variable named `snapshotData`. And the defer `rc.Close()` should be called towards the beginning of the function, where the `rc` object is first created. Please make this change.

pkg/etcdutil/etcdutil.go
```diff
@@ -265,22 +274,40 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
 	timeTaken := time.Since(startTime)
 	logger.Infof("Total time taken by Snapshot API: %f seconds.", timeTaken.Seconds())
 
+	var snapshotData io.ReadCloser
+	snapshotTempDBPath := filepath.Join(tempDir, "db")
+	if snapshotData, err = checkFullSnapshotIntegrity(rc, snapshotTempDBPath, logger); err != nil {
```
shreyas-s-rao, line 279:
```suggestion
	if rc, err = checkFullSnapshotIntegrity(rc, snapshotTempDBPath, logger); err != nil {
```

pkg/etcdutil/etcdutil.go
shreyas-s-rao, line 266: Adapt the function docstring to reflect the new responsibilities of the function - fetch snapshot from etcd, verify integrity, compress it if enabled, save the snapshot to object storage bucket.

shreyas-s-rao (changes requested):
@ishan16696 thanks for the much needed PR which verifies full snapshot integrity before pushing it to the store. I have a few comments about the handling of the file descriptors and ReadCloser objects in general. PTAL.

I also checked @seshachalam-yv 's [remaining open comment](https://github.com/gardener/etcd-backup-restore/pull/779/files#r1861701895) and I feel the structure suggested by him is better than the existing code, in terms of readability. I would request you to incorporate this comment as well. In terms of refactoring, it's not much of a change, since it just brings certain lines of code within the `TakeAndSaveFullSnapshot` function into smaller functions, and makes it easier to read, review and later modify the code if necessary. Your existing test results should still apply, since the requested refactoring doesn't really change the way the integrity check is done or the number of file descriptors or ReadClosers created to do so.

pkg/etcdutil/etcdutil.go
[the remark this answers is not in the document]
ishan16696: I have included `io. LimitReader ` change. PTAL

pkg/etcdutil/etcdutil.go
ishan16696: there is an issue with that, I made the changes but I have reverted that changes. Let me explain why:
Basically if error occurs inside func `checkFullSnapshotIntegrity` then `snapshotData.Close()` will cause a panic.

example:
```go
func main() {
	a := randomFunc()
	a.Close()
}

func randomFunc() io.ReadCloser {
	return nil
}
```

pkg/etcdutil/etcdutil.go
ishan16696: I have made the changes but due to this I'm reevaluating https://github.com/gardener/etcd-backup-restore/pull/779#discussion_r1930038621 this, I might would like to use `snapshotData`.

ishan16696: ok, I have now renamed it to `rc`, removed the `snapshotData` stream.

seshachalam-yv (changes requested):

example/01-etcd-config.yaml
```diff
@@ -8,10 +8,15 @@ snapshot-count: 75000
...
 listen-client-urls: http://0.0.0.0:2379
-advertise-client-urls: http://0.0.0.0:2379
-initial-advertise-peer-urls: http://0.0.0.0:2380
+advertise-client-urls:
+  etcd-main-0:
+  - http://0.0.0.0:2379
+initial-advertise-peer-urls:
+  etcd-main-0:
+  - http://0.0.0.0:2380
+listen-peer-urls: http://0.0.0.0:2380
```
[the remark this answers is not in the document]
ishan16696: opened a PR: https://github.com/gardener/etcd-backup-restore/pull/833/files, please review it.

pkg/etcdutil/etcdutil.go
```diff
@@ -265,32 +278,132 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
+		return nil, err
+	}
+
+	// Note: db file will be closed by caller function.
+	db, err := os.OpenFile(snapTempDBFilePath, os.O_RDWR|os.O_CREATE, 0600) // #nosec G304 -- this is a trusted file written by etcdbr.
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := io.Copy(db, snapshotData); err != nil {
```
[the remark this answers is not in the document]
ishan16696: that's a good point but I'm not sure will it be significant improvement for not as etcd usually run on powerful VMs and using small buffer (`32KB` in io.copy) or big buffer (`2MB or 4MB` in io. CopyBuffer) has it's pros and cons:
- small buffer will use less memory, and more no. of sys calls.
- big buffer will use more memory and less no. of sys calls.

what about buffer of size `1MB` for both which will reduce the sys calls as well as reduce the memory usage as well ? and we will use your benchmark test to analyse both case, wdyt ?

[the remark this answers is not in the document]
ishan16696: these are the benchmark result:
```bash
 withIOCopy                                                                                                                  

BenchmarkTakeAndSaveFullSnapshot-10    	       1	14372154417 ns/op	16053607536 B/op	 1494035 allocs/op
PASS


withIOCopyBuffer with BufferSize 1MB                                                                                                       

BenchmarkTakeAndSaveFullSnapshot-10    	       1	13873867167 ns/op	16101212656 B/op	 1495436 allocs/op
PASS

withIOCopyBuffer with BufferSize 2MB   
BenchmarkTakeAndSaveFullSnapshot-10    	       1	13262964167 ns/op	16138568776 B/op	 1480503 allocs/op
PASS

withIOCopyBuffer  with BufferSize 4MB          
BenchmarkTakeAndSaveFullSnapshot-10    	       1	12966707125 ns/op	16053292648 B/op	 1480557 allocs/op
PASS
```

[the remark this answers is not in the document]
ishan16696: these are result on running it 2 times:
```
withIOCopy                    
BenchmarkTakeAndSaveFullSnapshot-10    	       1	14222291917 ns/op	16077708912 B/op	 1450602 allocs/op
PASS

withIOCopyBuffer with 1 MB                                     
BenchmarkTakeAndSaveFullSnapshot-10    	       1	13410991000 ns/op	16139721648 B/op	 1490552 allocs/op
PASS

withIOCopyBuffer with 2MB           
BenchmarkTakeAndSaveFullSnapshot-10    	       1	13259688958 ns/op	16119761728 B/op	 1472478 allocs/op
PASS

withIOCopyBuffer with 4MB         
BenchmarkTakeAndSaveFullSnapshot-10    	       1	13524415958 ns/op	16034598544 B/op	 1501451 allocs/op
PASS
```
I don't see a significant improvement for using `io.CopyBuffer` over `io.Copy` as etcd runs on powerful VMs, so I'm leaving it to you as I don't have strong opinion on this.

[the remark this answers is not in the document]
ishan16696: I have made the changes to use `io.CopyBuffer` with `4MB` buffer. PTAL

pkg/etcdutil/etcdutil.go
shreyas-s-rao: You can always check the `rc` whether it is nil, before trying to close it. My point is, if the defer isn't called at the point of creation, it is possible that it never gets executed. And since you are re-using the `rc` variable for both the initial `client.Snapshot(ctx)` as well as `checkFullSnapshotIntegrity(rc, snapshotTempDBPath, logger)`, you are never closing the initial `rc` which contains the snapshot data and was passed to `checkFullSnapshotIntegrity()`. It's not closed inside `checkFullSnapshotIntegrity()` also, so it's never closed anywhere.

So I would recommend that you defer the rc.Close() right after calling `rc, err := client.Snapshot(ctx)`, and use a different variable name for the return value from `checkFullSnapshotIntegrity()`, and defer it's close here as expected.

shreyas-s-rao (changes requested):
@ishan16696 thank you so much for addressing my comments, and also for adding tests. I have one small concern regarding closure of ReadClosers, but other than that, looks good.

pkg/etcdutil/etcdutil.go
ishan16696: initially I did use the different name for the return value from `checkFullSnapshotIntegrity()`, I changed it here https://github.com/gardener/etcd-backup-restore/pull/779#discussion_r1930039590 as it was requested, I have no strong opinion on this, I have made the changes. PTAL

pkg/etcdutil/etcdutil.go
```diff
@@ -285,31 +286,32 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
-		rc, err = compressor.CompressSnapshot(rc, cc.CompressionPolicy)
+		snapshotData, err = compressor.CompressSnapshot(snapshotData, cc.CompressionPolicy)
 		if err != nil {
 			return nil, fmt.Errorf("unable to obtain reader for compressed file: %v", err)
 		}
 	}
 
 	defer func() {
-		if err := rc.Close(); err != nil {
+		if err := snapshotData.Close(); err != nil {
```
shreyas-s-rao, line 306: @ishan16696 This is called only if `cc.enabled` is true, ie, if compression is enabled. If compression is disabled, does `snapshotData.Close()` get called anywhere? Ideally, any readCloser should be defer-closed immediately after it is created. But one issue we have is that `compressor.CompressSnapshot()` also attempts to close the same readCloser. So we'll have multiple attempts to close the same readCloser, which leads to errors. 

The fact that `compressor.CompressSnapshot()` attempts to close the readCloser violates the single-responsibility principle, since the function that created the readCloser should have the sole responsibility of closing it as well.

Now I'm not sure how you would like to handle it. You can change the behavior of `compressor.CompressSnapshot()` to not close the rc, so that you can defer-close it here as soon `snapshotData` gets created. Or, you can retain the behavior of `compressor.CompressSnapshot()`, and instead in the defer func here you can check whether `snapshotData` was already closed before attempting to close it. I'll leave the choice to you.

pkg/etcdutil/etcdutil.go
```diff
@@ -285,31 +286,32 @@ func TakeAndSaveFullSnapshot(ctx context.Context, client client.MaintenanceClose
...
 			logger.Warnf("failed to close snapshot data file: %v", err)
 		}
 	}()
```
shreyas-s-rao, line 309: @ishan16696 can you please move this defer func to above the `if cc.enabled{}` block? Reason being, if there's an err in `compressor.CompressSnapshot()` call, then the code immediately returns, and keeps `snapshotData` open (unclosed).

pkg/etcdutil/etcdutil_test.go
```diff
@@ -0,0 +1,168 @@
...
+		factory = mockfactory.NewMockFactory(ctrl)
+		cm = mockfactory.NewMockMaintenanceCloser(ctrl)
+	})
+
+	Describe("To take Full Snapshot of etcd", func() {
+		var (
+			dummyLastRevision = int64(77)
+		)
+		BeforeEach(func() {
+			factory.EXPECT().NewMaintenance().Return(cm, nil).AnyTimes()
```
seshachalam-yv, line 56: **Is starting an embedded etcd instance necessary?**  
The tests in `etcdutil_test.go` use mocked etcd clients (`mockfactory.MockFactory`, `mockfactory.MockMaintenanceCloser`) and do not interact with a real etcd instance. Additionally, the embedded etcd is immediately stopped and closed in `SynchronizedBeforeSuite`, meaning it is not actually used.

seshachalam-yv (changes requested):

pkg/etcdutil/etcdutil_test.go
ishan16696: I used it to read the `db` file which is basically the snapshot, then calculates its SHA.

ishan16696: you can check these helper functions for more info `func getEtcdDBData()` and `func getCorruptedEtcdDBData()`

pkg/etcdutil/etcdutil.go
ishan16696: As I explained in this comment https://github.com/gardener/etcd-backup-restore/pull/779#discussion_r1930038621 that if there's an err occurs in `compressor.CompressSnapshot()` call, then you can't close the empty stream, and now it's not a resource leak as stream is empty.  Moreover, checking for emptiness for datastream is not very straight forward, it will just make the code complex.

shreyas-s-rao: Not necessary. You need to add defer func(){snapshotData.Close()}() after `checkFullSnapshotIntegrity()` call, but before `if cc.enabled{}` block. 

Once you run `checkFullSnapshotIntegrity()`, you will get and handle the error. If there's no error, then you know that `snapshotData` is populated, and it ca be closed, hence you then defer the closure of `snapshotData`. And then, you will check `if cc.enabled` and run `compressor.CompressSnapshot()`, which currently closes the rc by itself, but ideally shouldn't. And then the new rc returned by CompressSnapshot() needs to be defer-closed. 

So the overall flow will look something like this:
```
rc1, err1 := client.Snapshot()
if err1 != nil {<handle error>}
defer rc1.Close()

snapshotTempDBPath := filepath.Join(tempDir, "db")
os.Mkdir(snapshotTempDBPath)
defer os.Remove(snapshotTempDBPath)

rc2, err2 := checkFullSnapshotIntegrity(rc1, ...)
if err2 != nil {<handle error>}
defer rc2.Close()

var (
    snapshot Snapshot
    err4 error
)

if cc.enabled {
    rc3, err3 := compressor.CompressSnapshot(rc2)
    if err3 != nil {<handle error>}
    defer rc3.Close()

    return saveSnapshotToStore(rc3)
}

return snapshot, err4 = saveSnapshotToStore(rc2)
```

shreyas-s-rao: This makes it much easier to read. I don't like re-using the `snapshotData` variable to re-write the new compressed snapshot reader into it, since it makes closure of the old closer tricky.

ishan16696: this will make code more complex, I'm thinking to add this above `if cc.enabled{..}` , wdyt ?
```go
defer func() {
		if snapshotData != nil {
			if err := snapshotData.Close(); err != nil {
				logger.Warnf("failed to close snapshot data file: %v", err)
			}
		}
	}()
```

shreyas-s-rao: Ok, but please add a comment here that you are doing the nil check only because we re-use the `snapshotData` variable, and it can possibly be set to `nil` when it is re-set by the upcoming `CompressSnapshot()` function call.

ishan16696: I have now pushed the changes, PTAL

shreyas-s-rao (approved):
@ishan16696 thanks for making the requested change. 
/lgtm

seshachalam-yv (approved):
Thanks for addressing my review comments 
/lgtm

PR #780, opened 2024-09-23
Enhance the `docker-build` make target to build for multiple platforms.

renormalize:

* The `docker-build` make target builds the image for the architecture of the host machine (same behavior as before).

* The platform can be overridden by passing this variable directly to the make command like `PLATFORM=linux/arm64 make docker-build`, `PLATFORM=linux/amd64 make docker-build`.

**Release note**:

```improvement developer
Builds for non-native platforms can now be done using the `docker-build` make target instead of having to invoke the `docker buildx` command. The platform can be specified using the `PLATFORM` variable which is passed while invoking make.
```


seshachalam-yv (approved):
/lgtm

PR #788, opened 2024-10-04
Backup-restore now restart the etcd member incase of etcd's advertise peerURL found to be updated.

### LINKED ISSUE #787: [Enhancement] Backup-restore should take care of etcd member restart incase of etcd's advertise  peerURL found to be updated.
1. Backup-restpre should restart it's corresponding etcd member when etcd's advertise peer URLs found to be updated as it's mention in the etcd documentation that updating peerURLs necessitates restarts of member for the changes to properly propagate throughout the cluster, as it is a cluster-wide operation. More details can be found in the etcd documentation: https://etcd.io/docs/v3.5/op-guide/runtime-configuration/#update-advertise-peer-urls
2. Backup-restore should only call [updateMemberPeerURL](https://github.com/etcd-io/etcd/blob/fe7d1904d256f9eef66244b67069e6e87e08a816/clientv3/cluster.go#L49) API call only when peerURL is found to be updated, currently that's not the case.

Transition from non-TLS to TLS for PeerURLs is not supported for already deployed etcd HA cluster: https://github.com/gardener/etcd-druid/issues/886

cc @unmarshall

ishan16696:
1. Backup-restore will now call [updateMemberPeerURL](https://github.com/etcd-io/etcd/blob/fe7d1904d256f9eef66244b67069e6e87e08a816/clientv3/cluster.go#L49) API call only when peerURL is found to be updated.
2. Backup-restpre will now restart it's corresponding etcd member after updating etcd's advertise peer URLs  if found to be updated.
3. Introduced a CLI flag `--use-etcd-wrapper`(default: false)  to enable/disable backup-restore to use etcd-wrapper related functionality.

Fixes #787 

This PR needs to be tested  with PR of etcd-wrapper: https://github.com/gardener/etcd-wrapper/pull/31 where `/stop` endpoint is introduced with `--use-etcd-wrapper` set to `true` for backup-restore.

**Release note**:

```noteworthy operator
etcd-backup-restore now triggers a restart of the etcd member after updating etcd's advertise peer URLs if found updated.
```

```noteworthy user
Introduced a CLI flag `--use-etcd-wrapper` (default: false) to enable/disable the backup-restore to use etcd-wrapper related functionality. Note: enable this flag only if etcd-wrapper is deployed.
```

### DIFF
--- a/pkg/server/backuprestoreserver.go
+++ b/pkg/server/backuprestoreserver.go
@@ -7,12 +7,13 @@ package server
 import (
 	"context"
 	"fmt"
-	"github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"net/http"
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/gardener/etcd-backup-restore/pkg/backoff"
 	"github.com/gardener/etcd-backup-restore/pkg/defragmentor"
@@ -219,19 +220,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 		return err
 	}
 
-	m := member.NewMemberControl(b.config.EtcdConnectionConfig)
-	if err := retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
-		cli, err := etcdutil.NewFactory(*b.config.EtcdConnectionConfig).NewCluster()
-		if err != nil {
-			return err
-		}
-		defer cli.Close()
-
-		if err := m.UpdateMemberPeerURL(ctx, cli); err != nil {
-			return err
-		}
-		return nil
-	}); err != nil {
+	if err := b.updatePeerURLIfChanged(ctx, handler.EnableTLS); err != nil {
 		b.logger.Errorf("failed to update member peer url: %v", err)
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/server/backuprestoreserver.go
```diff
@@ -328,6 +319,45 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+		if err = retry.OnError(retry.DefaultBackoff, errors.IsErrNotNil, func() error {
+			if err = m.UpdateMemberPeerURL(ctx, cli); err != nil {
+				return err
+			}
+			b.logger.Info("Successfully updated the peerURLs for etcd member.")
+			return nil
+		}); err != nil {
+			return err
+		}
+		if err := miscellaneous.RestartEtcdWrapper(ctx, tlsEnabled, b.config.EtcdConnectionConfig); err != nil {
```
unmarshall, line 352: do you wish to also retry this operation? If the restart does not succeed then some TCs will not pass as we have seen.

ishan16696: TBH, calling a `stop` endpoint within the same network shouldn't fail in the first attempt itself as it's not a api call, so if it somehow fails then I don't think retry will help as we already exiting the container and restarting it.
Moreover, I also don't want to add a retry everywhere, I want to a retry backoff mechanism where ever necessary.

unmarshall:
@ishan16696 What is the impact of calling `etcd-wrapper` endpoint from etcdbr on consumers who are only using etcdbr?

ishan16696:
> What is the impact of calling etcd-wrapper endpoint from etcdbr on consumers who are only using etcdbr ?

yes, that's very valid point. 
I would suggest to introduce a `/version` endpoint to wrapper just like etcd has `version` endpoint to detect whether corresponding backup-restore is running with wrapper or not.
```
curl http://127.0.0.1:2379/version
{"etcdserver":"3.5.9","etcdcluster":"3.5.0"}
```

backup-restore can send REST request to this URL of wrapper: `http://127.0.0.1:9095/version` 
if response return this `{"etcdwrapper":"v0.x."}` then only backup-restore calls `/stop` endpoint in case of detection of peerURL change
else
backup-restore will try to call etcd endpoint `http://127.0.0.1:2379/version` then just print warning of `unable to restart as etcd-wrapper is not found` . wdyt ?

unmarshall:
>backup-restore will try to call etcd endpoint http://127.0.0.1:2379/version then just print warning of unable to restart as etcd-wrapper is not found . wdyt ?

I am not sure versioning of HTTP endpoints is the right approach. Instead one should start etcdbr with feature gate which should enable/disable wrapper handling.

ishan16696:
> Instead one should start etcdbr with feature gate which should enable/disable wrapper handling.

we already have a feature gate in druid as druid deploys etcd with either with `etcd-custom-image` or `etcd-wrapper` but IMO, backup-restore should not use feature gate for etcd

ishan16696:
@unmarshall @shreyas-s-rao and I have decided to introduced the CLI flag: `--use-etcd-wrapper` to consume etcd-wrapper related functionality in backup-restore with default value of false, hence causing no impact on our community users. We can set `--use-etcd-wrapper` to `true` in etcd-druid.

unmarshall (approved):
We have decided to make this backward-incompatible change. This has been added to the compatibility matrix in etcd-druid documentation.
/lgtm

shreyas-s-rao (approved):
Thanks for addressing the comment. 
/lgtm

PR #790, opened 2024-10-07
Use the standard Go distribution image for integration tests in the pipeline instead of the testmachinery image.

renormalize:

Use the standard Go distribution image for integration tests in the pipeline instead of the testmachinery image.

This will improve build reproducibility for older branches, since it specifies a version instead of a `latest` or `stable` tag.

**Release note**:

```improvement developer
NONE
```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -57,7 +57,7 @@ etcd-backup-restore:
       unit_test:
         image: 'golang:1.23.2'
       integration_test:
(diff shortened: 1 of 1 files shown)


renormalize:
Integration tests won't pass for this PR since the image that will be used for the test is specified in `master`; and this PR updates that image.

Same issue as any other update to definitions in the `.ci` directory.

Will merge once all other tests pass.

cc @seshachalam-yv

seshachalam-yv (approved):
Thanks @renormalize for fixing it 
/lgtm

PR #793, opened 2024-10-16
Improved error handling in deletion of delta snapshots during snapshot garbage collection

### LINKED ISSUE #652: [Enhancement] Error Handling in Delta Snapshot Deletion Process

Improve the error handling mechanism in the delta snapshot deletion process [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/snapshotter/garbagecollector.go#L226-L230). The proposed changes include:
1. Continuing with the deletion of the rest of the "deletable" delta snapshots even when an error is encountered, instead of halting the process.
2. Concatenating all errors encountered during the deletion process using `errors.Join()`, and returning this concatenated error at the end of the process.

>__Note__: This enhancement is based on the review comment in this https://github.com/gardener/etcd-backup-restore/pull/640#discussion_r1271316167.

The [current implementation](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/snapshotter/garbagecollector.go#L226-L230) stops the garbage collection process of delta snapshots if there's an error with the deletion of a particular snapshot. This could be due to an intermittent error or an issue specific to a single snapshot. The proposed enhancement will ensure that a one-off error doesn't halt the entire process, while still effectively communicating to the function caller that there were errors in deleting delta snapshots.

1. Modify the deletion process to continue deleting the remaining snapshots even when an error is encountered.
2. Use the `errors.Join()` function to concatenate all errors encountered during the process.

Shreyas-s14:
This PR changes the error handling for deletion of delta snapshots by replacing the current implementation with concatenation of errors using the errors.Join() method upto a threshold of 5. The process of continues with the deletion of the rest of the "deletable" delta snapshots even when an error is encountered, instead of halting the process. When the number of errors becomes greater than the threshold, it errors out for the current cycle.

Fixes #652 

**Release note**:

```other operator
Improved error handling in deletion of delta snapshots during snapshot garbage collection.
```

### DIFF
--- a/pkg/snapshot/snapshotter/garbagecollector.go
+++ b/pkg/snapshot/snapshotter/garbagecollector.go
@@ -6,7 +6,6 @@ package snapshotter
 
 import (
 	"errors"
-	"fmt"
 	"math"
 	"path"
 	"time"
@@ -72,7 +71,6 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
 			}
 
 			fullSnapshotIndexList := getFullSnapshotIndexList(snapList)
-			fmt.Println(fullSnapshotIndexList)
 			switch ssr.config.GarbageCollectionPolicy {
 			case brtypes.GarbageCollectionPolicyExponential:
 				// Overall policy:
--- a/pkg/snapshot/snapshotter/garbagecollector.go
+++ b/pkg/snapshot/snapshotter/garbagecollector.go
@@ -86,14 +86,15 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
 					eod          = now.Truncate(24 * time.Hour).Add(23 * time.Hour).Add(59 * time.Minute).Add(59 * time.Second)
 					trackingWeek = 0
 				)
+				// snapStream indicates the list of snapshot, where first snapshot is base/full snapshot followed by list of incremental snapshots based on it.
 				// Here we start processing from second last snapstream, because we want to keep last snapstream
 				// including delta snapshots in it.
-				for snapStreamIndex := len(fullSnapshotIndexList) - 1; snapStreamIndex > 0; snapStreamIndex-- {
-					snap := snapList[fullSnapshotIndexList[snapStreamIndex]]
-					nextSnap := snapList[fullSnapshotIndexList[snapStreamIndex-1]]
+				for fullSnapshotIndex := len(fullSnapshotIndexList) - 1; fullSnapshotIndex > 0; fullSnapshotIndex-- {
+					snap := snapList[fullSnapshotIndexList[fullSnapshotIndex]]
+					nextSnap := snapList[fullSnapshotIndexList[fullSnapshotIndex-1]]
 
 					// garbage collect delta snapshots.
-					deletedSnap, err := ssr.GarbageCollectDeltaSnapshots(snapList[fullSnapshotIndexList[snapStreamIndex-1]:fullSnapshotIndexList[snapStreamIndex]])
+					deletedSnap, err := ssr.GarbageCollectDeltaSnapshots(snapList[fullSnapshotIndexList[fullSnapshotIndex-1]:fullSnapshotIndexList[fullSnapshotIndex]])
 					total += deletedSnap
 					if err != nil {
 						continue
@@ -162,14 +163,14 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
 			case brtypes.GarbageCollectionPolicyLimitBased:
 				// Delete delta snapshots in all snapStream but the latest one.
 				// Delete all snapshots beyond limit set by ssr.maxBackups.
-				for snapStreamIndex := 0; snapStreamIndex < len(fullSnapshotIndexList)-1; snapStreamIndex++ {
-					deletedSnap, err := ssr.GarbageCollectDeltaSnapshots(snapList[fullSnapshotIndexList[snapStreamIndex]:fullSnapshotIndexList[snapStreamIndex+1]])
+				for fullSnapshotIndex := 0; fullSnapshotIndex < len(fullSnapshotIndexList)-1; fullSnapshotIndex++ {
+					deletedSnap, err := ssr.GarbageCollectDeltaSnapshots(snapList[fullSnapshotIndexList[fullSnapshotIndex]:fullSnapshotIndexList[fullSnapshotIndex+1]])
 					total += deletedSnap
 					if err != nil {
 						continue
 					}
-					if snapStreamIndex < len(fullSnapshotIndexList)-int(ssr.config.MaxBackups) {
-						snap := snapList[fullSnapshotIndexList[snapStreamIndex]]
+					if fullSnapshotIndex < len(fullSnapshotIndexList)-int(ssr.config.MaxBackups) {
+						snap := snapList[fullSnapshotIndexList[fullSnapshotIndex]]
 						snapPath := path.Join(snap.SnapDir, snap.SnapName)
 						ssr.logger.Infof("GC: Deleting old full snapshot: %s", snapPath)
 						if err := ssr.store.Delete(*snap); err != nil {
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-1:
Thank you @Shreyas-s14 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -528,6 +528,58 @@ var _ = Describe("Snapshotter", func() {
...
+
+						ssr, err := NewSnapshotter(logger, snapshotterConfig, store, etcdConnectionConfig, compressionConfig, healthConfig, snapstoreConfig)
+						Expect(err).ShouldNot(HaveOccurred())
+
+
+						// Threshold is set as 5
+						for i := len(list)-3 ; i > len(list)-9 ; i-- { // first 2 are deleted and next 5 return error while deleting
+							err = os.Remove(path.Join(list[i].Prefix, list[i].SnapName))
+							Expect(err).ShouldNot(HaveOccurred())
+						}
```
anveshreddy18, line 575: Can the comment here be made better, when I first read it I felt the initial removal of objects itself errors out, but it should error out in the `ssr.GarbageCollectDeltaSnapshots()`. So moving this comment there would be better. I leave it to you, but it would be better to comment closer to the actual functionality.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -528,6 +528,58 @@ var _ = Describe("Snapshotter", func() {
...
+						err = os.Remove(path.Join(list[3].Prefix, list[3].SnapName))
+						Expect(err).ShouldNot(HaveOccurred())
+						
+						err = os.Remove(path.Join(list[2].Prefix, list[2].SnapName))
+						Expect(err).ShouldNot(HaveOccurred())
+
+						deleted, err := ssr.GarbageCollectDeltaSnapshots(list)
+						Expect(deleted).ToNot(BeZero())
+						Expect(deleted).Should(Equal(6))
+						Expect(err).ToNot(BeNil())
```
anveshreddy18, line 557:
```suggestion
						Expect(err).To(HaveOccurred())
```

This is to be consistent across the testcases, refererence: the other test case you've added.

anveshreddy18 (changes requested):
Thanks for the PR @Shreyas-s14. Have few suggestions. PTAL

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -528,6 +528,51 @@ var _ = Describe("Snapshotter", func() {
 						Expect(len(list)).Should(Equal(3))
 					})
 				})
```
anveshreddy18, line 530: You can also add a positive test case when there is no error in deleting snapshots.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -528,6 +528,51 @@ var _ = Describe("Snapshotter", func() {
...
+							err := os.Remove(path.Join(list[i].Prefix, list[i].SnapName))
+							Expect(err).ShouldNot(HaveOccurred())
+						}
+
+						deleted, err := ssr.GarbageCollectDeltaSnapshots(list)
+						Expect(deleted).Should(Equal(6))
+						Expect(err).Should(HaveOccurred())
+					})
+				})
+				Context("When the number of errors while deleting are greater than the threshold", func() {
```
anveshreddy18, line 553:
```suggestion
				Context("When the number of errors while deleting are greater than or equal to threshold", func() {
```

pkg/snapshot/snapshotter/garbagecollector.go
```diff
@@ -85,14 +86,15 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
 					eod          = now.Truncate(24 * time.Hour).Add(23 * time.Hour).Add(59 * time.Minute).Add(59 * time.Second)
 					trackingWeek = 0
 				)
+				// snapStream indicates the list of snapshot, where first snapshot is base/full snapshot followed by list of incremental snapshots based on it.
```
anveshreddy18, line 89: You can move this comment to above as `snapStream` is used there but defined here.

anveshreddy18 (changes requested):

anveshreddy18 (approved):
Thanks for the PR @Shreyas-s14. Things are looking good. 
LGTM !!

pkg/snapshot/snapshotter/garbagecollector.go
```diff
@@ -161,14 +163,14 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
 			case brtypes.GarbageCollectionPolicyLimitBased:
 				// Delete delta snapshots in all snapStream but the latest one.
 				// Delete all snapshots beyond limit set by ssr.maxBackups.
-				for snapStreamIndex := 0; snapStreamIndex < len(snapStreamIndexList)-1; snapStreamIndex++ {
-					deletedSnap, err := ssr.GarbageCollectDeltaSnapshots(snapList[snapStreamIndexList[snapStreamIndex]:snapStreamIndexList[snapStreamIndex+1]])
+				for fullSnapshotIndex := 0; fullSnapshotIndex < len(fullSnapshotIndexList)-1; fullSnapshotIndex++ {
+					deletedSnap, err := ssr.GarbageCollectDeltaSnapshots(snapList[fullSnapshotIndexList[fullSnapshotIndex]:fullSnapshotIndexList[fullSnapshotIndex+1]])
```
renormalize, line 167: https://github.com/gardener/etcd-backup-restore/pull/793#discussion_r1818848135

pkg/snapshot/snapshotter/garbagecollector.go
```diff
@@ -68,8 +70,8 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
 				ssr.logger.Infof("GC: Total number garbage collected chunks: %d", chunksDeleted)
 			}
 
-			snapStreamIndexList := getSnapStreamIndexList(snapList)
-
+			fullSnapshotIndexList := getFullSnapshotIndexList(snapList)
+			// snapStream indicates the list of snapshot, where first snapshot is base/full snapshot followed by list of incremental snapshots based on it.
```
renormalize, line 74:
```suggestion
			// snapStream indicates a list of snapshots, where the first snapshot is base/full snapshot followed by a list of incremental snapshots based on it. Garbage collection is performed on one snapStream at a time.
```
Agree with @anveshreddy18's [review comment](https://github.com/gardener/etcd-backup-restore/pull/793/files#r1805250327), but I don't think the code comment is currently very useful in its state.

I don't particularly like that currently there is explanation about what `snapStream` is, but there is no `snapStream` token in the function where the explanation is present.
There is no variable/type in the function with `snapStream` in it. Unless someone searches for that word in the rest of the file, the comment causes more confusion.

See the suggestion in https://github.com/gardener/etcd-backup-restore/pull/793#discussion_r1818848135, would make the above suggested code read better.

renormalize (changes requested):
Thanks for the PR @Shreyas-s14!
I'd like to suggest a few changes.

pkg/snapshot/snapshotter/snapshotter_test.go
```diff
@@ -576,8 +576,7 @@ var _ = Describe("Snapshotter", func() {
 						Expect(err).ShouldNot(HaveOccurred())
 
 
-						// Threshold is set as 5
-						for i := len(list)-3 ; i > len(list)-9 ; i-- { // 5 snapshots from the list of snapshots to be deleted, are removed before passing it to the function ssr.GarbageCollectDeltaSnapshots. This will cause an error when the function tries to delete them.
+						for i := len(list)-3 ; i > len(list)- 3 - ErrorThreshold-1 ; i-- { // 5 snapshots from the list of snapshots to be deleted, are removed before passing it to the function ssr.GarbageCollectDeltaSnapshots. This will cause an error when the function tries to delete them.
```
renormalize, line 579: - The formatting looks odd in the loop condition. 
Do you have `gofmt` configured? Please run `gofmt`.
- Can you move the comment above the for loop? It's quite hard to follow with so many characters in one line.
- Update the comment to not mention `5 snapshots from the list of snapshots`. Mention the threshold.
- Mention that the function will error when the number of individual errors seen meets the threshold set.

renormalize (changes requested):

pkg/snapshot/snapshotter/garbagecollector.go
```diff
@@ -89,13 +90,16 @@ func (ssr *Snapshotter) RunGarbageCollector(stopCh <-chan struct{}) {
...
 				// Here we start processing from second last snapstream, because we want to keep last snapstream
 				// including delta snapshots in it.
-				for fullSnapshotIndex := len(fullSnapshotIndexList) - 1; fullSnapshotIndex > 0; fullSnapshotIndex-- {
-					snap := snapList[fullSnapshotIndexList[fullSnapshotIndex]]
-					nextSnap := snapList[fullSnapshotIndexList[fullSnapshotIndex-1]]
+				for fullSnapshotIndex := 0; fullSnapshotIndex < len(fullSnapshotIndexList) - 1; fullSnapshotIndex++ {
+					descIndex := len(fullSnapshotIndexList) - 1 - fullSnapshotIndex
+    				nextDescIndex := descIndex - 1
+					snap := snapList[fullSnapshotIndexList[descIndex]]
+					nextSnap := snapList[fullSnapshotIndexList[nextDescIndex]]
```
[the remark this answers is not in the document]
renormalize: The tests are not written in a way which is agnostic to the order in which the loop is run, oddly. When the order is changed of the loop, the tests fail.

We've decided to correct this later.

pkg/snapshot/snapshotter/garbagecollector.go
```diff
@@ -5,6 +5,8 @@
 package snapshotter
 
 import (
+	"errors"
+	// "fmt"
```
renormalize, line 9: This commented import can be removed.

pkg/snapshot/snapshotter/garbagecollector.go
```diff
@@ -16,6 +18,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const DeltaSnapshotGCErrorThreshold = 5
```
renormalize, line 21: A comment has to be added to the exported `const` for `check` to pass in CI. Please add a comment.

renormalize (changes requested):

renormalize (approved):

PR #799, opened 2024-11-11
Minimise the impact of flaky unit tests in the `defragmentor` pkg by making it fail fast

### LINKED ISSUE #798: [Enhancement] Minimise the impact of flaky unit tests in the `defragmentor` pkg by making it fail fast

I would like to minimise the impact of the `pkg/defragmentor` package unit tests failure in concourse pipeline caused by a flaky test. While a proper solution is needed to fix the flaky test, a simple solution could be to just run the `defragmentor` pkg first when the unit tests are run. This way it will fail fast and the PR owner or other operators can trigger the new build. This will save a lot of time for everyone involved in the PR. 

To reduce the waiting time to see the error in case of a flaky test failure in the `defragmentor` suite. With the randomisation of test suites in the current setup, this flaky test could end up running near the end and failing. This only causes delays in review or getting the PR merged.

To provide a dirty solution for now, which I want to quickly provide as I had problems with this. The proper solution would be to fix the flakiness, which may be dealt with later in a separate PR whenever we have the capacity.

anveshreddy18:

This PR attempts to run the `pkg/defragmentor` unit test pkg before running remaining packages, this is done to reduce the waiting time to see the error & retry tests in case of a flaky test failure in the `defragmentor` suite. With the randomisation of test suites in the current setup, this flaky test could end up running near the end and failing. This only causes delays in review or getting the PR merged.

Fixes #798 

This is just a dirty fix for now, a proper solution would be to fix the flaky tests. We can remove this extra handling once that gets fixed. 

**Release note**:

```improvement operator
run the `pkg/defragmentor` test package first before running remaining pkgs, to fail-fast incase of flaky tests
```

### DIFF
--- a/.ci/unit_test
+++ b/.ci/unit_test
@@ -28,31 +28,76 @@ test_with_coverage() {
   local output_dir=test/output
   local coverprofile_file=coverprofile.out
   mkdir -p "${output_dir}"
+
   ginkgo "${GINKGO_COMMON_FLAGS[@]}" -gcflags=all=-d=checkptr=0 --coverprofile "${coverprofile_file}" -covermode=set -outputdir "${output_dir}" ${TEST_PACKAGES}
   sed -i='' '/mode: set/d' "${output_dir}/${coverprofile_file}"
   { echo "mode: set"; cat "${output_dir}/${coverprofile_file}"; } > "${output_dir}/${coverprofile_file}.temp"
   mv "${output_dir}/${coverprofile_file}.temp" "${output_dir}/${coverprofile_file}"
+
   go tool cover -func "${output_dir}/${coverprofile_file}"
 }
 
 ################################################################################
 
 # To run a specific package, run TEST_PACKAGES=<PATH_TO_PACKAGE> make test
-TEST_PACKAGES="${TEST_PACKAGES:-"cmd pkg"}"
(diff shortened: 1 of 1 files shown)


.ci/unit_test
```diff
@@ -28,31 +28,76 @@ test_with_coverage() {
...
+./pkg/health/membergarbagecollector \
+./pkg/initializer/validator \
+./pkg/leaderelection \
+./pkg/member \
+./pkg/metrics \
+./pkg/miscellaneous \
+./pkg/server \
+./pkg/snapshot/copier \
+./pkg/snapshot/restorer \
+./pkg/snapshot/snapshotter \
```
renormalize, line 58: Multiple packages can be run using `./pkg/snapshot/...` instead of specifying the individual packages.

renormalize (approved):
Thank you for making our lives incredibly easier!

.ci/unit_test
anveshreddy18: Done. Thanks

renormalize:
@anveshreddy18 feel free to merge this whenever you have the time.

ishan16696 (approved):

anveshreddy18:
> no need to add a release note for this PR.

Sure, removed it.

PR #804, opened 2024-11-21
Prevent skipping scheduled full snapshots when there are no new updates

### LINKED ISSUE #797: [Enhancement] Don't skip the schedule full snapshot incase of no new updates
Sometimes, backup-restore skips a scheduled full snapshot as no new updates have been made to etcd since the last full snapshot.
https://github.com/gardener/etcd-backup-restore/blob/61bf7a0239c03680e59ef8241717fbb32621163a/pkg/snapshot/snapshotter/snapshotter.go#L348-L349
This behavior should be modified to ensure that full snapshots are not skipped as it might result in violation of SLA as well as prometheus alert of not taking full-snapshot since last scheduled one.

There have been instances where no new events have occurred since the last full snapshot was taken. As a result, when a new scheduled full snapshot is triggered, it is skipped due to the absence of new updates in etcd. While this is functionally correct, it can cause issues with our monitoring. Prometheus checks the timestamp of the last full snapshot and if it is more than 24 hours old, it is considered a violation of the SLA, triggering an alert. This situation can be avoided if the backup-restore process is adjusted to trigger a full snapshot regardless of whether there are new updates or not. This will ensure a scheduled full snapshot is always taken, preventing potential SLA violations.

Logs of backup-restore
```
time="2024-11-05T02:17:07Z" level=info msg="Starting snapshotter..." actor=backup-restore-server
time="2024-11-05T02:17:07Z" level=info msg="Starting the Snapshot EventHandler." actor=snapshotter
time="2024-11-05T02:17:07Z" level=info msg="Taking scheduled full snapshot for time: 2024-11-05 02:17:07.800297004 +0000 UTC" actor=snapshotter
time="2024-11-05T02:17:07Z" level=info msg="Starting the FullSnapshot lease renewal with interval 1m0s" actor=FullSnapLeaseUpdater
time="2024-11-05T02:17:07Z" level=info msg="There are no updates since last snapshot, skipping full snapshot." actor=snapshotter
time="2024-11-05T02:17:07Z" level=info msg="Applied watch on etcd from revision: 410470578" actor=snapshotter
time="2024-11-05T02:17:07Z" level=info msg="Stopping full snapshot..." actor=snapshotter
time="2024-11-05T02:17:07Z" level=info msg="Resetting full snapshot to run after 16h46m52.143914121s" actor=snapshotter
time="2024-11-05T02:17:07Z" level=info msg="Will take next full snapshot at time: 2024-11-05 19:04:00 +0000 UTC" actor=snapshotter
```

Shreyas-s14:
This PR ensures that the backup-restore process does not skip a scheduled full snapshot, even when no events occur during the interval between two full snapshot schedules.

Fixes #797 

The current implementation allows for taking a full snapshot in all cases, except the case where the preceding full snapshot has been set as final.

**Release note**:

```improvement operator
Don't skip full snapshot, always trigger a full snapshot independent of new updates to etcd, preventing prometheus alerts of not taking a scheduled full snapshot.
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -344,8 +344,8 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
 		}
 	}
 	lastRevision := resp.Header.Revision
-	if ssr.PrevSnapshot.IsFinal == true {
-		ssr.logger.Infof("Previous snapshot was set as final, skipping full snapshot.")
+	if isFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal{
+		ssr.logger.Infof("There are no updates since final last snapshot, skipping final full snapshot.")
 	} else {
 		// Note: As FullSnapshot size can be very large, so to avoid context timeout use "SnapshotTimeout" in context.WithTimeout()
 		ctx, cancel = context.WithTimeout(context.TODO(), ssr.etcdConnectionConfig.SnapshotTimeout.Duration)
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -344,8 +344,8 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
 		}
 	}
 	lastRevision := resp.Header.Revision
-	if isFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal{
-		ssr.logger.Infof("There are no updates since final last snapshot, skipping final full snapshot.")
+	if isFinal && ssr.PrevSnapshot.IsFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision {
+		ssr.logger.Infof("There are no new updates since previous final full snapshot, skipping new final full snapshot.")
 	} else {
 		// Note: As FullSnapshot size can be very large, so to avoid context timeout use "SnapshotTimeout" in context.WithTimeout()
 		ctx, cancel = context.WithTimeout(context.TODO(), ssr.etcdConnectionConfig.SnapshotTimeout.Duration)
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -344,9 +344,8 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
 		}
 	}
 	lastRevision := resp.Header.Revision
-
-	if ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal == isFinal {
-		ssr.logger.Infof("There are no updates since last snapshot, skipping full snapshot.")
+	if ssr.PrevSnapshot.IsFinal == true {
+		ssr.logger.Infof("Previous snapshot was set as final, skipping full snapshot.")
 	} else {
 		// Note: As FullSnapshot size can be very large, so to avoid context timeout use "SnapshotTimeout" in context.WithTimeout()
 		ctx, cancel = context.WithTimeout(context.TODO(), ssr.etcdConnectionConfig.SnapshotTimeout.Duration)


gardener-robot-ci-1:
Thank you @Shreyas-s14 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -344,9 +344,8 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
 		}
 	}
 	lastRevision := resp.Header.Revision
-
-	if ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal == isFinal {
-		ssr.logger.Infof("There are no updates since last snapshot, skipping full snapshot.")
+	if ssr.PrevSnapshot.IsFinal == true {
```
shreyas-s-rao, line 347:
```suggestion
	if ssr.PrevSnapshot.IsFinal == true && isFinal == true {
```

This checks whether the previous full snapshot is a final full snapshot, AND whether the final-ness of the previous full snapshot needs to be respected or ignored. In this context, I would also recommend that you rename the variable `isFinal` to a more meaningful name, like `checkPreviousFinalFullSnapshot`.

There are certain cases where we might not want to block a full snapshot because there's a final full snapshot already present in the bucket - this is because "final" full snapshot doesn't really strictly mean final, but rather a temporary final state. For instance, during Gardener CPM, a final full snapshot is taken on the old seed before stopping the etcd, and then the bucket is copied over to the new seed and the etcd is started again. At this point, even though there's a final full snapshot present in the bucket, the new etcdbr agent should continue taking new snapshots on top of the final full snapshot.

@ishan16696 please correct me if I'm wrong.

shreyas-s-rao (changes requested):
@Shreyas-s14 thanks for the quick fix! I have a suggestion. PTAL. @ishan16696 please weigh in here too.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: > This checks whether the previous full snapshot is a final full snapshot, AND whether the final-ness of the previous full snapshot needs to be respected or ignored. In this context, I would also recommend that you rename the variable isFinal to a more meaningful name, like checkPreviousFinalFullSnapshot

boolean `isFinal` is not here to indicate about previous final full snapshot. This check of previous full snapshot was final or not is indicated by `ssr.PrevSnapshot.IsFinal`, so this flag `isFinal` indicates wether this current request to trigger out-of-schedule full snapshot is for `final full snapshot` or not. 

@shreyas-s-rao your concern is valid, I have updated the condition:
```go
	if isFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal {
		ssr.logger.Infof("There are no updates since last snapshot, skipping full snapshot.")
	}
````

@shreyas-s-rao @Shreyas-s14 please take a look.

Comment (shreyas-s-rao) [reply pkg/snapshot/snapshotter/snapshotter.go:347]: @ishan16696 thanks. Why do we still need to check `ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision`? We no longer want to care about the revision from the previous full snapshot right?

Comment (ishan16696) [reply pkg/snapshot/snapshotter/snapshotter.go:347]:
```go
	if isFinal && ssr.PrevSnapshot.IsFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision  {
		ssr.logger.Infof("There are no updates since final last snapshot, skipping final full snapshot.")
	}
```

as discussed offline with @shreyas-s-rao @Shreyas-s14 this above condition will cover all of backup-restore requirements as well as for CPM.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -344,9 +344,8 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
 		}
 	}
 	lastRevision := resp.Header.Revision
-
-	if ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal == isFinal {
-		ssr.logger.Infof("There are no updates since last snapshot, skipping full snapshot.")
+	if isFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision && ssr.PrevSnapshot.IsFinal{
```
shreyas-s-rao, line 347:
```suggestion
	if isFinal && ssr.PrevSnapshot.IsFinal && ssr.PrevSnapshot.Kind == brtypes.SnapshotKindFull && ssr.PrevSnapshot.LastRevision == lastRevision {
```

Better placement of conditions ensures that we save CPU cycles by checking the highly necessary checks first (like isFinal and isPrevSnapshotFinal), and then the subsequent checks.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -344,9 +344,8 @@ func (ssr *Snapshotter) takeFullSnapshot(isFinal bool) (*brtypes.Snapshot, error
...
+		ssr.logger.Infof("There are no updates since final last snapshot, skipping final full snapshot.")
```
shreyas-s-rao, line 348:
```suggestion
		ssr.logger.Infof("There are no new updates since previous final full snapshot, skipping new final full snapshot.")
```

shreyas-s-rao (changes requested):
@Shreyas-s14 thanks making the changes. I have couple of nits. PTAL, thanks

ishan16696:
I think this PR should also contain release note, can you update with this release note:
```
Don't skip full snapshot, always trigger a full snapshot independent of new updates to etcd, preventing prometheus alerts of not taking a scheduled full snapshot.
```

shreyas-s-rao (approved):
@Shreyas-s14 thanks for making the requested changes.
/lgtm

ishan16696 (approved):
LGTM!!

PR #815, opened 2024-12-19
Introduce `gosec` for Static Application Security Testing (SAST)

### LINKED ISSUE #812: Introduce gosec for Static Application Security Testing (SAST) for etcd-backup-restore

/area security
/kind enhancement

Introduce gosec for Static Application Security Testing (SAST) for etcd-backup-restore.

**Why is this needed**:
To make the code more inline with secure development guidelines.

shreyas-s-rao:
/area security
/kind enhancement

Introduce `gosec` for Static Application Security Testing (SAST) for etcd-backup-restore.

Fixes #812 

/invite @ashwani2k @ishan16696 

**Release note**:

```improvement developer
Introduce `gosec` for Static Application Security Testing (SAST) via `make sast` and `make sast-report`.
```

### DIFF
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -401,7 +401,9 @@ func AddLearnerWithRetry(ctx context.Context, m Control, retrySteps int, dataDir
 		}
 		if err := m.AddMemberAsLearner(ctx); err != nil {
 			if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCTooManyLearners)) {
-				miscellaneous.SleepWithContext(ctx, EtcdTimeout)
+				if err1 := miscellaneous.SleepWithContext(ctx, EtcdTimeout); err != nil {
+					return errors.Join(err, err1)
+				}
 			}
 			return err
 		}
--- a/pkg/member/member_control.go
+++ b/pkg/member/member_control.go
@@ -401,7 +401,7 @@ func AddLearnerWithRetry(ctx context.Context, m Control, retrySteps int, dataDir
 		}
 		if err := m.AddMemberAsLearner(ctx); err != nil {
 			if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCTooManyLearners)) {
-				if err1 := miscellaneous.SleepWithContext(ctx, EtcdTimeout); err != nil {
+				if err1 := miscellaneous.SleepWithContext(ctx, EtcdTimeout); err1 != nil {
 					return errors.Join(err, err1)
 				}
 			}
--- a/pkg/server/httpAPI.go
+++ b/pkg/server/httpAPI.go
@@ -45,6 +45,7 @@ const (
 	initializationStatusSuccessful = "Successful"
 	initializationStatusFailed     = "Failed"
 
+	// serverReadHeaderTimeout is the timeout for reading the request headers, to avoid slowloris attacks.
 	serverReadHeaderTimeout = 5 * time.Second
 )
 
@@ -211,8 +212,7 @@ func (h *HTTPHandler) serveHealthz(rw http.ResponseWriter, req *http.Request) {
 		h.Logger.Errorf("Unable to marshal health status to json: %v", err)
 		return
 	}
-	_, err = rw.Write(out)
-	if err != nil {
+	if _, err = rw.Write(out); err != nil {
 		h.Logger.Errorf("Unable to write health status response: %v", err)
 	}
 }
@@ -278,8 +278,7 @@ func (h *HTTPHandler) serveInitializationStatus(rw http.ResponseWriter, req *htt
 
 	rw.WriteHeader(http.StatusOK)
 
-	_, err := rw.Write([]byte(h.initializationStatus))
-	if err != nil {
+	if _, err := rw.Write([]byte(h.initializationStatus)); err != nil {
 		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
 		return
 	}
@@ -327,8 +326,7 @@ func (h *HTTPHandler) serveFullSnapshotTrigger(rw http.ResponseWriter, req *http
 		return
 	}
 	rw.WriteHeader(http.StatusOK)
-	_, err = rw.Write(out)
-	if err != nil {
+	if _, err = rw.Write(out); err != nil {
 		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
 		return
 	}
@@ -362,8 +360,7 @@ func (h *HTTPHandler) serveDeltaSnapshotTrigger(rw http.ResponseWriter, req *htt
 		return
 	}
 	rw.WriteHeader(http.StatusOK)
-	_, err = rw.Write(out)
-	if err != nil {
+	if _, err = rw.Write(out); err != nil {
 		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
 		return
 	}
@@ -408,8 +405,7 @@ func (h *HTTPHandler) serveLatestSnapshotMetadata(rw http.ResponseWriter, req *h
 		return
 	}
 	rw.WriteHeader(http.StatusOK)
-	_, err = rw.Write(out)
-	if err != nil {
+	if _, err = rw.Write(out); err != nil {
 		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
 		return
 	}
@@ -612,14 +608,14 @@ func (h *HTTPHandler) delegateReqToLeader(rw http.ResponseWriter, req *http.Requ
 		}
 	}()
 
-	client, err := factory.NewCluster()
+	cl, err := factory.NewCluster()
 	if err != nil {
 		h.Logger.Warnf("failed to create etcd cluster client:  %v", err)
 		rw.WriteHeader(http.StatusMethodNotAllowed)
 		return
 	}
 	defer func() {
-		if err1 := client.Close(); err1 != nil {
+		if err1 := cl.Close(); err1 != nil {
 			h.Logger.Errorf("Error closing etcd cluster client: %v", err1)
 		}
 	}()
@@ -633,7 +629,7 @@ func (h *HTTPHandler) delegateReqToLeader(rw http.ResponseWriter, req *http.Requ
 		return
 	}
 
-	_, etcdLeaderEndPoint, err := miscellaneous.GetLeader(ctx, clientMaintenance, client, h.EtcdConnectionConfig.Endpoints[0])
+	_, etcdLeaderEndPoint, err := miscellaneous.GetLeader(ctx, clientMaintenance, cl, h.EtcdConnectionConfig.Endpoints[0])
 	if err != nil {
 		h.Logger.Warnf("Unable to get the etcd leader endpoint: %v", err)
 		rw.WriteHeader(http.StatusMethodNotAllowed)
--- a/pkg/server/httpAPI.go
+++ b/pkg/server/httpAPI.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/gardener/etcd-backup-restore/pkg/etcdutil"
 	etcdclient "github.com/gardener/etcd-backup-restore/pkg/etcdutil/client"
@@ -43,6 +44,8 @@ const (
 	initializationStatusProgress   = "Progress"
 	initializationStatusSuccessful = "Successful"
 	initializationStatusFailed     = "Failed"
+
+	serverReadHeaderTimeout = 5 * time.Second
 )
 
 var emptyStruct struct{}
@@ -139,8 +142,9 @@ func (h *HTTPHandler) RegisterHandler() {
 	mux.Handle("/metrics", promhttp.Handler())
 
 	h.server = &http.Server{
-		Addr:    fmt.Sprintf(":%d", h.Port),
-		Handler: mux,
+		Addr:              fmt.Sprintf(":%d", h.Port),
+		Handler:           mux,
+		ReadHeaderTimeout: serverReadHeaderTimeout,
 	}
 }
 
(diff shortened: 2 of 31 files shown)


pkg/member/member_control.go
```diff
@@ -401,7 +401,9 @@ func AddLearnerWithRetry(ctx context.Context, m Control, retrySteps int, dataDir
 		}
 		if err := m.AddMemberAsLearner(ctx); err != nil {
 			if errors.Is(err, rpctypes.Error(rpctypes.ErrGRPCTooManyLearners)) {
-				miscellaneous.SleepWithContext(ctx, EtcdTimeout)
+				if err1 := miscellaneous.SleepWithContext(ctx, EtcdTimeout); err != nil {
+					return errors.Join(err, err1)
+				}
```
ishan16696, line 406:
```suggestion
				if err1 := miscellaneous.SleepWithContext(ctx, EtcdTimeout); err1 != nil {
					return errors.Join(err, err1)
				}
```

pkg/server/httpAPI.go
```diff
@@ -43,6 +44,8 @@ const (
 	initializationStatusProgress   = "Progress"
 	initializationStatusSuccessful = "Successful"
 	initializationStatusFailed     = "Failed"
+
+	serverReadHeaderTimeout = 5 * time.Second
```
ishan16696, line 48: can you also add 1 line of comment why this ReadHeaderTimeout is required ? 
It will be helpful for future reference.

pkg/server/httpAPI.go
```diff
@@ -202,12 +206,15 @@ func (h *HTTPHandler) serveHealthz(rw http.ResponseWriter, req *http.Request) {
...
+	out, err := json.Marshal(healthCheck)
 	if err != nil {
 		h.Logger.Errorf("Unable to marshal health status to json: %v", err)
 		return
 	}
-	rw.Write([]byte(json))
+	_, err = rw.Write(out)
+	if err != nil {
+		h.Logger.Errorf("Unable to write health status response: %v", err)
+	}
```
ishan16696, line 217:
```suggestion
	if _, err = rw.Write(out); err != nil {
		h.Logger.Errorf("Unable to write health status response: %v", err)
	}
```

pkg/server/httpAPI.go
```diff
@@ -271,7 +278,11 @@ func (h *HTTPHandler) serveInitializationStatus(rw http.ResponseWriter, req *htt
 
 	rw.WriteHeader(http.StatusOK)
 
-	rw.Write([]byte(h.initializationStatus))
+	_, err := rw.Write([]byte(h.initializationStatus))
+	if err != nil {
+		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
+		return
+	}
```
ishan16696, line 285:
```suggestion
	if _, err := rw.Write([]byte(h.initializationStatus)); err != nil {
		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
		return
	}
```

pkg/server/httpAPI.go
```diff
@@ -309,14 +320,18 @@ func (h *HTTPHandler) serveFullSnapshotTrigger(rw http.ResponseWriter, req *http
...
 		rw.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 	rw.WriteHeader(http.StatusOK)
-	rw.Write(json)
+	_, err = rw.Write(out)
+	if err != nil {
+		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
+		return
+	}
```
ishan16696, line 334:
```suggestion
	if _, err = rw.Write(out); err != nil {
		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
	}
```

pkg/server/httpAPI.go
```diff
@@ -340,14 +355,18 @@ func (h *HTTPHandler) serveDeltaSnapshotTrigger(rw http.ResponseWriter, req *htt
...
```
ishan16696, line 369:
```suggestion
	if _, err = rw.Write(out); err != nil {
		h.Logger.Errorf("Unable to write latest snapshot metadata response: %v", err)
	}
```

ishan16696 (changes requested):

pkg/server/httpAPI.go
```diff
@@ -521,7 +544,11 @@ func getInitialCluster(ctx context.Context, initialCluster string, etcdConn brty
 		logger.Warnf("Error with NewCluster() : %v", err)
 		return initialCluster
 	}
-	defer cli.Close()
+	defer func() {
+		if err1 := cli.Close(); err1 != nil {
```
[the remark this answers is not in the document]
shreyas-s-rao: I used different error name to avoid clashes during function exit when defer is called, while an `err` is already defined and returned. The scope of the main function is used for the defer func as well, hence the possibility of a clash.

pkg/server/httpAPI.go
```diff
@@ -579,15 +606,23 @@ func (h *HTTPHandler) delegateReqToLeader(rw http.ResponseWriter, req *http.Requ
 		rw.WriteHeader(http.StatusMethodNotAllowed)
 		return
 	}
-	defer clientMaintenance.Close()
+	defer func() {
+		if err1 := clientMaintenance.Close(); err != nil {
```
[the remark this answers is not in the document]
shreyas-s-rao: Same as above

pkg/server/httpAPI.go
```diff
@@ -579,15 +606,23 @@ func (h *HTTPHandler) delegateReqToLeader(rw http.ResponseWriter, req *http.Requ
...
 
 	client, err := factory.NewCluster()
 	if err != nil {
 		h.Logger.Warnf("failed to create etcd cluster client:  %v", err)
 		rw.WriteHeader(http.StatusMethodNotAllowed)
 		return
 	}
-	defer client.Close()
+	defer func() {
+		if err1 := client.Close(); err1 != nil {
```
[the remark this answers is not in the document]
shreyas-s-rao: Same as above

pkg/server/httpAPI.go
shreyas-s-rao: Done

ishan16696 (changes requested):

pkg/server/httpAPI.go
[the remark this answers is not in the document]
shreyas-s-rao: Sure, makes sense. I'll make the change.

pkg/server/httpAPI.go
```diff
@@ -656,32 +687,36 @@ func (h *HTTPHandler) delegateReqToLeader(rw http.ResponseWriter, req *http.Requ
...
 
-	response, err := client.Get(backupRestoreURL)
+	response, err := cl.Get(backupRestoreURL)
 	if err != nil {
 		return false, err
 	}
-	defer response.Body.Close()
+	defer func() {
+		if err1 := response.Body.Close(); err1 != nil {
+			err = fmt.Errorf("failed to close response body: %v", err1)
```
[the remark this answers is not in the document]
shreyas-s-rao: The `err` is returned, and can be modified by the defer func. But I realised just now that I should have joined the existing `err` from main function with the error from the defer func. I'll make that change.

shreyas-s-rao:
@ishan16696 thanks for your review. I have addressed them. PTAL

pkg/snapshot/restorer/restorer.go
```diff
@@ -126,7 +126,10 @@ func (r *Restorer) Restore(ro brtypes.RestoreOptions, m member.Control) (*embed.
 				r.logger.Errorf("failed to close etcd cluster client: %v", err)
 			}
 		}()
-		m.UpdateMemberPeerURL(context.TODO(), clientCluster)
+
+		if err := m.UpdateMemberPeerURL(context.TODO(), clientCluster); err != nil {
+			return e, err
+		}
```
ishan16696, line 132: Is it make sense to fail the restoration if backup-restore fail to update the `PeerURL` after restoration ? what if we just log the warning ?

ishan16696 (commented):
overall looks good. only 1 nit,

test/e2e/integration/utils.go
```diff
@@ -62,8 +69,12 @@ func (cmd *Cmd) RunCmdWithFlags() error {
...
 			}
 			if len(line) > 0 {
-				logWriter.WriteString(fmt.Sprintf("%s\n", line))
-				logWriter.Flush()
+				if _, err1 := logWriter.WriteString(fmt.Sprintf("%s\n", line)); err1 != nil {
+					cmd.Logger.Errorf("failed to write to log file for command %s: %v", cmd.Task, err1)
+				}
+				if err1 := logWriter.Flush(); err1 != nil {
+					cmd.Logger.Errorf("failed to flush log file for command %s: %v", cmd.Task, err1)
+				}
```
ishan16696, line 77: here as well, you can use `err` variable name as it's inside the if block, it won't cause an issue.

test/e2e/integration/utils.go
```diff
@@ -76,18 +87,20 @@ func (cmd *Cmd) RunCmdWithFlags() error {
...
```
ishan16696, line 95: same as above

test/e2e/integration/utils.go
```diff
@@ -25,18 +26,24 @@ type Cmd struct {
...
 	if err != nil {
 		cmd.Logger.Errorf("failed to create log file for command %s: %v", cmd.Task, err)
 		return err
 	}
-	defer file.Close()
+	defer func() {
+		if err1 := file.Close(); err1 != nil {
+			cmd.Logger.Errorf("failed to close log file for command %s: %v", cmd.Task, err1)
+		}
+	}()
```
ishan16696, line 46: can you also change this `err1` to `err` as discuss in above comments

ishan16696 (changes requested):

shreyas-s-rao:
@ishan16696 thanks for your review. I've addressed your comments. PTAL.

test/e2e/integration/utils.go
```diff
@@ -76,18 +88,20 @@ func (cmd *Cmd) RunCmdWithFlags() error {
...
-	if err := cmd.process.Wait(); err != nil {
-		if exiterr, ok := err.(*exec.ExitError); ok {
-			cmd.Logger.Infof("command execution stopped (signal: %s)", exiterr.Error())
-		} else if err != nil {
-			cmd.Logger.Errorf("command %s failed with error: %v", cmd.Task, err)
-			return err
+	if err = cmd.process.Wait(); err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			cmd.Logger.Infof("command execution stopped (signal: %s)", exitErr.Error())
```
ishan16696, line 104: do you want to print this log as info ?

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Failing to update peer URLs is definitely an error, rather than a warning. If clustering fails, there's no use starting the etcd member in the first place. And it would be difficult to debug such an error, because the etcd config file will look fine but runtime config wasn't updated, and going through warning logs isn't ideal. I would recommend throwing an error and failing early in such cases.

test/e2e/integration/utils.go
```diff
@@ -25,18 +26,25 @@ type Cmd struct {
...
-	cmd.process.Process.Signal(os.Interrupt)
+	if err := cmd.process.Process.Signal(os.Interrupt); err != nil {
+		cmd.Logger.Errorf("failed to stop process %s: %v", cmd.Task, err)
+	}
 }
 
 // RunCmdWithFlags creates a process out of the  Cmd object.
 func (cmd *Cmd) RunCmdWithFlags() error {
-	cmd.process = exec.Command(cmd.Task, cmd.Flags...)
+	var err error
```
ishan16696, line 36: this var declaration is not required if we create local variables with their scope belongs to their block.

test/e2e/integration/utils.go
shreyas-s-rao: I didn't change anything here, just changed the error parsing on line 103. I have no strong opinion on how we print the log line here, since it's just a test and will be printed anyway

test/e2e/integration/utils.go
shreyas-s-rao: I just created it to avoid redefining it later if we remove/change the function call that first creates the `err` variable.

pkg/snapshot/restorer/restorer.go
ishan16696: > Failing to update peer URLs is definitely an error, rather than a warning

yeah, correct then we can just log an error rather than a warning.

> If clustering fails, there's no use starting the etcd member in the first place. And it would be difficult to debug such an error, because the etcd config file will look fine but runtime config wasn't updated

this update to member Peer URL is only introduce for multi-node (recovering from permanent quorum loss),
so for `cluster size =1`, it won't cause an issue even if this call get failed.
for `HA cluster`, it can cause an issue while doing recovering from permanent quorum loss, so I don't have any hard opinion on this.  If you want to return the `err` then it's fine to me, just test the scenario for restoration for `cluster size =1`. Hope that's fine.

test/e2e/integration/utils.go
ishan16696: I think error log would make more sense while debugging in case of e2e tests failure.

test/e2e/integration/utils.go
ishan16696: yeah, but if look at the code, some part of code is re-declaring it, where you have made changes, you are not redeclaring it, can we have uniformity across this func ?

pkg/snapshot/restorer/restorer.go
shreyas-s-rao: Sure. I would like to retain the returning of the error.

test/e2e/integration/utils.go
shreyas-s-rao: Sure, I'll remove this and make it uniform then

ishan16696 (approved):
thanks for making for changes.
LGTM!!

PR #818, opened 2024-12-24
Add `Bucket Lock` immutable objects support for AWS S3

### LINKED ISSUE #777: [Feature] ☂️ Support `Bucket Lock` mechanism for backups taken by backup-restore for various storage provider

Users might wish to enable the bucket lock mechanism for their objects stored in buckets to use write-once-read-many(WORM) model. As the storage providers exhibit varying functionality for bucket lock, we need to adapt our approach based on each provider's specific bucket lock implementation.

To support the bucket lock mechanism, the following features of the backup-restore process need to be enhanced:
1. Garbage collection of backups.
2. The ability to skip certain snapshots during restoration, if required.

- [x] AWS's S3
     - https://github.com/gardener/etcd-backup-restore/pull/818
- [x] GCP's GCS
   - https://github.com/gardener/etcd-backup-restore/pull/776
- [x] Azure's ABS 
    - https://github.com/gardener/etcd-backup-restore/pull/781

ishan16696:
This PR enables backup-restore to handle AWS S3's object lock mechanism which will make snapshots taken by backup-restore immutable. It also adjusted the Restoration and garbage collector functionality to handle immutable snapshots for S3 object store.

Fixes #777 

These case I have tested:

For Restoration functionality
  - [x] Restoration functionality of backup-restore should work with buckets which don't have bucket/object lock enabled (Backward compatibility).
  - [x] Restoration functionality of backup-restore should work with buckets which have both versioned immutable as well as non-immutable, non-versioned snapshots present.
  - [x] Restoration functionality of backup-restore should work with buckets which have only immutable versioned snapshots(full as well as delta) present.
  - [x] Restoration functionality of backup-restore should work when attacker corrupted/deleted immutable versioned snapshots present in bucket.

For Garbage collection of snapshots:
  - [x] Backward compatibility with buckets which don't have immutability enabled.
  - [x] GC should work with buckets which have both immutable as well as non immutable objects.
  - [x] GC should work with buckets which have only immutable objects.

**Release note**:

```noteworthy operator
etcd-backup-restore now supports immutable objects for storage provider: AWS S3, provided by the [Bucket Lock](https://aws.amazon.com/s3/features/object-lock/) feature.
```

```noteworthy operator
Snapshots garbage collection performed by etcd-backup-restore (if enabled) for S3 is performed only when the snapshots's retention period get expires.
```

```noteworthy operator
Support for Bucket lock (S3) in etcd-backup-restore is backward compatible. For more info please refer to this doc: https://github.com/gardener/etcd-backup-restore/blob/master/docs/usage/enabling_immutable_snapshots.md
```


pkg/snapstore/s3_snapstore.go
```diff
@@ -510,35 +518,108 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+						snap, err := ParseSnapshot(path.Join(prefix, k))
+						if err != nil {
+							// Warning
+							logrus.Warnf("Invalid snapshot found. Ignoring it: %s", k)
+						} else {
+							// capture the versionID of snapshot and immutability expiry time of snapshot.
+							snap.VersionID = *version.VersionId
+							if bucketImmutableExpiryTimeInDays != nil {
+								// To get S3 object's "RetainUntilDate" or "ImmutabilityExpiryTime", backup-restore need to make an API call for each snapshots.
+								// To avoid API calls for each snapshots, backup-restore is calculating the "ImmutabilityExpiryTime" using bucket retention period.
```
shreyas-s-rao, line 576:
```suggestion
								// To get S3 object's "RetainUntilDate" or "ImmutabilityExpiryTime", backup-restore need to make an API call for each snapshot.
								// To avoid API calls for each snapshot, backup-restore is calculating the "ImmutabilityExpiryTime" using bucket retention period.
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -510,35 +518,108 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+								// ImmutabilityExpiryTime = SnapshotCreationTime + ObjectRetentionTimeInDays
+								snap.ImmutabilityExpiryTime = snap.CreatedOn.Add(time.Duration(*bucketImmutableExpiryTimeInDays) * 24 * time.Hour)
+							} else {
+								// retry to get bucketImmutableExpiryTimeInDays
+								_, bucketImmutableExpiryTimeInDays, err = GetBucketImmutabilityTime(s)
```
shreyas-s-rao, line 581: How is `bucketImmutableExpiryTimeInDays` being used here? The snapshot gets added to the `snapList` without setting `snap.ImmutabilityExpiryTime`, which doesn't seem right. Can you please confirm?

pkg/snapstore/s3_snapstore.go
```diff
@@ -510,35 +518,108 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+									logrus.Warnf("unable to get bucket immutability expiry time: %v", err)
+								}
+							}
+							snapList = append(snapList, snap)
+						}
+					}
 				} else {
-					snapList = append(snapList, snap)
+					// Warning
+					logrus.Warnf("Snapshot: %s with versionID: %s found to be not latest, it was last modified: %s. Ignoring it.", *version.Key, *version.VersionId, version.LastModified)
```
shreyas-s-rao, line 591:
```suggestion
					logrus.Warnf("Snapshot %s with versionID %s found not to be the latest version, it was last modified at %s. Ignoring it.", *version.Key, *version.VersionId, version.LastModified)
```

pkg/snapstore/s3_snapstore_test.go
shreyas-s-rao, line 1: Can you please add some tests specific to handling immutable backups for S3 buckets? Like garbage collection for instance.

shreyas-s-rao (changes requested):
@ishan16696 thanks for the well-written PR and the docs. I have a few nits. PTAL, thanks.

pkg/snapstore/s3_snapstore.go
```diff
@@ -622,3 +717,20 @@ func getSSECreds(sseCustomerKey, sseCustomerAlgorithm *string) (SSECredentials,
...
+	objectConfig, err := s.client.GetObjectLockConfiguration(&s3.GetObjectLockConfigurationInput{
+		Bucket: aws.String(s.bucket),
+	})
+	if err != nil {
+		return false, nil, err
+	}
+
+	if *objectConfig.ObjectLockConfiguration.ObjectLockEnabled == "Enabled" {
+		// assumption: retention period of bucket will always be in days, not years.
+		return true, objectConfig.ObjectLockConfiguration.Rule.DefaultRetention.Days, nil
```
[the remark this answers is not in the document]
ishan16696: here, it won't be an issue as caller for this func is checking the nil pointer reference.

```go
if bucketImmutableExpiryTimeInDays != nil {
....
}
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -320,9 +319,18 @@ func NewS3FromClient(bucket, prefix, tempDir string, maxParallelChunkUploads uin
...
+			Bucket:    aws.String(s.bucket),
+			Key:       aws.String(path.Join(snap.Prefix, snap.SnapDir, snap.SnapName)),
+			VersionId: aws.String(snap.VersionID),
+		}
+	} else {
+		getObjectInput = &s3.GetObjectInput{
+			Bucket: aws.String(s.bucket),
+			Key:    aws.String(path.Join(snap.Prefix, snap.SnapDir, snap.SnapName)),
+		}
 	}
```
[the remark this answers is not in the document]
ishan16696: done, PTAL

pkg/snapstore/s3_snapstore.go
```diff
@@ -547,11 +628,25 @@ func (s *S3SnapStore) List(_ bool) (brtypes.SnapList, error) {
...
+		// to delete non-versioned snapshots present in bucket.
+		if _, err := s.client.DeleteObject(&s3.DeleteObjectInput{
+			Bucket: aws.String(s.bucket),
+			Key:    aws.String(path.Join(snap.Prefix, snap.SnapDir, snap.SnapName)),
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
 }
```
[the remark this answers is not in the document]
ishan16696: done, PTAL

pkg/snapstore/s3_snapstore.go
```diff
@@ -622,3 +717,20 @@ func getSSECreds(sseCustomerKey, sseCustomerAlgorithm *string) (SSECredentials,
...
+// GetBucketImmutabilityTime check objectlock is enabled for given bucket, if it does returns the retention period.
+func GetBucketImmutabilityTime(s *S3SnapStore) (bool, *int64, error) {
+	objectConfig, err := s.client.GetObjectLockConfiguration(&s3.GetObjectLockConfigurationInput{
+		Bucket: aws.String(s.bucket),
+	})
+	if err != nil {
+		return false, nil, err
+	}
+
+	if *objectConfig.ObjectLockConfiguration.ObjectLockEnabled == "Enabled" {
```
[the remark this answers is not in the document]
ishan16696: thanks for pointing out. made the changes.

pkg/snapstore/s3_snapstore.go
ishan16696: yes, you are correct. Actually I added this in case of previous api call failed.
Another way to handle this is to add a retry mechanism in first api call, and remove this else handling part , wdyt ?

ishan16696:
I have rebased the PR on master.

pkg/snapstore/s3_snapstore.go
shreyas-s-rao: Yes, retry mechanism seems more intuitive here.

pkg/snapstore/s3_snapstore.go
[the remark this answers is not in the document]
shreyas-s-rao: `bucketImmutableExpiryTimeInDays` is deduced from `GetBucketImmutabilityTime()`, which in turn returns `objectConfig.ObjectLockConfiguration.Rule.DefaultRetention.Days`. Simply checking `bucketImmutableExpiryTimeInDays` will not help, since you could have `Rule` an non-nil, but `DefaultRetention` could be nil, and throw an exception. That's the reason the entire path should be checked here.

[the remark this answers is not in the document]
ishan16696: ok, got your point. I will make the change.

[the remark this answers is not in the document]
ishan16696: made the changes. Also added a some unit tests for the same. PTAL

pkg/snapstore/s3_snapstore.go
ishan16696: made the changes, PTAL

pkg/snapstore/snapstore_test.go
```diff
@@ -649,13 +650,41 @@ var _ = Describe("Server Side Encryption Customer Managed Key for S3", func() {
...
+	})
+})
+
+var _ = Describe("Get Immutability time for S3 bucket", func() {
+	awsS3Client := &mockS3Client{
+		objects:          objectMap,
+		prefix:           prefixV2,
+		multiPartUploads: map[string]*[][]byte{},
+	}
+	var s3ObjectLockBucketBUTRulesNotDefined = "mock-s3ObjectLockBucketBUTRulesNotDefined"
```
shreyas-s-rao, line 685: Can you please add a comment about what `s3ObjectLockBucketBUTRulesNotDefined` exactly means here?

shreyas-s-rao (changes requested):
@ishan16696 thanks for addressing my review comments! Just a small nit, but overall LGTM.

pkg/snapstore/snapstore_test.go
shreyas-s-rao: Ah then can you please rename to `s3ObjectLockBucketButRulesNotDefined`. The `BUT` was capitalized, so that threw me off, since I thought it was an abbreviation for something 😅

ishan16696: ok, no worries, I will rename it.

shreyas-s-rao (approved):
/lgtm

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,35 +517,115 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+		}
+
+		in := &s3.ListObjectVersionsInput{
+			Bucket: aws.String(s.bucket),
+			Prefix: aws.String(prefix),
+		}
+
+		if err := s.client.ListObjectVersionsPages(in, func(page *s3.ListObjectVersionsOutput, lastPage bool) bool {
+			for _, version := range page.Versions {
+				if *version.IsLatest {
```
renormalize, line 576: I do not understand why the latest version is being checked for here? Shouldn't we be checking for the **oldest** version?

Checking the [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html):

> Placing a retention period or a legal hold on an object protects only the version that's specified in the request. Retention periods and legal holds don't prevent new versions of the object from being created, or delete markers to be added on top of the object.

If an object is uploaded, and for some reason a newer version is created, the *original* object is no longer `IsLatest == true`. The unwanted version will be `IsLatest == true`.

Can you please help me understand?

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,35 +517,115 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+								// ImmutabilityExpiryTime = SnapshotCreationTime + ObjectRetentionTimeInDays
+								snap.ImmutabilityExpiryTime = snap.CreatedOn.Add(time.Duration(*bucketImmutableExpiryTimeInDays) * 24 * time.Hour)
+							}
+							snapList = append(snapList, snap)
+						}
+					}
 				} else {
-					snapList = append(snapList, snap)
+					// Warning
+					logrus.Warnf("Snapshot %s with versionID [%s] found not to be latest, it was last modified at %s. Ignoring it.", *version.Key, *version.VersionId, *version.LastModified)
```
renormalize, line 597: If the versionID corresponds to *not* being the latest, wouldn't that mean the object was modified, and thus we need to find the oldest version?

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,35 +517,115 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+	}
+
+	if versioningStatus.Status != nil && *versioningStatus.Status == s3.BucketVersioningStatusEnabled {
+		// versioning is found to be enabled on given bucket.
+		logrus.Info("Bucket versioning is found to be enabled.")
+
+		var bucketImmutableExpiryTimeInDays *int64
+		var isObjectLockEnabled bool
+
+		if err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
```
renormalize, line 550: What's the reason behind implementing a retry mechanism for a standard S3 API call? Did you observe the request not responding at all?
If the response time is too long, perhaps we can make use of `GetObjectLockConfigurationWithContext`?

pkg/types/snapstore.go
```diff
@@ -76,6 +77,7 @@ type SnapStore interface {
 	// List returns a sorted list (based on the last revision, ascending) of all snapshots in the store.
 	// includeAll specifies whether to include all snapshots while listing, including those with exclude tags.
 	// Snapshots with exclude tags are not listed unless includeAll is set to true.
+	// Note: "includeAll" boolean is only applicable for storage provider: ABS and GCS.
```
renormalize, line 80: What's the reason behind not supporting it for S3?
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html

renormalize (changes requested):
Thanks for the PR @ishan16696!

There's a few places where I'd like to understand the code better, so I've left my review comments there.

pkg/types/snapstore.go
```diff
@@ -96,6 +98,8 @@ type Snapshot struct {
 	CompressionSuffix      string    `json:"compressionSuffix"` // CompressionSuffix depends on compression policy
 	IsFinal                bool      `json:"isFinal"`
 	ImmutabilityExpiryTime time.Time `json:"immutabilityExpriyTime"`
+	// It is used only for AWS S3 object lock immutability.
```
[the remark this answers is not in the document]
shreyas-s-rao: 100% agree with you @renormalize . I also thought about this, but didn't want to bring it into the scope of this PR. Can you please raise an issue regarding this? If not now, we should definitely address this when refactoring the code.

pkg/types/snapstore.go
ishan16696: > What's the reason behind not supporting it for S3?

the reason is to get tag of any object, s3 client have to get all object first then make API call for each object, hence to avoid `n` api call for `n` object, I choose to go with different method.
I have mention this here in doc but I have only mention that this can be achieved with different method (soft delete in this case)
https://github.com/gardener/etcd-backup-restore/blob/295d4103ffd3366c514ee72936a9ba8899cd24ad/docs/usage/enabling_immutable_snapshots.md#aws-s3-1

pkg/snapstore/s3_snapstore.go
```diff
@@ -549,6 +634,17 @@ func (s *S3SnapStore) List(_ bool) (brtypes.SnapList, error) {
...
+		// to delete versioned snapshots present in bucket.
+		_, err := s.client.DeleteObject(&s3.DeleteObjectInput{
+			Bucket:    aws.String(s.bucket),
+			Key:       aws.String(path.Join(snap.Prefix, snap.SnapDir, snap.SnapName)),
+			VersionId: &snap.VersionID,
+		})
+		return err
+	}
+
+	// to delete non-versioned snapshots present in bucket.
```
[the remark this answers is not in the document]
ishan16696: made the changes, PTAL

pkg/snapstore/s3_snapstore_test.go
```diff
@@ -207,6 +205,95 @@ func (m *mockS3Client) ListObjectsPages(in *s3.ListObjectsInput, callback func(*
...
+	var (
+		count    int64 = 0
+		limit    int64 = 1 // aws default is 1000.
+		lastPage bool  = false
+		keys     []string
+		out      = &s3.ListObjectVersionsOutput{
+			Prefix:   in.Prefix,
+			Versions: make([]*s3.ObjectVersion, 0),
+		}
+	)
```
anveshreddy18, line 219:
```suggestion
	var (
		count    int64
		limit    int64 = 1 // aws default is 1000.
		lastPage bool
		keys     []string
		out      = &s3.ListObjectVersionsOutput{
			Prefix:   in.Prefix,
			Versions: make([]*s3.ObjectVersion, 0),
		}
	)
```
Just a nit as the zero values are what you wanted only.

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,35 +517,115 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+				return err
+			}
+			if !isObjectLockEnabled {
+				logrus.Warnf("Bucket versioning is found to be enabled but object lock is not found to be enabled.")
+				logrus.Warnf("Please enable the object lock for the given bucket to to have immutable snapshots.")
+			}
+			if isObjectLockEnabled && bucketImmutableExpiryTimeInDays == nil {
+				logrus.Warnf("Object lock is found to be enabled but object lock rules or default retention period are not found to be set.")
+			}
+			return nil
```
anveshreddy18, line 564:
```suggestion
			if isObjectLockEnabled, bucketImmutableExpiryTimeInDays, err = GetBucketImmutabilityTime(s); err != nil {
				logrus.Errorf("unable to check bucket immutability retention period: %v", err)
				return err
			} else if !isObjectLockEnabled {
				logrus.Warnf("Bucket versioning is found to be enabled but object lock is not found to be enabled.")
				logrus.Warnf("Please enable the object lock for the given bucket to to have immutable snapshots.")
			} else if bucketImmutableExpiryTimeInDays == nil {
				logrus.Warnf("Object lock is found to be enabled but object lock rules or default retention period are not found to be set.")
			}
			return nil
```

Just to make sure the condition `isObjectLockEnabled` is checked only once. Feel free to ignore if you feel it doesn't matter.

anveshreddy18 (changes requested):
Thanks for the PR @ishan166696. I have few basic nits PTAL, otherwise it looks good.

pkg/snapstore/s3_snapstore.go
ishan16696: > for some reason a newer version is created

backup-restore don't perform any operation on snapshot once it is created, apart from GC which will be done after snapshot is no longer protected by object lock.

Right now, I'm only considering the latest snapshots while `listing` the snapshots from bucket, If snapshot found to be no longer latest then backup-restore will just print the warning. 
- It is done to avoid `tagging method to ignore the snapshots` as explain in this comment https://github.com/gardener/etcd-backup-restore/pull/818#discussion_r1906517909, tagging method is not feasible for S3.
- If you want your snapshot back, then simply remove the deletionMarker present for that particular snapshot.


> Shouldn't we be checking for the oldest version?

I got your point of checking the oldest snapshot, but then backup-restore won't be able to support feature of `ignoring any snapshots`. Moreover, this checking of oldest version is not full proof, take an illustration:

pkg/snapstore/s3_snapstore.go
ishan16696: I have made the changes, PTAL

pkg/snapstore/s3_snapstore_test.go
ishan16696: pushed the changes.

pkg/snapstore/s3_snapstore.go
```diff
@@ -325,6 +325,11 @@ func (s *S3SnapStore) Fetch(snap brtypes.Snapshot) (io.ReadCloser, error) {
 		Bucket: aws.String(s.bucket),
 		Key:    aws.String(path.Join(snap.Prefix, snap.SnapDir, snap.SnapName)),
 	}
+
+	if len(snap.VersionID) > 0 {
+		// To fetch the versioned snapshots incase is object enabled for bucket.
```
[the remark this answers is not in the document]
ishan16696: done.

anveshreddy18 (approved):
lgtm!

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,35 +517,113 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+						} else {
+							// capture the versionID of snapshot and immutability expiry time of snapshot.
+							snap.VersionID = version.VersionId
+							if bucketImmutableExpiryTimeInDays != nil {
+								// To get S3 object's "RetainUntilDate" or "ImmutabilityExpiryTime", backup-restore need to make an API call for each snapshot.
+								// To avoid API calls for each snapshot, backup-restore is calculating the "ImmutabilityExpiryTime" using bucket retention period.
+								// ImmutabilityExpiryTime = SnapshotCreationTime + ObjectRetentionTimeInDays
+								snap.ImmutabilityExpiryTime = snap.CreatedOn.Add(time.Duration(*bucketImmutableExpiryTimeInDays) * 24 * time.Hour)
+							}
+							snapList = append(snapList, snap)
```
renormalize, line 590: The `List` call only appends only one particular version of the snapshot (in the current implementation, the latest) into the `snapList`. 
However, this would be not be sufficient for garbage collection since all versions should be returned by `List`, to ensure all versions are cleaned up.

renormalize (changes requested):
Capturing @shreyas-s-rao's comment from a discussion offline.

pkg/snapstore/s3_snapstore.go
shreyas-s-rao: Thanks for capturing this @renormalize . On those lines, there should ideally be a config item to tell the LIST call whether to return all versions of snapshots, or only a specific version. While garbage collection requires all versions, restoration will only require specific versions of the snapshots.

I increasingly feel that the `List()` method needs to have a `ListOptions` parameter which we can be used to pass provider-specific configs to the call. So that we don't blow up the List method parameters with a bunch of provider-specific arguments.

ishan16696: I disagree with your point of `List` call should return the all versions of snapshot.
I agree with the restoration point of view of this comment: https://github.com/gardener/etcd-backup-restore/pull/818#discussion_r1905135692 that our restoration should only consider the oldest version of snapshot, so that our restoration should use the snapshots created/uploaded by backup-restore as only those will be the oldest one, and any modification/corruption on top of snapshot should be ignored.
But here, GC should also garbage collect snapshot created/uploaded by backup-restore, if there are different versions of snapshot present then it's a indication of attack. We should use that info to detect attack rather than GC all the corrupted snapshot as well and with no information about the attack.

pkg/snapstore/s3_snapstore.go
ishan16696: As discussed offline with @renormalize and @shreyas-s-rao, we decided to that our restoration should only consider the oldest version of snapshot, so that our restoration should use the snapshots created/uploaded by backup-restore as only those will be the oldest one, and any modification/corruption on top of snapshot should be ignored.

pkg/types/snapstore.go
ishan16696: cc @seshachalam-yv

pkg/snapstore/s3_snapstore.go
ishan16696: I have addressed this. PTAL.

pkg/snapstore/s3_snapstore.go
ishan16696: https://github.com/gardener/etcd-backup-restore/pull/818#discussion_r1910015643

pkg/snapstore/s3_snapstore.go
renormalize: Thanks for your explanation @ishan16696! I understand your train of thought now.
I guess I am okay with this.

renormalize (changes requested):
Thanks for the changes @ishan16696. I have a couple suggestions for the changes in 8141c0028c8490a23d85d74b331dc1e3affe2285.

Once these are addressed, it's an approval from my side.

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,47 +517,153 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+		in := &s3.ListObjectVersionsInput{
+			Bucket: aws.String(s.bucket),
+			Prefix: aws.String(prefix),
+		}
+
+		// allSnapKeysMapToCreationTime contains oldest snapshots mapped to their creation timestamp.
+		allSnapKeysMapToCreationTime := make(map[string]time.Time)
+
+		// allSnapKeysMapToVersionID contains oldest snapshots mapped to their versionID.
+		allSnapKeysMapToVersionID := make(map[string]string)
```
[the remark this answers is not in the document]
ishan16696: yes, make sense

pkg/snapstore/s3_snapstore.go
```diff
@@ -512,47 +517,153 @@ func (s *S3SnapStore) partUploader(wg *sync.WaitGroup, stopCh <-chan struct{}, s
...
+							// update both of the map.
+							allSnapKeysMapToVersionID[*version.Key] = *version.VersionId
+							allSnapKeysMapToCreationTime[*version.Key] = *version.LastModified
+						}
+					} else {
+						// snapshot key doesn't present in map
+						// then add the key to both of the maps.
+						allSnapKeysMapToVersionID[*version.Key] = *version.VersionId
+						allSnapKeysMapToCreationTime[*version.Key] = *version.LastModified
+					}
```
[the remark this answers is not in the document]
ishan16696: > There's only one situation where the map is not updated, when the map has a time.Time value present, and it is already (*version.LastModified).After(value)

yes, you are correct, I also thought that but to make code simple I choose this , so that anyone can understand code better as it improves the readability imo.

pkg/snapstore/s3_snapstore.go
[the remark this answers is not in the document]
ishan16696: made the changes, PTAL.

pkg/snapstore/s3_snapstore.go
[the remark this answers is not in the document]
renormalize: I think the conditional expression I suggested is pretty simple to understand, but I don't have strong opinions on this. We can close this thread once shreyas-s-rao also takes a look.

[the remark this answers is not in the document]
ishan16696: ok, I also don't have strong opinion. It could be my personal bias as I like separate condition for better readability.
@shreyas-s-rao can take a look on this.

pkg/snapstore/s3_snapstore.go
shreyas-s-rao: Ok, so you suggest that GC ignore the snapshots not created by etcdbr. That's a valid argument. But think of this scenario:
1. etcdbr uploads a set of snapshots
2. attacker creates a new version of a snapshot, so let's assume that `snapshot123` has a second version
3. GC runs, garbage collecting the oldest versions of the snapshots. The oldest version for `snapshot123` is deleted, while the new version of `snapshot123` remains.
4. Upon the next GC run, the only remaining version of `snapshot123` becomes the oldest snapshot now, and GC will anyway garbage collect it.

So the argument `GC should only garbage collect snapshots created by etcdbr` is no longer adhered to. So eventually, GC will anyway clean up all versions of a snapshot, with enough GC runs over time. Is that ok with you @ishan16696 ?

pkg/snapstore/s3_snapstore.go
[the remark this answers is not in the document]
ishan16696: yeah, I will make the change ... no worries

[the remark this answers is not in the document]
ishan16696: I have made the change. PTAL

pkg/snapstore/s3_snapstore.go
shreyas-s-rao: In an out-of-band discussion between me, @ishan16696 and @renormalize , we decided that the scope of this PR is to prevent modification/deletion of objects, and not handle addition of new malicious objects. Restoration is not affected, since the code only fetches the oldest versions of the objects, so addition of new object versions will not cause any disruptions to the functioning of etcd-backup-restore. Hence, we can mark this comment as resolved, and revisit it once we take up the task of handling addition of malicious object versions.

shreyas-s-rao (approved):
@ishan16696 thanks a lot for making the requested changes!
/lgtm

renormalize (approved):

PR #819, opened 2024-12-24
Read `emulatorEnabled` value for storage emulators from mounted secret instead of environment variables

shreyas-s-rao:
/area compliance security
/kind task

Read `emulatorEnabled` value for storage emulators from mounted secret instead of environment variables, to avoid mounting and exposing data from secrets as environment variables, even if they are non-credential fields.
Changes in this PR:
- [x] Read `emulatorEnabled` fields for providers Azure ABS and Google GCS from mounted secret files/JSON, instead of from environment variables
- [x] Refactor provider GCS snapstore construction code
- [x] Remove usage of env var `GOOGLE_EMULATOR_ENABLED`
- [x] Remove usage of env var `AZURE_EMULATOR_ENABLED`
- [x] Add `IsEmulatorEnabled` to `SnapstoreConfig` struct, to allow programatic downstream users of etcdbr code to set emulator enablement when creating `SnapstoreConfig`. For instance, etcd-druid can set this value to true when running local e2e tests against Azurite emulator.

Fixes #

/invite @anveshreddy18 @renormalize 

**Release note**:

```improvement user
Remove usage of environment variables to determine whether storage emulators are enabled.
```

### DIFF
--- a/pkg/snapstore/abs_snapstore.go
+++ b/pkg/snapstore/abs_snapstore.go
@@ -15,7 +15,6 @@ import (
 	"path"
 	"path/filepath"
 	"sort"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -91,10 +90,11 @@ type ABSSnapStore struct {
 }
 
 type absCredentials struct {
-	BucketName     string  `json:"bucketName"`
-	StorageAccount string  `json:"storageAccount"`
-	StorageKey     string  `json:"storageKey"`
-	Domain         *string `json:"domain,omitempty"`
+	BucketName      string  `json:"bucketName"`
+	StorageAccount  string  `json:"storageAccount"`
+	StorageKey      string  `json:"storageKey"`
+	Domain          *string `json:"domain,omitempty"`
+	EmulatorEnabled bool    `json:"emulatorEnabled,omitempty"`
 }
 
 // NewABSSnapStore creates a new ABSSnapStore using a shared configuration and a specified bucket
@@ -114,7 +114,7 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		domain = *absCreds.Domain
 	}
 
-	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain)
+	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain, absCreds.EmulatorEnabled)
 	if err != nil {
 		return nil, fmt.Errorf("failed to construct the blob service URL with error: %w", err)
 	}
@@ -146,21 +146,12 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 
 // ConstructBlobServiceURL constructs the Blob Service URL based on the activation status of the Azurite Emulator.
 // It checks the environment variable for emulator configuration and constructs the URL accordingly.
-// The function expects the following environment variable:
-// - AZURE_EMULATOR_ENABLED: Indicates whether the Azurite Emulator is enabled (expects "true" or "false").
-func ConstructBlobServiceURL(storageAccount, domain string) (string, error) {
+func ConstructBlobServiceURL(storageAccount, domain string, emulatorEnabled bool) (string, error) {
 	scheme := "https"
 
-	emulatorEnabled, ok := os.LookupEnv(EnvAzureEmulatorEnabled)
-	if ok {
-		isEmulator, err := strconv.ParseBool(emulatorEnabled)
-		if err != nil {
-			return "", fmt.Errorf("invalid value for %s: %s, error: %w", EnvAzureEmulatorEnabled, emulatorEnabled, err)
-		}
-		if isEmulator {
-			// TODO: going forward, use Azurite with HTTPS (TLS) communication
-			scheme = "http"
-		}
+	if emulatorEnabled {
+		// TODO: going forward, use Azurite with HTTPS (TLS) communication
+		scheme = "http"
 	}
 
 	return fmt.Sprintf("%s://%s.%s", scheme, storageAccount, domain), nil
--- a/pkg/snapstore/abs_snapstore.go
+++ b/pkg/snapstore/abs_snapstore.go
@@ -15,6 +15,7 @@ import (
 	"path"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


pkg/snapstore/gcs_snapstore.go
```diff
@@ -103,20 +100,32 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
 	return NewGCSSnapStoreFromClient(config.Container, config.Prefix, config.TempDir, config.MaxParallelChunkUploads, config.MinChunkSize, chunkDirSuffix, gcsClient), nil
 }
 
-func getGCSStorageAPIEndpoint(path string) (string, error) {
-	if _, err := os.Stat(path); err == nil {
-		data, err := os.ReadFile(path)
+func getGCSStorageAPIEndpoint(config *brtypes.SnapstoreConfig) (string, error) {
+	if gcsApplicationCredentialsPath, isSet := os.LookupEnv(getEnvPrefixString(config.IsSource) + envStoreCredentials); isSet {
+		storageAPIEndpointFilePath := path.Join(path.Dir(gcsApplicationCredentialsPath), fileNameStorageAPIEndpoint)
+		endpoint, err := os.ReadFile(storageAPIEndpointFilePath)
```
anveshreddy18, line 106: In this current flow, we are mandating that the `storageAPIEndpoint` field should always be present in the secret, otherwise this errors out. But that shouldn't be the case when using with real GCS isn't it? 

Previously, we were checking if the file is existing or not using `os.Stat` but now we are directly looking to read the file which may not always be present. Can you let me know if this is intended.

anveshreddy18 (commented):
Thanks @shreyas-s-rao for the PR. 

I've tested with fakegcs and it works as expected there. But when testing with real GCS, it's not working, checking that I found the below issue. Can you PTAL.

pkg/snapstore/gcs_snapstore.go
shreyas-s-rao: You're right @anveshreddy18 , I had missed this. Thanks for the find! Will push the fix.

shreyas-s-rao:
New addition as part of https://github.com/gardener/etcd-backup-restore/pull/819/commits/3d32db9af48ec7751b62fd0155e6f3b9211d07a0:
- Added field `IsEmulatorEnabled` to `SnapstoreConfig` struct, to allow programatic downstream users of etcdbr code to set emulator enablement when creating `SnapstoreConfig`. For instance, etcd-druid can set this value to true when running local e2e tests against Azurite emulator, since druid needs to construct its own snapstore client to access the object storage bucket when running e2e tests, to clean up any existing objects before starting the test run.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -60,21 +60,18 @@ const (
...
-		}
-		if endpoint != "" {
-			opts = append(opts, option.WithEndpoint(endpoint))
-			if emulatorConfig.enabled {
-				emulatorConfig.endpoint = endpoint
-			}
+	var opts []option.ClientOption // no need to explicitly set store credentials here since the Google SDK picks it up from the standard environment variable
+	var emulatorConfig gcsEmulatorConfig
+	emulatorConfig.enabled = config.IsEmulatorEnabled || isEmulatorEnabled(config)
+	endpoint, err := getGCSStorageAPIEndpoint(config)
```
anveshreddy18, line 67: Just a thought: Can we add a validation here that if emulator is enabled but endpoint is not present or is empty, then return an error? If agreed, we can do the same for Azure as well.

shreyas-s-rao: Sure, makes sense to add this validation.

pkg/snapstore/abs_snapstore.go
```diff
@@ -145,24 +147,13 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 
 // ConstructBlobServiceURL constructs the Blob Service URL based on the activation status of the Azurite Emulator.
 // It checks the environment variable for emulator configuration and constructs the URL accordingly.
```
renormalize, line 149: Can this comment, line 147 be removed/changed to reflect the changes?

renormalize (changes requested):
Thanks @shreyas-s-rao for this PR!
Everything looks good, just one tiny nit from me for the sake of correctness.

pkg/snapstore/abs_snapstore.go
```diff
@@ -113,7 +114,8 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		domain = *absCreds.Domain
 	}
 
-	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain)
+	emulatorEnabled := config.IsEmulatorEnabled || absCreds.EmulatorEnabled
+	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain, emulatorEnabled)
```
renormalize, line 118:
```suggestion
	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain, config.IsEmulatorEnabled || absCreds.EmulatorEnabled)
```

renormalize (changes requested):

pkg/snapstore/gcs_snapstore.go
renormalize: If the host is not present, we would get an error complaining that no host is present with a TCP connection failure which is quite self explanatory, in my opinion. I don't think this extra check is necessary.

With an empty "domain" file for ABS:

```
FATA[0015] failed to create snapstore from configured storage provider: failed to get properties of the container i585850-test with error: Get "https://devstoreaccount1./xxxxxx-test?restype=container": dial tcp: lookup devstoreaccount1.: no such host  actor=backup-restore-server
```

shreyas-s-rao: It would be better to throw the error early, than to try and decipher it from the URL in the error message.

pkg/snapstore/abs_snapstore.go
shreyas-s-rao: Sure, will do

pkg/snapstore/abs_snapstore.go
shreyas-s-rao: I deliberately used a separate variable `emulatorEnabled` for better readability.

pkg/snapstore/gcs_snapstore.go
shreyas-s-rao: @renormalize Btw, in ABS snapstore, we can never end up with URL `https://devstoreaccount1./xxxxxx-test?restype=container`, since domain is first defaulted to `brtypes.AzureBlobStorageGlobalDomain (blob.core.windows.net)`. So even if no domain is provided by user, the URL should still look like `https://devstoreaccount1.blob.core.windows.net/xxxxxx-test?restype=container`. And if emulator is enabled but domain isn't provided, the URL would look like `http://blob.core.windows.net/devstoreaccount1/xxxxxx-test?restype=container`. The `domain` will never be empty. It makes sense to throw the error early in the code, rather than try and figure out the malformed URL from the error logs.

renormalize (approved):
Thanks!

anveshreddy18 (approved):
Thanks for the PR @shreyas-s-rao.

PR #845, opened 2025-02-18
Fixed a bug which causes restoration to fail when first delta snapshot completely overlaps latest full snapshot.

### LINKED ISSUE #844: Handling of overlapping of first delta snapshot with full snapshot isn't done correctly.

/area disaster-recovery
/kind bug

**What happened**:
@unmarshall  and I observed that the Druid end-to-end tests were failing because the restoration process encountered an error. The backup-restore applied an incorrect delta event, resulting in the following error:
```
time="2025-02-17T09:45:57Z" level=error msg="Failed initialization: error while restoring corrupt data: failed to restore snapshot: mismatched event revision while applying delta snapshot, expected 11 but applied 15 " actor=backup-restore-server
```
Upon further investigation, we discovered that a separate handling mechanism for applying the first delta snapshot was introduced in this PR: https://github.com/gardener/etcd-backup-restore/pull/29/.
This mechanism was introduced to address the overlap of events between the full snapshot and the first delta snapshot. However, it fails to account for the case where there is a complete overlap of events between the delta snapshot and the full snapshot: 
https://github.com/gardener/etcd-backup-restore/blob/44b7d1b228c4d2aa92e8104957a1812302d8c4b9/pkg/snapshot/restorer/restorer.go#L512-L522

As a result, backup-restore re-applies some etcd events that should not be reapplied, causing the restoration verification checks to fail and ultimately leading to the restoration failure: 
https://github.com/gardener/etcd-backup-restore/blob/44b7d1b228c4d2aa92e8104957a1812302d8c4b9/pkg/snapshot/restorer/restorer.go#L649-L658

1. Start an etcd server.
2. Put some dummy data.
3. Start the backup-restore and make sure to take a full snapshot and delta snapshot which completely overlaps each other and timestamp of an overlapping delta snapshot should be later than that of the latest full snapshot.
5. example here, full snapshot with `0 to 11` revision completely overlaps with delta snapshot `8 to 11` revision with same timestamp.
```
Full-00000000-00000011-1739782988.gz
Incr-00000008-00000011-1739782988.gz
```
5. Trigger the restoration.

We seen several occurence of this issue in the past, Example: https://github.com/gardener/etcd-backup-restore/issues/763

ishan16696:

/area disaster-recovery
/kind bug

This PR fixes a bug which causes restoration to fail when first delta snapshot completely overlaps latest full snapshot.

Fixes #844 

**Release note**:

```bugfix user
Fixed a bug that caused restoration to fail when the first delta snapshot completely overlap with the latest full snapshot.
```

### DIFF
--- a/pkg/snapshot/restorer/restorer.go
+++ b/pkg/snapshot/restorer/restorer.go
@@ -509,6 +509,15 @@ func (r *Restorer) applyFirstDeltaSnapshot(clientKV client.KVCloser, snap *brtyp
 	}
 	lastRevision := resp.Header.Revision
 
+	if lastRevision == snap.LastRevision {
+		// there is no need to apply this fist delta snapshot
+		// as it's completely overlaps with full snapshot data.
+		// please refer: https://github.com/gardener/etcd-backup-restore/issues/844
+		r.logger.Infof("First delta snapshot %s found to be completely overlap with full snapshot with db revisions: %d", path.Join(snap.SnapDir, snap.SnapName), lastRevision)
+		r.logger.Info("Skipping this delta snapshot...")
+		return nil
+	}
+
 	var newRevisionIndex int
 	for index, event := range events {
 		if event.EtcdEvent.Kv.ModRevision > lastRevision {


ishan16696:
I'm yet to add the unit tests for this scenario.
/do-not-merge

ishan16696:
I will not be able to add unit tests for this, I can add unit tests for this later. Let's not block the fix.

shreyas-s-rao (approved):
@ishan16696 thanks for the fix! I've tested this and it fixes the issue of overlapping revisions as mentioned.

/lgtm

PR #857, opened 2025-03-22
Added a support of skipping/ignoring of any snapshot in an object lock enabled S3 bucket.

### LINKED ISSUE #843: To add a support of skipping/ignoring of any snapshot in an object lock enabled S3 bucket.

/area security
/kind enhancement

In this PR: https://github.com/gardener/etcd-backup-restore/pull/818, we have added support for handling AWS S3's object lock mechanism in the backup-restore, ensuring that snapshots are immutable. However, if users want backup-restore to skip or ignore certain snapshots, we need to implement a similar approach to what is currently available for other providers like GCS and ABS. These providers allow users to tag snapshots (with certain key), which they wish to be ignored by backup-restore . For more information, please refer to the [documentation](https://github.com/gardener/etcd-backup-restore/blob/master/docs/usage/enabling_immutable_snapshots.md#ignoring-snapshots-during-restoration)

**Why is this needed**:
When snapshots were mutable(if object lock isn't enabled), operators could simply delete any snapshots present in the object store, and subsequent restorations would not include them. However, once immutability is enabled, it is no longer possible to delete these snapshots. Hence, there should be a mechanism available for backup-restore to ignore any of snapshots during restoration.

ishan16696:

/area security
/kind technical-debt

This PR adds a support of skipping/ignoring of any snapshot in an object lock enabled S3 bucket using tag method.
It is achieved in a two-step:
  1. Put a delete marker on the top of object. Please refer [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html) for more information regarding deletion in versioning enabled S3 bucket.
  2. Then add a tag to snapshot object which you wish to be skipped/ignored during restoration:
    - **Key:** `x-etcd-snapshot-exclude`
    - **Value:** `true`

Fixes #843 

**Release note**:

```noteworthy user
Added a support for skipping or ignoring any snapshot in an object lock-enabled S3 bucket. For more information please refer to this doc: https://github.com/gardener/etcd-backup-restore/blob/master/docs/usage/enabling_immutable_snapshots.md#aws-s3
```

### DIFF
--- a/docs/usage/enabling_immutable_snapshots.md
+++ b/docs/usage/enabling_immutable_snapshots.md
@@ -425,7 +425,35 @@ After adding the annotation or tag, `etcd-backup-restore` will ignore these snap
 
 #### AWS S3
 
-- The approach of tagging snapshot objects to exclude them during restoration is not supported for `AWS S3` buckets.
+- Ignoring or skipping any snapshot object(s) present in AWS S3 bucket is a two-step process:
+  1. Put a delete marker on the top of object. Please refer [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html) for more information regarding deletion in versioning enabled S3 bucket.
+  2. Then add a tag to snapshot object which you wish to be skipped/ignored during restoration:
+  - **Key:** `x-etcd-snapshot-exclude`
+  - **Value:** `true`
+
+To add the tag:
+
+1. **Using the `aws` CLI**
+
+  ```bash
+  aws s3api put-object-tagging --bucket [BUCKET_NAME] --key [SNAPSHOT_PATH] --version-id [SNAPSHOT_VERSION_ID] --tagging '{"TagSet": [{ "Key": "x-etcd-snapshot-exclude", "Value": "true" }]}'
+  ```
+
+  **Example:**
+
+  ```bash
+  aws s3api put-object-tagging --bucket --key shoot1/etcd-main/v2/Incr-000000xx-000000yy-xxyyy.gz --version-id abcdefgh --tagging '{"TagSet": [{ "Key": "x-etcd-snapshot-exclude", "Value": "true" }]}'
+  ```
+
+2. **Using the AWS S3 Portal**
+
+   - Navigate to the bucket
+   - Locate the object which need to be skipped/ignored.
+   - Click on the object to view its details.
+   - Scroll down to **Tags** section, then add tag:
+      - **Key:** `x-etcd-snapshot-exclude`
+      - **Value:** `true`
+   - Save the changes.
 
 > Note: With S3 object lock, S3 versioning will automatically get enabled, it only prevent locked object versions from being permanently deleted.
 
--- a/docs/usage/enabling_immutable_snapshots.md
+++ b/docs/usage/enabling_immutable_snapshots.md
@@ -426,8 +426,8 @@ After adding the annotation or tag, `etcd-backup-restore` will ignore these snap
 #### AWS S3
 
 - Ignoring or skipping any snapshot object(s) present in AWS S3 bucket is a two-step process:
-  1. Put a delete marker on the top of object. Please refer [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html) for more information regarding deletion in versioning enabled S3 bucket.
-  2. Then add a tag to snapshot object which you wish to be skipped/ignored during restoration:
+  1. Put a delete marker on the top of snapshot object. Please refer [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html) for more information regarding deletion in versioning enabled S3 bucket.
+  2. Then tag the snapshot object which you wish to skip or ignore during restoration with following key and value:
   - **Key:** `x-etcd-snapshot-exclude`
   - **Value:** `true`
 
--- a/docs/usage/enabling_immutable_snapshots.md
+++ b/docs/usage/enabling_immutable_snapshots.md
@@ -263,6 +263,8 @@ To lock the immutability policy:
     1. **Governance mode**: Only users with special permissions can overwrite, delete or alter object lock settings.
     2. **Compliance mode**: No users(including root user) can overwrite, delete or alter object lock settings.
 
+> Note: With S3 object lock, S3 versioning will automatically get enabled, it only prevent locked object versions from being permanently deleted.
+
 #### Create S3 Buckets with object lock enabled
 
 > Note: With S3 object lock, S3 versioning will automatically get enabled.
@@ -454,8 +456,6 @@ To add the tag:
       - **Key:** `x-etcd-snapshot-exclude`
       - **Value:** `true`
    - Save the changes.
-
-> Note: With S3 object lock, S3 versioning will automatically get enabled, it only prevent locked object versions from being permanently deleted.
 
 ---
 
--- a/pkg/snapstore/s3_snapstore.go
+++ b/pkg/snapstore/s3_snapstore.go
@@ -792,9 +792,10 @@ func GetBucketImmutabilityTime(s *S3SnapStore) (bool, *int64, error) {
 	return false, nil, fmt.Errorf("got nil object lock configuration")
 }
 
+// isSnapshotMarkToBeIgnored checks whether snapshot object key with given versionID is marked to be ignored or not.
 func isSnapshotMarkToBeIgnored(client s3iface.S3API, bucketName, key, versionID string) bool {
 	out, err := client.GetObjectTagging(&s3.GetObjectTaggingInput{
-		Bucket:    &bucketName,
+		Bucket:    aws.String(bucketName),
 		Key:       aws.String(key),
 		VersionId: aws.String(versionID),
 	})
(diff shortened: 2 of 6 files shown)


ishan16696:
### Testing results:

Here's screenshot of my S3 bucket:
Screenshot 2025-03-23 at 5 33 48 PM

For testing, I have marked the `Full-00000000-00000031-1742649050.gz` and `Full-00000000-00000033-1742656727.gz` to get skipped or ignored during restoration.


### Result

Backup-restore ignored both of full snapshot as expected and fall back to latest full snapshot which wasn't marked to be ignored to trigger restoration.

Logs of backup-restore during restoration:
```
time="2025-03-23T12:00:58Z" level=info msg="Snapshot: shoot1/etcd-main/v2/Full-00000000-00000031-1742649050.gz with versionID: 55cevktsV4V0mEGS6mfgqVbZGvV6qPkK is marked to be ignored"
time="2025-03-23T12:00:58Z" level=info msg="Snapshot: shoot1/etcd-main/v2/Full-00000000-00000033-1742656727.gz with versionID: CHFOZ7C.tjlNzUWdC6uRxSd2S4WcGBk_ is marked to be ignored"
time="2025-03-23T12:00:58Z" level=info msg="Removing directory(/var/etcd/data/new.etcd.part)."
time="2025-03-23T12:00:58Z" level=info msg="Creating temporary directory /var/etcd/data/restoration.temp for persisting full and delta snapshots locally." actor=restorer
time="2025-03-23T12:00:58Z" level=info msg="Restoring from base snapshot: Full-00000000-00000017-1739886840.gz" actor=restorer
```


Even though there were full snapshot `Full-00000000-00000031-1742649050.gz` and `Full-00000000-00000033-1742656727.gz` present in bucket but as both of them marked to be ignored, so for the restoration backup-restore ignored both of them and restoration got triggered by using `Full-00000000-00000017-1739886840.gz`.

pkg/snapstore/s3_snapstore.go
```diff
@@ -605,7 +608,29 @@ func (s *S3SnapStore) List(_ bool) (brtypes.SnapList, error) {
 			return nil, err
 		}
 
+		// traverse all the deletion markers present(if any) in the bucket
+		if err := s.client.ListObjectVersionsPages(in, func(page *s3.ListObjectVersionsOutput, lastPage bool) bool {
```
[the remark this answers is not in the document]
ishan16696: yes, I think that should be possible, I will test and make necessary changes. thanks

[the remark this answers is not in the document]
ishan16696: addressed, PTAL

renormalize (approved):

docs/usage/enabling_immutable_snapshots.md
```diff
@@ -425,7 +425,35 @@ After adding the annotation or tag, `etcd-backup-restore` will ignore these snap
...
+
+   - Navigate to the bucket
+   - Locate the object which need to be skipped/ignored.
+   - Click on the object to view its details.
+   - Scroll down to **Tags** section, then add tag:
+      - **Key:** `x-etcd-snapshot-exclude`
+      - **Value:** `true`
+   - Save the changes.
 
 > Note: With S3 object lock, S3 versioning will automatically get enabled, it only prevent locked object versions from being permanently deleted.
```
anveshreddy18, line 458: Is the line
> Note: With S3 object lock, ...

out of place? It immediately comes after the section `Using the AWS S3 portal to tag the snapshot` but doesn't connect to it 🤔 

Also I'm not sure if with `Object lock` here, you mean the object getting locked or the `Object level` locking option? Can you please clarify it a bit here.

pkg/snapstore/s3_snapstore.go
```diff
@@ -577,6 +577,9 @@ func (s *S3SnapStore) List(_ bool) (brtypes.SnapList, error) {
 		// allSnapKeysMapToCreationTime contains oldest snapshots keys mapped to their versionID and creation timestamp.
 		allSnapKeyMapToSnapshotInfo := make(map[string]*snapshotMetaInfo)
```
anveshreddy18, line 579:
```suggestion
		// allSnapKeyMapToSnapshotInfo contains oldest snapshots keys mapped to their versionID and creation timestamp.
		allSnapKeyMapToSnapshotInfo := make(map[string]*snapshotMetaInfo)

```
The constant `allSnapKeysMapToCreationTime` doesn't exist anywhere. I guess you changed the name after that but missed the comment.

docs/usage/enabling_immutable_snapshots.md
```diff
@@ -425,7 +425,35 @@ After adding the annotation or tag, `etcd-backup-restore` will ignore these snap
 
 #### AWS S3
 
-- The approach of tagging snapshot objects to exclude them during restoration is not supported for `AWS S3` buckets.
+- Ignoring or skipping any snapshot object(s) present in AWS S3 bucket is a two-step process:
+  1. Put a delete marker on the top of snapshot object. Please refer [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html) for more information regarding deletion in versioning enabled S3 bucket.
+  2. Then tag the snapshot object which you wish to skip or ignore during restoration with following key and value:
+  - **Key:** `x-etcd-snapshot-exclude`
+  - **Value:** `true`
```
anveshreddy18, line 432: Can you help me understand this a little better? why does the delete marker version has to be present on the snapshot to exclude a snapshot, can't we simply add the tag to skip that snapshot? And let say we do put the delete marker as you mentioned, then the restoration should skip it automatically since the `GET` request will return a 404 error right? then why do we need to have the extra step of adding the tag as well? 

One more question: Once we put the delete marker and place the tag, then someone has to remove the delete marker (so that the current version points to the actual snapshot version before placing the delete marker) after the restoration operation is done, so does `backup-restore` do that?

pkg/snapstore/snapstore_test.go
```diff
@@ -713,6 +713,44 @@ var _ = Describe("Get Immutability time for S3 bucket", func() {
 	})
 })
 
+var _ = Describe("S3 snapshot object mark to be ignored/skipped", func() {
```
anveshreddy18, line 716:
```suggestion
var _ = Describe("S3 snapshot object is marked to be ignored/skipped", func() {
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -766,3 +786,23 @@ func GetBucketImmutabilityTime(s *S3SnapStore) (bool, *int64, error) {
 
 	return false, nil, fmt.Errorf("got nil object lock configuration")
 }
+
+// IsSnapshotMarkToBeIgnored checks whether snapshot object key with given versionID is tagged to be ignored or not.
+func IsSnapshotMarkToBeIgnored(client s3iface.S3API, bucketName, key, versionID string) bool {
```
anveshreddy18, line 791:
```suggestion
// IsSnapshotMarkedToBeIgnored checks whether snapshot object key with given versionID is tagged to be ignored or not.
func IsSnapshotMarkedToBeIgnored(client s3iface.S3API, bucketName, key, versionID string) bool {
```
Minor: Marked feels right instead of Mark

anveshreddy18 (changes requested):

docs/usage/enabling_immutable_snapshots.md
ishan16696: > It immediately comes after the section Using the AWS S3 portal to tag the snapshot but doesn't connect to it 

yeah, agree with you I was also thinking the same .... may be I will move to some other more relevant place.

> Also I'm not sure if with Object lock here, you mean the object getting locked or the Object level locking option? Can you please clarify it a bit here.

I got your point of ambiguity as you're trying to map things with other providers but S3 official doc uses term `object lock` (it doesn't use the bucket lock). So I decided to stick with official doc terms.
It says S3 supports object lock on bucket level as well as on object level.
Offical doc link, itself starts with S3 object lock https://aws.amazon.com/s3/features/object-lock/

docs/usage/enabling_immutable_snapshots.md
ishan16696: ok, If you only put the delete marker on top of one of the snapshot, then backup-restore `LIST` call will still consider that versioned snapshot while restoration (which uses LIST call to get list of snapshot).
This is done as preventive measure against attack as Azure and GCS won't allow you to delete any snapshot but S3 allows you to put delete markers(not permanently deletion), so backup-restore ignore all delete marker and still consider the snapshot object for restoration.

Now, moving to ignoring of snapshots:
1. The challenge which we faced is, to get a tag of all snapshot is not a 1 API call as tag of snapshot isn't  present in metadata of snapshot.
2. First you have to get all snapshot with versionID then their is a separete API call to get tag, so `n` snapshot can result in `n` api call just to check presence of any tag, which is not good.
3. Hence, we decided to go with this approach: if their is a delete marker present for any snapshot then only we check the tag for that snapshot, by this way we are reducing the no. of S3 api call.

I hope now you get the approach.

docs/usage/enabling_immutable_snapshots.md
ishan16696: > ( not to be confused with object level lock)

no, IMO that will cause more confusion as if somebody want to enable S3 object lock then I will assume that he/she should be familiar with S3 terminologies.

docs/usage/enabling_immutable_snapshots.md
ishan16696: > then someone has to remove the delete marker (so that the current version points to the actual snapshot version before placing the delete marker) after the restoration operation is done, so does backup-restore do that?

why someone want to remove the delete marker if they have decided to mark certain snapshot to get ignore, I leave those decision to operator. If they want they can remove the delete marker.

anveshreddy18 (approved):

renormalize (approved):

PR #743, opened 2024-05-22
Enable e2e tests using emulators for cloud providers [ aws, gcp, azure ]

anveshreddy18:
**What this PR does / why we need it**:

This PR revives the currently non functional e2e tests ( previously `integrationcluster` tests ) and enable them to run with real infra and `emulators` of 3 major cloud providers i.e AWS, GCP, AZURE. The reason for using emulators is to reduce the cost of using real infra every time a PR is raised because with the enablement of e2e tests with this PR, these tests will be run as prow job for every PR. 

To run these e2e tests locally, just need to do `make ci-e2e-kind PROVIDERS="aws,gcp,azure"`. The provider list can be edited to run tests with only those specific providers.

This PR also upgrades the helm version from `k8.io/helm/v2` to `helm.sh/helm/v3`, with this changes a lot of indirect dependency graph, as a result, the `go.mod` and `vendor/` changes.

**Which issue(s) this PR fixes**:
Fixes #722

**Special notes for your reviewer**:

**Release note**:

```improvement developer
Revive e2e tests (previously `integrationcluster` tests) and enable them to run with AWS, GCP, AZURE cloud providers and their emulators.
```

PR #862, opened 2025-04-07
Changes in e2e tests setup required for running them in prow job environment

anveshreddy18:

/area testing
/kind test

The PR #743 restored the e2e tests and made them run with cloud providers(real infra & their emulators), to make that happen it installs the provider CLIs such as `awscli`, `az` and `gcloud` to interact with both real providers and emulators. The current code uses `pip3` to install these CLI tools but we plan to run these tests in prow environment where they run inside a Debian linux container where pip3 is not present by default. So the PR makes changes to install these tools using `apt` pkg manager which is native to Debian environments. 

This PR also fixes a failing e2e test case where the `pgrep` command to get the PID of the `etcd-wrapper` process is not working as intended, changed it to use combination of `ps`, `grep` and `awk` through piping.

Fixes #

**Release note**:

```other developer
Updated e2e test setup to install provider CLIs using `apt` package manager and fixed a failing test case involving `etcd-wrapper` process PID retrieval.
```

### DIFF
--- a/hack/e2e-test/run-e2e-test.sh
+++ b/hack/e2e-test/run-e2e-test.sh
@@ -192,9 +192,15 @@ function create_gcp_container() {
   echo "Setting up GCS infrastructure..."
   echo "Creating test bucket..."
   if [[ -z ${GOOGLE_EMULATOR_HOST:-""} ]]; then
-    gsutil mb "gs://${TEST_ID}"
+    if ! gsutil mb "gs://${TEST_ID}"; then
+      echo "Failed to create GCS bucket ${TEST_ID}."
+      return 1
+    fi
   else 
-    gsutil -o "Credentials:gs_json_host=127.0.0.1" -o "Credentials:gs_json_port=4443" -o "Boto:https_validate_certificates=False" mb "gs://${TEST_ID}"
+    if ! gsutil -o "Credentials:gs_json_host=127.0.0.1" -o "Credentials:gs_json_port=4443" -o "Boto:https_validate_certificates=False" mb "gs://${TEST_ID}"; then
+      echo "Failed to create GCS bucket ${TEST_ID}."
+      return 1
+    fi
   fi
   echo "Successfully created test bucket."
   echo "Setting up GCS infrastructure completed."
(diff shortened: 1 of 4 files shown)


hack/e2e-test/run-e2e-test.sh
```diff
@@ -243,28 +261,34 @@ function authorize_gcloud() {
   fi
 }
 
-# setup_gcp_e2e sets up the GCP infrastructure for the e2e tests including deploying fake-gcs and/or installing the gsutil
+# setup_gcp_e2e sets up the GCP infrastructure for the e2e tests including deploying installing the gcloud sdk and/or fake-gcs
 function setup_gcp_e2e() {
+  setup_gcloud
```
ishan16696, line 266: Is it fine to install whole gcloud sdk instead of just gsutil ?
Is there a way to install only gsutil or we have to install gcloud sdk ?

test/e2e/backup_test.go
```diff
@@ -300,7 +300,7 @@ var _ = Describe("Backup", func() {
 		Context("when data is corrupt", func() {
 			It("should restore data from latest snapshot", func() {
 				testDataCorruptionRestoration := func() {
-					cmd := "export ETCD_PID=$(pgrep etcd-wrapper) && rm -r proc/${ETCD_PID}/root/var/etcd/data/new.etcd/member"
+					cmd := "export ETCD_PID=$(ps -A | grep 'etcd-wrapper' | awk '{print $1}') && rm -r proc/${ETCD_PID}/root/var/etcd/data/new.etcd/member"
```
ishan16696, line 303: do you know why `pgrep` is not working as expected ?
```
I have no name!@etcd-main-0:/$ ps -A | grep etcd-wrapper | awk '{print $1}'
7
I have no name!@etcd-main-0:/$ ps -A | pgrep etcd-wrapper
7
```

ishan16696 (commented):
Overall looks good, just few doubts:

hack/e2e-test/run-e2e-test.sh
anveshreddy18: We have to install whole gcloud sdk as `gcloud` is used for authentication and `gsutil` inherits that and can do the required CRUID ops on the bucket.

test/e2e/backup_test.go
anveshreddy18: I'm not sure but I get the following error when I try to do it. 

```bash
I have no name!@main-aws-etcd-0:/$ ps -A | pgrep etcd-wrapper
assertion failed [true_path_length_other >= 0]: true_path_length_other underflowed!
(ThreadContextFcntl.cpp:179 is_rosetta_process)
 Trace/breakpoint trap
 ```
 This could be because the `pgrep` process is built for x86_64, and docker uses rosetta emulator to convert it into arm64 arch. It worked previously so I guess it's a docker emulator problem mostly, but not sure why other binaries like `ps` work. So something could be wrong with the built of `pgrep` itself. 
 
 This is why I wanted to eliminate this dependency by not using `pgrep` altogether. Wdyt?

ishan16696: ok

PR #869, opened 2025-04-24
Clean up snapshot temp directory during initialization

shreyas-s-rao:

/area security compliance
/kind task cleanup

Clean up snapshot temp directory during initialization, to ensure any previously created temp files for snapshot upload will be cleaned up, which helps in two ways (a) to remove unnecessary files which may not have been properly cleaned up in a previous run of etcdbr and (b) these older files might have old file permissions before umask of 0077 was set, which might violate [DISA STIG](https://www.stigviewer.com/stig/kubernetes/) standards, hence cleaning up these files is beneficial from a security perspective as well.

Fixes #

**Release note**:

```other user
Clean up snapshot temp directory during initialization.
```

### DIFF
--- a/pkg/snapstore/utils.go
+++ b/pkg/snapstore/utils.go
@@ -46,24 +46,6 @@ func GetSnapstore(config *brtypes.SnapstoreConfig) (brtypes.SnapStore, error) {
 		return nil, fmt.Errorf("storage container name not specified")
 	}
 
-	if len(config.TempDir) == 0 {
-		config.TempDir = path.Join("/tmp")
-	}
-	if _, err := os.Stat(config.TempDir); err != nil {
-		if !os.IsNotExist(err) {
-			return nil, fmt.Errorf("failed to get file info of temporary directory %s: %v", config.TempDir, err)
-		}
-
-		logrus.Infof("Temporary directory %s does not exist. Creating it...", config.TempDir)
-		if err := os.MkdirAll(config.TempDir, 0700); err != nil {
-			return nil, fmt.Errorf("failed to create temporary directory %s: %v", config.TempDir, err)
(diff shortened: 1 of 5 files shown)


ishan16696:
@shreyas-s-rao please resolve the merge conflicts.

shreyas-s-rao:
@ishan16696 done. PTAL

pkg/snapstore/utils.go
```diff
@@ -46,24 +46,6 @@ func GetSnapstore(config *brtypes.SnapstoreConfig) (brtypes.SnapStore, error) {
...
-
-		logrus.Infof("Temporary directory %s does not exist. Creating it...", config.TempDir)
-		if err := os.MkdirAll(config.TempDir, 0700); err != nil {
-			return nil, fmt.Errorf("failed to create temporary directory %s: %v", config.TempDir, err)
-		}
-	}
-
-	if config.MaxParallelChunkUploads <= 0 {
-		config.MaxParallelChunkUploads = 5
-	}
```
ishan16696, line 65: any reason for removing this check ?

ishan16696 (commented):
overall looks good, I have only 1 doubt @shreyas-s-rao

pkg/snapstore/utils.go
shreyas-s-rao: We already validate this field here - https://github.com/gardener/etcd-backup-restore/blob/2fcbadd7e6d57b302e8ae3b9c0913bd539b1b693/pkg/types/snapstore.go#L225-L227

So there's no point in checking again and setting to a default value, because it's already validated during initialization of the object.

ishan16696 (approved):
LGTM!!

PR #892, opened 2025-07-08
[GEP-26] Add support for web identity authentication to S3 snapstore

vpnachev:
/area backup security ipcei
/kind enhancement

Add support for web identity authentication to S3 snapstore

Part of https://github.com/gardener/gardener/issues/9586

cc @dimityrmirchev

**Release note**:

```other operator
The container base image has been updated to debian12. 
```

```feature operator
S3 snapstore now support web identity authentication.
To be enabled, the `roleARN` and `token` credentials files or json fields must be set, while `accessKeyID` and `secretAccessKey` has to be unset.
```

### DIFF
--- a/docs/development/local_setup.md
+++ b/docs/development/local_setup.md
@@ -131,17 +131,10 @@ s3credentials
 
 Sticking to the S3 example, the `.json` file which contains the credentials should look like so:
 
-> [!NOTE]
-> Similarly to credentials [credentials in a directory](#credentials-in-a-directory), not all fields must be set at once. The `region` field is always required.
-> If static credentials are used, then the fields `accessKeyID` and `secretAccessKey` must be set,
-> otherwise, when web identity role is used the fields `roleARN` and `tokenPath` must be set.
-
 ```json
 {
-  "region": "<region>",
   "accessKeyID": "<accessKeyID>",
+  "region": "<region>",
   "secretAccessKey": "<secretAccessKey>",
-  "roleARN": "<roleARN>",
-  "tokenPath": "<path-to-jwt>",
 }
 ```
--- a/pkg/snapstore/s3_snapstore.go
+++ b/pkg/snapstore/s3_snapstore.go
@@ -765,8 +765,8 @@ func GetS3CredentialsLastModifiedTime() (time.Time, error) {
 			webIdentityCredentialsFiles = []string{filepath.Join(dir, "region"), filepath.Join(dir, "roleARN"), filepath.Join(dir, "token")}
 		)
 
-		awsTimestamp, err := getLatestCredentialsModifiedTime(staticCredentialsFiles)
-		if err == nil {
+		awsTimestamp, err1 := getLatestCredentialsModifiedTime(staticCredentialsFiles)
+		if err1 == nil {
 			return awsTimestamp, nil
 		}
 
@@ -775,7 +775,7 @@ func GetS3CredentialsLastModifiedTime() (time.Time, error) {
 			return awsTimestamp, nil
 		}
 
-		return time.Time{}, fmt.Errorf("failed to get AWS credential timestamp from the directory %v with error: %w", dir, errors.Join(err, err2))
+		return time.Time{}, fmt.Errorf("failed to get AWS credential timestamp from the directory %v with error: %w", dir, errors.Join(err1, err2))
 	}
 
 	if filename, isSet := os.LookupEnv(awsCredentialJSONFile); isSet {
--- a/docs/deployment/getting_started.md
+++ b/docs/deployment/getting_started.md
@@ -1,6 +1,6 @@
 # Getting started
 
-The binary builds are not published with each release, but it is pretty straight forward to build it by following the steps mentioned [here](../development/local_setup.md#build). But we do publish the docker image with each release, please check the [release page](https://github.com/gardener/etcd-backup-restore/releases) for the same. Currently, release docker images are pushed to `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl` to container registry.
+The binary builds are not published with each release, but it is pretty straight forward to build it by following the steps mentioned [here](../development/local_setup.md#build). But we do publish the docker image with each release, please check the [release page](https://github.com/gardener/etcd-backup-restore/releases) for the same. Currently, release docker images are pushed to `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl` container  repository.
 
 ## Usage
 
@@ -13,7 +13,9 @@ The procedure to provide credentials to access the cloud provider object store v
 ### Passing Credentials
 
 * For `AWS S3`:
-   1. The secret file should be provided, and the file path should be made available as an environment variable: `AWS_APPLICATION_CREDENTIALS`.
+   1. The secret file should be provided, and the file path should be made available as an environment variable: `AWS_APPLICATION_CREDENTIALS`. AWS S3 supports two alternative authentication options:
+      1. Static credentials: `accessKeyID` and `secretAccessKey` must be set.
+      1. Web Identity Role: `roleARN` and `tokenPath` must be set.
    2. For `S3-compatible providers` such as MinIO, `endpoint`, `s3ForcePathStyle`, `insecureSkipVerify` and `trustedCaCert`, can also be made available in an above file to configure the S3 client to communicate to a non-AWS provider.
    3. To enable Server-Side Encryption using Customer Managed Keys for `S3-compatible providers`, use `sseCustomerKey` and `sseCustomerAlgorithm` in the credentials file above. For example, `sseCustomerAlgorithm` could be set to `AES256`, and correspondingly the `sseCustomerKey` is set to a valid AES-256 key.
 
--- a/example/storage-provider-secrets/01-aws-s3-storage-secret.yaml
+++ b/example/storage-provider-secrets/01-aws-s3-storage-secret.yaml
@@ -8,6 +8,8 @@ data:
   region: ZXUtd2VzdC0x # eu-west-1
   accessKeyID: YWRtaW4= # admin
   secretAccessKey: YWRtaW4= # admin
+# roleARN: YXJuOmF3czppYW06OjEyMzQ1Njc4OTAxMjpyb2xlL3JvbGUtbmFtZQ== # arn:aws:iam::123456789012:role/role-name
+# token: base64(<jwt>) # base64 encoded JSON Web Token
 # endpoint: # used for S3 compatible providers
 # s3ForcePathStyle: # set to `true` for S3 compliant providers
 # insecureSkipVerify: # set to `true` for S3 private deployments to skip certificate check


docs/operations/generating_ssl_certificates.md
```diff
@@ -66,7 +66,7 @@ mkdir openssl && cd openssl
 #### Generating CA cert bundle
 
 ```console
-openssl genrsa -out ca.key 2048
+openssl genrsa -out ca.key 4096
```
dimityrmirchev, line 69: In gardener we use 3072 bits ([reference](https://github.com/gardener/gardener/blob/2c677126795dd8587acdec9f2e8fc651ccde7160/pkg/utils/secrets/certificate.go#L114)). Should we use the same values here?

pkg/snapstore/generic_s3_snapstore.go
```diff
@@ -35,13 +39,23 @@ func newGenericS3FromAuthOpt(bucket, prefix, tempDir string, maxParallelChunkUpl
...
 	)
+
+	if ao.accessKeyID != "" {
+		credentialsProvider = credentials.NewStaticCredentialsProvider(ao.accessKeyID, ao.secretAccessKey, "")
+	} else {
+		credentialsProvider = stscreds.NewWebIdentityRoleProvider(sts.NewFromConfig(aws.Config{Region: ao.region}), ao.roleARN, stscreds.IdentityTokenFile(ao.tokenPath))
+	}
+	cfgOpts = append(cfgOpts, config.WithCredentialsProvider(aws.NewCredentialsCache(credentialsProvider)))
+
+	cfg, err := config.LoadDefaultConfig(context.TODO(), cfgOpts...)
```
dimityrmirchev: Should we change this to a Background context?

```suggestion
	cfg, err := config.LoadDefaultConfig(context.Background(), cfgOpts...)
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -318,6 +336,14 @@ func readAWSCredentialFromDir(dirname string) (*awsCredentials, error) {
 				return nil, err
 			}
 			awsConfig.SSECustomerAlgorithm = ptr.To(string(data))
+		case "roleARN":
+			roleARN, err := os.ReadFile(dirname + "/roleARN") // #nosec G304 -- this is a trusted file, obtained via user input.
```
dimityrmirchev: I think we can use [filepath.Join](https://pkg.go.dev/path/filepath#Join) here.

```suggestion
			roleARN, err := os.ReadFile(filepath.Join(dirname, "roleARN")) // #nosec G304 -- this is a trusted file, obtained via user input.
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -318,6 +336,14 @@ func readAWSCredentialFromDir(dirname string) (*awsCredentials, error) {
...
+			if err != nil {
+				return nil, err
+			}
+			awsConfig.RoleARN = string(roleARN)
+		case "token":
+			awsConfig.TokenPath = dirname + "/token"
```
dimityrmirchev:
```suggestion
			awsConfig.TokenPath = filepath.Join(dirname, "token ")
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -733,15 +759,23 @@ func GetS3CredentialsLastModifiedTime() (time.Time, error) {
...
-		if err != nil {
-			return time.Time{}, fmt.Errorf("failed to get AWS credential timestamp from the directory %v with error: %w", dir, err)
+
+		awsTimestamp, err2 := getLatestCredentialsModifiedTime(webIdentityCredentialsFiles)
+		if err2 == nil {
+			return awsTimestamp, nil
 		}
-		return awsTimeStamp, nil
+
+		return time.Time{}, fmt.Errorf("failed to get AWS credential timestamp from the directory %v with error: %w, %w", dir, err, err2)
```
dimityrmirchev:
```suggestion
		return time.Time{}, fmt.Errorf("failed to get AWS credential timestamp from the directory %v with error: %w", dir, errors.Join(err, err2))
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -757,10 +791,12 @@ func GetS3CredentialsLastModifiedTime() (time.Time, error) {
...
 
 func isAWSConfigEmpty(config *awsCredentials) error {
-	if len(config.AccessKeyID) != 0 && len(config.Region) != 0 && len(config.SecretAccessKey) != 0 {
+	if len(config.Region) != 0 &&
+		(len(config.AccessKeyID) != 0 && len(config.SecretAccessKey) != 0 ||
+			len(config.RoleARN) != 0 && len(config.TokenPath) != 0) {
 		return nil
 	}
-	return fmt.Errorf("aws s3 credentials: region, secretAccessKey or accessKeyID is missing")
+	return fmt.Errorf("aws s3 credentials: region, and secretAccessKey and accessKeyID, or roleARN and token are missing")
```
dimityrmirchev: Can we split this error to multiple smaller errors? Something like

```golang
if len(config.Region) == 0 {
	return fmt.Errorf("aws s3 credentials: region is missing")
}

if len(config.AccessKeyID) != 0 && len(config.SecretAccessKey) != 0 {
	return nil
}

if len(config.RoleARN) != 0 && len(config.TokenPath) != 0 {
	return nil
}

return fmt.Errorf("aws s3 credentials: either set secretAccessKey and accessKeyID or roleARN and token")
```

chart/etcd-backup-restore/values.yaml
```diff
@@ -108,6 +108,8 @@ backup:
   #   s3ForcePathStyle: "true" # optional
   #   sseCustomerKey: aes-256-sse-customer-key # optional
   #   sseCustomerAlgorithm: aes-256-sse-customer-algorithm # optional
+  #   token: <JWT>
```
dimityrmirchev: Do you have any insights where is this chart used?

I feel like the token should not be passed through the values file as this token need to be renewed. It would probably make more sense that this is a path to a token then something should be responsible for renewing the token.

dimityrmirchev (changes requested):
Thanks! I left some small comments/questions.

chart/etcd-backup-restore/values.yaml
vpnachev: I have the same concerns, not sure whether the chart is used anywhere. Apart from the Gardener scenario with Workload Identity, maybe it is possible to use Service Account tokens, then probably the chart should be re-worked to mount projected Service Account token in this file. 
Let's see maintainers opinion on that, I am fine to re-work it with SA tokens.

docs/operations/generating_ssl_certificates.md
vpnachev: I believe such ~certificates~ keys are not created at scale and it should not be a problem the longer time it takes to generated 4096 bits key.

pkg/snapstore/generic_s3_snapstore.go
vpnachev: No hard opinion, it looks like the repo is more widely using the `TODO` context instead of `Background`, that is also what the S3 snapstore is using, but I am fine to change it to `Background`.

pkg/snapstore/s3_snapstore.go
vpnachev: Nice suggestion, should I change all other occurrances of `dirPath+"/<something>"`, too?

pkg/snapstore/s3_snapstore.go
vpnachev: Nice suggestion, I will do it.

pkg/snapstore/s3_snapstore.go
dimityrmirchev: Yes, why not. I think that this would be an improvement.

docs/operations/generating_ssl_certificates.md
dimityrmirchev: Seems fair.

pkg/snapstore/generic_s3_snapstore.go
dimityrmirchev: If the repo uses `TODO` maybe we leave it like that for consistency. I leave the decision to you.

chart/etcd-backup-restore/values.yaml
ishan16696: > Do you have any insights where is this chart used?

charts are right now used to run `cluster based integration tests` https://github.com/gardener/etcd-backup-restore/blob/master/docs/development/tests.md#run-the-tests. 
Moreover, any community user can also consume etcd-backup-restore via charts (with limited functionality like scale-up feature isn't available) if they don't want to consume etcd-backup-restore via druid

vpnachev: c8da0280f05d21f9f1a351c491e1ff19172f3443 adds the option service account token to be used for authentication.

vpnachev:
@gardener/etcd-druid-maintainers kindly review this PR

docs/development/local_setup.md
```diff
@@ -110,24 +110,38 @@ Export the following environment variables according to your host OS:
 
 Taking the example of S3 again here, the following is the necessary structure for a directory which holds the S3 credential files (extra credentials correspond to extra files, which are not compulsory).
 
+> [!NOTE]
+> `etcdbrctl` supports different authentication methods, therefore not all files must be set. The always required file is `region`.
+> If static credentials are used, then the files `accessKeyID` and `secretAccessKey` must be set.
+> Alternatively, if web identity role is used, then the files `roleARN` and `token` must be set.
```
ishan16696, line 116: can you please add a example in this file: https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/01-aws-s3-storage-secret.yaml

docs/development/local_setup.md
```diff
@@ -110,24 +110,38 @@ Export the following environment variables according to your host OS:
...
 #### Credentials in a JSON file
 
 Sticking to the S3 example, the `.json` file which contains the credentials should look like so:
 
-```console
+> [!NOTE]
+> Similarly to credentials [credentials in a directory](#credentials-in-a-directory), not all fields must be set at once. The `region` field is always required.
+> If static credentials are used, then the fields `accessKeyID` and `secretAccessKey` must be set,
+> otherwise, when web identity role is used the fields `roleARN` and `tokenPath` must be set.
+
```
ishan16696: I don't think this is the correct place to mention this as here we are just giving an example for local-setup.
you can mention these things here: https://github.com/gardener/etcd-backup-restore/blob/master/docs/deployment/getting_started.md#passing-credentials

pkg/snapstore/s3_snapstore.go
```diff
@@ -733,15 +759,23 @@ func GetS3CredentialsLastModifiedTime() (time.Time, error) {
...
-		credentialFiles := []string{"accessKeyID", "region", "secretAccessKey"}
-		for i := range credentialFiles {
-			credentialFiles[i] = filepath.Join(dir, credentialFiles[i])
+		// either static or web identity credential files should be set
+		var (
+			staticCredentialsFiles      = []string{filepath.Join(dir, "region"), filepath.Join(dir, "accessKeyID"), filepath.Join(dir, "secretAccessKey")}
+			webIdentityCredentialsFiles = []string{filepath.Join(dir, "region"), filepath.Join(dir, "roleARN"), filepath.Join(dir, "token")}
+		)
+
+		awsTimestamp, err := getLatestCredentialsModifiedTime(staticCredentialsFiles)
```
ishan16696:
```suggestion
		awsTimestamp, err1 := getLatestCredentialsModifiedTime(staticCredentialsFiles)
```
you're using `err2` in next line, either change that or change this ... no strong opinion

ishan16696 (changes requested):
few nits:

docs/development/local_setup.md
vpnachev: done

ishan16696 (approved):
LGTM!!

PR #907, opened 2025-08-22
Check for snapshot names with the `.final` suffix trimmed in the destination store before upload.

### LINKED ISSUE #906: `.final` snapshots are not checked for presence in destination object store before being copied.

/area backup
/area control-plane-migration
/kind bug

**What happened**:

The `copy` command copies all backups from source to destination object stores.

Certain snapshots are stored with the `.final` suffix in the source store, which are copied into the destination store by removing this `.final` suffix.

If the copy command is triggered again between the same source and destination object stores, it is currently unable to recognize that the `.final` snapshot has already been uploaded to the destination store by removing the `.final` suffix, and attempts to copy this `.final` snapshot again.

This wasn't a problem until now, since the snapshot is just uploaded again, and the new upload overwrites the old file in the object store. However, since etcd-backup-restore now supports object stores with immutability (WORM) properties, attempting to write to a snapshot that already exists in the object store will cause the copy command to fail.

The command should not fail, and simply not copy any backups. It should exit successfully without any errors.

etcd-backup-restore's `copy` command must handle `.final` snapshots correctly, and not reattempt to upload them to the destination store if they are already present there.

Generate backups in the source bucket, with a snapshot with the `.final` suffix.

Trigger the `copy` command once to copy all the backups, and once more to cause the `copy` command to fail due to immutability.

**Environment**:

- etcd-backup-restore version: v0.37.0 (b6719f939ddf2ec262a8a41099122d25ae81d7d2)
- etcd-druid version (if relevant): v0.31.0

cc @ishan16696 @plkokanov

renormalize:

/area control-plane-migration
/area backup
/kind bug

This PR trims the `.final` suffix from the name of snapshots when their presence is being checked in the destination, as snapshots are not uploaded with this suffix to the destination.

Fixes #906

cc @ishan16696 

**Release note**:

```bugfix operator
The `copy` command does not attempt to upload snapshots with the `.final` suffix in the source to the destination store if they are already in the destination.
```

### DIFF
--- a/pkg/snapshot/copier/copier.go
+++ b/pkg/snapshot/copier/copier.go
@@ -7,6 +7,7 @@ package copier
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -135,7 +136,8 @@ func (c *Copier) copyBackups() error {
 	// find snapshots missing in destination
 	var snapshotsToCopy brtypes.SnapList
 	for _, snapshot := range sourceSnapshot {
-		if _, ok := destSnapshotsMap[snapshot.SnapName]; !ok {
+		snapNameWithoutSuffix := strings.TrimSuffix(snapshot.SnapName, brtypes.FinalSuffix)
+		if _, ok := destSnapshotsMap[snapNameWithoutSuffix]; !ok {
 			snapshotsToCopy = append(snapshotsToCopy, snapshot)
 		} else {
 			c.logger.Infof("Skipping %s snapshot %s as it already exists", snapshot.Kind, snapshot.SnapName)


ishan16696:
thanks @renormalize for the fix.
Can we add a unit tests for this here: https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapshot/copier/copier_test.go

Idea is to check this:
- source should contain 2 snapshots, one with final and one without final
- destination should only contain 1 snapshot, one without final.

ishan16696 (approved):
LGTM!!
As discussed @renormalize will address the review comments later.

PR #917, opened 2025-09-19
Multiple backup storage locations support

### LINKED ISSUE #914: Support multiple backup storage locations

/area backup
/area robustness
/kind enhancement

Mulitple Backend support for backups

**Why is this needed**:
The backup tool only supports configuring one provider for storing backups. To improve resiliency and fault-tolerance, and to enable off-site backups, it'll be great to support configuring more than one provider. This could be limited to 2 providers in a primary/secondary configuration OR a become a mutli-provider setup. 

The backups could be stored in every configured provider, and during restoration, the backed up objects across providers should be aggregated and the latest object can be used for restoration.

tchinmai7:

/area backup
/area disaster-recovery
/kind enhancement

This PR introduces a secondary backup feature.

Fixes #914 

Please read the discussion in the issue

**Release note**:

```feature operator
Operators can now configure a secondary target for backup storage, please check this usage doc for more information: https://github.com/gardener/etcd-backup-restore/tree/master/docs/usage/backup_sync_dual_site.md
```

### DIFF
--- a/docs/usage/backup_sync_dual_site.md
+++ b/docs/usage/backup_sync_dual_site.md
@@ -178,12 +178,10 @@ INFO[0004] Backups copied
 ```
 
 ### Verification Steps
-The examples assume AWS S3 is the backend storage provider. Replace the commands with the appropriate ones for your provider
+The examples assume AWS S3 is the backend storage provider. Backups present in other storage provider's bucket like Azure, GCS, Swift etc can be listed using their respective API calls or via their dashboards
 
 1. **Check Primary Backups**:
-
 ```bash
-aws s3 ls s3://<primary_bucket_name> --recursive
 aws s3 ls s3://<primary_bucket_name> --recursive
 ```
 
--- a/example/00-backup-restore-server-config.yaml
+++ b/example/00-backup-restore-server-config.yaml
@@ -39,7 +39,7 @@ snapstoreConfig:
 #     envPrefix: SECONDARY_
 #     maxParallelChunkUploads: 5
 #     minChunkSize: 5242880
-#     prefix: v2
+#     prefix: "etcd-test"
 #     provider: Local
 #     tempDir: /tmp
 #   backupSyncEnabled: true


ghost:
@tchinmai7 Labels area/todo, kind/todo do not exist.

gardener-robot-ci-2:
Thank you @tchinmai7 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

DUAL_ENDPOINT_README.md
```diff
@@ -0,0 +1,218 @@
...
+```
+
+## Restore Behavior
+
+### Endpoint Selection
+
+1. **Cross-endpoint Search**: The restore process searches for snapshots across both endpoints
+2. **Primary Preference**: When the same snapshot exists on both endpoints, primary is preferred
+3. **Latest Available**: The latest available snapshot across both endpoints is used
+4. **Source Logging**: The restore process logs which endpoint provided the snapshot
```
ishan16696: this is my concern https://github.com/gardener/etcd-backup-restore/issues/914#issuecomment-3317142113

ishan16696:
can you please also add a document for this feature, so that it will be easy for us to follow ... may be here: https://github.com/gardener/etcd-backup-restore/tree/master/docs/usage with any relevant name.

tchinmai7:
> can you please also add a document for this feature, so that it will be easy for us to follow ... may be here: https://github.com/gardener/etcd-backup-restore/tree/master/docs/usage with any relevant name.

added in 9694aec

tchinmai7:
@ishan16696 did you get a chance to review?

ishan16696:
Hi @tchinmai7 ,
I didn't get a time to review it, I will be able to review it next week, sorry for the delay.

.golangci.yaml
```diff
@@ -1,3 +1,4 @@
+version: 2
```
renormalize: Can you please revert this? v1 to v2 is not a simple version line addition, the schema was changed, and we need to run a `migrate` command.

We are also running `v1.64.8` of golangci-lint in etcd-backup-restore, and will migrate to `v2.x.y` in a later PR.

`make check` target fails with these current changes.

docs/usage/enabling_dual_s3_endpoint.md
```diff
@@ -0,0 +1,198 @@
+# Enabling Dual S3 Endpoint Support
+
+The etcd-backup-restore tool supports dual-endpoint configuration, which allows snapshots to be stored simultaneously on two different storage endpoints. This feature provides enhanced data redundancy and fault tolerance by maintaining synchronized backups across multiple storage backends.
+
+## Overview
+
+The dual-endpoint feature works by:
+- Storing all snapshots to a **primary endpoint** (the main storage)
+- Automatically copying snapshots to a **secondary endpoint** (backup storage)
```
renormalize: Current implementation only works with a secondary endpoint for S3. However, I am not comfortable this feature only working with S3 as secondary endpoint.

docs/usage/enabling_dual_s3_endpoint.md
```diff
@@ -0,0 +1,198 @@
...
+- Synchronizing snapshots between endpoints using a background copier service
```
renormalize: I think the background copier service is wasteful. I am of the opinion that we simply upload the snapshot to the secondary provider directly.
I think secondary backups can be useful for human intervention, when an operator instructs etcd-backup-restore to restore from object store 2, and not object store 1.

Since we already agreed in https://github.com/gardener/etcd-backup-restore/issues/914#issuecomment-3339101215 that the initial support would be just for snapshotting to the secondary object store, I think it would be fine to just upload the snapshot directly.

pkg/types/snapstore.go
```diff
@@ -183,22 +184,25 @@ func (s SnapList) Less(i, j int) bool {
...
+	BackupSyncPeriod                 time.Duration `json:"backupSyncPeriod,omitempty"`
+	SecondaryMinChunkSize            int64         `json:"secondaryMinChunkSize,omitempty"`
+	MaxParallelChunkUploads          uint          `json:"maxParallelChunkUploads,omitempty"`
+	SecondaryMaxParallelChunkUploads uint          `json:"secondaryMaxParallelChunkUploads,omitempty"`
+	BackupSyncMaxRetries             int           `json:"backupSyncMaxRetries,omitempty"`
+	BackupSyncConcurrentCopies       int           `json:"backupSyncConcurrentCopies,omitempty"`
+	IsSource                         bool          `json:"isSource,omitempty"`
+	IsEmulatorEnabled                bool          `json:"isEmulatorEnabled,omitempty"`
+	IsSecondary                      bool          `json:"isSecondary,omitempty"`
+	BackupSyncEnabled                bool          `json:"backupSyncEnabled,omitempty"`
```
renormalize: Could existing comments be added back, and could new comments be added for newly introduced fields?

hack/check.sh
```diff
@@ -20,7 +20,7 @@ done
 echo "> Check"
 
 echo "Executing golangci-lint"
-golangci-lint run $GOLANGCI_LINT_CONFIG_FILE --timeout 10m "$@"
+golangci-lint run $GOLANGCI_LINT_CONFIG_FILE --timeout 10m "$@"
```
renormalize: Not sure why a whitespace was added here.

```suggestion
golangci-lint run $GOLANGCI_LINT_CONFIG_FILE --timeout 10m "$@"
```

pkg/snapstore/backup_copier.go
```diff
@@ -0,0 +1,338 @@
...
+	// MaxRetries defines maximum number of retries for failed copy operations
+	MaxRetries int
+	// RetryBackoff defines the backoff duration between retries
+	RetryBackoff time.Duration
+	// ConcurrentCopies defines maximum number of concurrent copy operations
+	ConcurrentCopies int
+}
+
+// DefaultBackupCopierConfig returns default configuration for backup copier
+func DefaultBackupCopierConfig() *BackupCopierConfig {
```
renormalize:
```suggestion
func defaultBackupCopierConfig() *BackupCopierConfig {
```

pkg/snapstore/backup_copier.go
```diff
@@ -0,0 +1,338 @@
...
+		MaxRetries:       3,
+		RetryBackoff:     30 * time.Second,
+		ConcurrentCopies: 2,
+	}
+}
+
+// NewBackupCopier creates a new BackupCopier instance
+func NewBackupCopier(primary, secondary brtypes.SnapStore, config *BackupCopierConfig, logger *logrus.Entry) *BackupCopier {
+	if config == nil {
+		config = DefaultBackupCopierConfig()
```
renormalize: https://github.com/gardener/etcd-backup-restore/pull/917#discussion_r2498228416
```suggestion
		config = defaultBackupCopierConfig()
```

pkg/snapstore/utils.go
```diff
@@ -20,16 +20,109 @@ import (
...
+
+	// Determine what to return based on success/failure of each endpoint
+	if primaryErr != nil && secondaryErr != nil {
+		return nil, fmt.Errorf("failed to create both primary and secondary snapstores - Primary: %v, Secondary: %v", primaryErr, secondaryErr)
+	}
+
+	// Create snapstore with copier based on what we have
+	if primaryStore != nil && secondaryStore != nil {
+		logger.Infof("Created snapstore with backup copier for dual endpoints")
+		copier := NewBackupCopier(primaryStore, secondaryStore, DefaultBackupCopierConfig(), logger)
```
renormalize: `nil` config uses default config.

```suggestion
		copier := NewBackupCopier(primaryStore, secondaryStore, nil, logger)
```

pkg/types/snapstore.go
renormalize: The `Snapstore` API abstracts an object store. As you can see in the fields previously, `Provider`, `Container` (bucket name), `IsSource`, etc all abstract a singular bucket. Adding a `SecondaryProvider` here is definitely not the right thing to do, in my opinion.

Any consumer would just create `Snapstore` for their primary bucket, and another `Snapstore` for their secondary bucket. There is no reason to couple them whatsoever. These abstractions are very simple to begin with, and I am definitely not a fan of breaking away from them after them being in place for years without any issues.

In `BackupRestoreComponentConfig`, you can add an additional field `SecondarySnapstoreConfig`, and define the flags for it, and so on.

Please revert all of this.

pkg/types/snapstore.go
```diff
@@ -218,6 +222,20 @@ func (c *SnapstoreConfig) addFlags(fs *flag.FlagSet, parameterPrefix string) {
...
+	fs.UintVar(&c.SecondaryMaxParallelChunkUploads, parameterPrefix+"secondary-max-parallel-chunk-uploads", c.SecondaryMaxParallelChunkUploads, "secondary maximum number of parallel chunk uploads allowed")
+	fs.Int64Var(&c.SecondaryMinChunkSize, parameterPrefix+"secondary-min-chunk-size", c.SecondaryMinChunkSize, "secondary minimum size for multipart chunk upload")
+
+	// Backup sync configuration flags
+	fs.DurationVar(&c.BackupSyncPeriod, parameterPrefix+"backup-sync-period", 3*time.Minute, "period for checking and copying new backups to secondary endpoint")
+	fs.IntVar(&c.BackupSyncMaxRetries, parameterPrefix+"backup-sync-max-retries", 5, "maximum number of retry attempts for failed backup copies")
+	fs.DurationVar(&c.BackupSyncRetryBackoff, parameterPrefix+"backup-sync-retry-backoff", 10*time.Second, "backoff duration between backup copy retry attempts")
+	fs.IntVar(&c.BackupSyncConcurrentCopies, parameterPrefix+"backup-sync-concurrent-copies", 10, "maximum number of concurrent backup copy operations")
+	fs.BoolVar(&c.BackupSyncEnabled, parameterPrefix+"backup-sync-enabled", false, "enable background backup synchronization to secondary endpoint")
```
renormalize: https://github.com/gardener/etcd-backup-restore/pull/917#discussion_r2498437276

pkg/types/snapstore.go
```diff
@@ -254,3 +272,68 @@ func (c *SnapstoreConfig) MergeWith(other *SnapstoreConfig) {
...
+
+// DualOperationResult represents the results of an operation on both endpoints.
+type DualOperationResult struct {
+	// Primary holds the result for the primary endpoint.
+	Primary OperationResult `json:"primary"`
+	// Secondary holds the result for the secondary endpoint.
+	Secondary OperationResult `json:"secondary"`
+	// OverallSuccess indicates if at least one endpoint operation succeeded.
+	OverallSuccess bool `json:"overallSuccess"`
+}
```
renormalize: Neither type is used anywhere. Please remove them, or adapt code to use them appropriately.

pkg/types/snapstore.go
```diff
@@ -254,3 +272,68 @@ func (c *SnapstoreConfig) MergeWith(other *SnapstoreConfig) {
...
+	}
+	if secondaryConfig.MaxParallelChunkUploads == 0 {
+		secondaryConfig.MaxParallelChunkUploads = c.MaxParallelChunkUploads
+	}
+	if secondaryConfig.MinChunkSize == 0 {
+		secondaryConfig.MinChunkSize = c.MinChunkSize
+	}
+
+	return secondaryConfig
+}
```
renormalize: This would not be needed since the second snapstore config is parsed directly from the flags defined for it.

pkg/snapstore/backup_copier.go
```diff
@@ -0,0 +1,338 @@
...
+type BackupCopierConfig struct {
+	// SyncPeriod defines how often to check for new backups to copy
+	SyncPeriod time.Duration
+	// MaxRetries defines maximum number of retries for failed copy operations
+	MaxRetries int
+	// RetryBackoff defines the backoff duration between retries
+	RetryBackoff time.Duration
+	// ConcurrentCopies defines maximum number of concurrent copy operations
+	ConcurrentCopies int
+}
```
renormalize: We already have code that copies snapshots from one bucket to another in the copier. Can you explain the benefits you see in implementing this again?

cmd/snapshot.go
```diff
@@ -56,10 +63,38 @@ storing snapshots on various cloud storage providers as well as local disk locat
...
+
 			go ssr.RunGarbageCollector(ctx.Done())
 			if err := ssr.Run(ctx.Done(), true); err != nil {
 				logger.Fatalf("Snapshotter failed with error: %v", err)
 			}
+
+			if ss.Copier != nil && ss.Copier.IsRunning() {
+				logger.Info("Stopping backup copier for dual endpoint...")
+				ss.Copier.Stop()
+			}
```
renormalize: I'm not convinced yet that the snapshotter must also support the dual-endpoint backups. Do you feel strongly about this?

renormalize (changes requested):
Thanks for the PR @tchinmai7, and apologies from the slow review.

Here are my opinions about the feature being proposed:

- I would like this feature to be extensible to any provider and not just S3. Although I understand that your requirements are for S3 compatible providers, etcd-backup-restore as a project aims to be provider agnostic to the furthest extent.

  Obviously, features like SSE-C which were contributed for S3 compatible providers were accepted since it is essentially just configuration passed to the specific provider. It does not inherently change the functionality of etcd-backup-restore.

  However, the changes being proposed in this PR are not just simple configuration, but a fundamental difference in the capability of etcd-backup-restore. I strongly feel that if such a feature were to come to etcd-backup-restore, it should come for all providers, since there is nothing inherently provider specific about uploading snapshots to two separate buckets in different infrastructure providers.
  
  I would really appreciate it, if the scope of the proposed changes is changed from S3 to all providers. In reality, this just adds ABS, GCS, and OpenStack Swift. I am more than happy to assist you in reviews for these providers, as well as help in testing the feature for these providers. ABS and GCS are quite popular and also have emulators used widely across the community for development (also supported by etcd-backup-restore); OpenStack Swift on the other hand is not as popular, and I don't mind helping you out there with development and testing.

- I do not think the code changes for this PR are on the right path. I don't think the `Snapstore` abstraction is to be modified to achieve the desired functionality. Each `Snapstore` abstracts one remote object store, and that is it. We should simply create a additional `Snapstore` for the secondary remote object store.

  The `server` should accept two `Snapstore` configurations. If a second is present, we can enable the copier service. There need not be any further configuration. Since restoration is not planned to be supported right away from two different buckets without making the application quite a bit more complicated, I think it is good enough default behavior to begin copy of one bucket to another when a secondary is specified in the configuration.

- I do not think we should implement all of the copier yet again for this feature. We already have code that performs a copy from one bucket to another. `Copier.copyBackups()`. This function can be exported and be reused fully.

- I am not fully convinced about a complete copy of snapshots from bucket 1 to 2. I understand that this was the original decision due to complications with respect to consistency between both the buckets, but if the restoration from the secondary bucket is going to be a special scenario and not automatic, do we need to bother with copying, and instead simply upload to the second bucket directly?

  It seems extremely wasteful to upload a snapshot and be charged for its ingress, download it right after and be charged for its egress, and then upload the newly downloaded snapshot to another bucket and then be charged for its ingress. Egress is not cheap, and the costs will add up when you're running a large number of clusters. I want to continue discussion on this.

  If automatic restoration from a second bucket is a potential feature in the future, then we can go with the current solution but I need some time to discuss it with the other maintainers.

docs/usage/enabling_dual_s3_endpoint.md
tchinmai7: do you mean this should support other providers as well?

docs/usage/enabling_dual_s3_endpoint.md
tchinmai7: I was going by the background copier idea mentioned in https://github.com/gardener/etcd-backup-restore/issues/914#issuecomment-3337820447. Open to directly writing to the second provider if you'd prefer that

pkg/types/snapstore.go
tchinmai7: okay, that makes sense, will change!

pkg/types/snapstore.go
tchinmai7: This is the remnant of when i was doing both writes in parallel. If we switch to writing to both endpoints, this will be used.

tchinmai7:
@renormalize Thanks for your review
> I would like this feature to be extensible to any provider and not just S3. Although I understand that your requirements are for S3 compatible providers, etcd-backup-restore as a project aims to be provider agnostic to the furthest extent.

I agree - happy to contribute a similar change to other providers as well. starting off with s3 since that's the one i'm most familiar with :) 

> We already have code that performs a copy from one bucket to another. I would like to avoid implementing the same thing twice as the copy flow was implemented separately in this PR. Copier.copyBackups(). This function can be exported and be reused fully.

this is a good point - i can switch to using the existing `copyBackups` functions instead of rewriting that. its an oversight on my part


>I am not fully convinced about a complete copy of snapshots from bucket 1 to 2. I understand that this was the original decision due to complications with respect to consistency between both the buckets, but if the restoration from the secondary bucket is going to be a special scenario and not automatic, do we need to bother with copying, and instead simply upload to the second bucket directly?

This was the original approach that i outlined in an earlier version of this PR, and switched to a background copy based on https://github.com/gardener/etcd-backup-restore/issues/914#issuecomment-3337820447 - @ishan16696 thoughts about moving to an inline copy again?

renormalize:
@tchinmai7 thanks for the quick reply! Also, thank you for being up for contributing to the other providers as well!

> this is a good point - i can switch to using the existing copyBackups functions instead of rewriting that. its an oversight on my part

That's alright! Happy to help!

> This was the original approach that i outlined in an earlier version of this PR, and switched to a background copy based on https://github.com/gardener/etcd-backup-restore/issues/914#issuecomment-3337820447 - @ishan16696 thoughts about moving to an inline copy again?

I think the point of contention has been restoration. If you prefer restoration to be done automatically, then the background copy is the way to go.

However, if the restoration is going to be a manual signal provided to etcd-backup-restore, then I do not see an issue with the double upload, since it is a fall-back.

docs/usage/enabling_dual_s3_endpoint.md
renormalize: Yes.

docs/usage/enabling_dual_s3_endpoint.md
renormalize: As you mentioned in https://github.com/gardener/etcd-backup-restore/pull/917#issuecomment-3498262609, let's wait for some input from Ishan.

ishan16696:
I think we have now 2 approaches:
i) Either backup-restore should copy the snapshots (using copy sub-command) from primary bucket to secondary bucket. (achieving eventual consistency b/w 2 buckets).
ii) Or while taking the snapshots backup-restore should just simply upload the snapshot to both of bucket (primary as well as secondary bucket).

Few things I would like to mention here regarding both of above mentioned approaches :
1. The only problem I see with second approach is how to make sure there is a consistency b/w 2 buckets ?
Let me explain this:
Assuming we go with 2nd approach, and now snapshotter uploads the snapshot to primary bucket and also checks if secondary bucket is configured or not if configured then also uploads the snapshot to secondary buckets.
So, In good case scenario everything works fine but there can be many bad scenario possible, for eg:


|    | Primary bucket                                 | Secondary bucket |
| ------ | ------------------------------------------- | ------- |
| Case 1 | Available but upload is slow :snail:  |     Available :white_check_mark:  |
| case 2 | Available :white_check_mark:  | Available but upload is slow :snail:      |
| case 3 | Not available :x:  | Available  :white_check_mark:    |
|case 4 | Available :white_check_mark:  | Not available  :x:    |

- Now, consider a scenario when case 2 and case 4 happened and then snashotter uploads the snapshot to primary bucket then backup-restore crashed (due to any reason) then how do you resume/reupload the snapshot to secondary bucket ?
- Another problem I see is assume case 2 happens but here's the problem is currently snapshotter takes the snapshot in seqential way(one after another), if upload to secondary bucket is slow, will the snapshotter keep moving ahead for its upload to primary bucket and take another snapshot or keep waiting until upload finishes for secondary bucket ? or if upload fails to secondary bucket then who will gonna retry to take snapshot again ?
https://github.com/gardener/etcd-backup-restore/blob/0949c9598053e24e187b6600af0591d4cec345a6/pkg/snapshot/snapshotter/snapshotter.go#L629

Likewise there can be many scenarios possible, I'm not saying these things can't be handled, they can be like we handled them today (only for primary bucket upload) but it will make the snapshotter more complicated with so many conditions to handle with more edge cases.

2. As @renormalize has rightly pointed to out that if we go with first approach then cost of ingress will increase due to download of snapshot.
I have couple of counter argument to this:
-  User wants to have a secondary backups, they are configuring it. It means they ready to pay to have higher fault tolerance then I don't think cost should be an issue.
- Another point is [compaction job](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/02-snapshot-compaction.md) which we run today is also downloading the snapshots to take a bigger full snapshot to minimize the time of restoration, and it is also increasing the cost and we are fine it as it's reduces the restoration time, I think cost won't be that high as object storage are already very cost effective.

renormalize:
@ishan16696 thank you for the specifying your opinions in great detail! I am convinced that we should choose the `copy` design. The snapshotter will end up becoming significantly more complicated in the process of making it robust against this, which we should obviously avoid.

@tchinmai7 let us proceed with the `copy` approach as suggested by @ishan16696. Feel free to address the previously posted comments, and re-request a review. 

One more thing - multi-provider support should be feature gated until all providers have been implemented and tested. There isn't any code which allows `featureGate`s in etcd-backup-restore. I will introduce that before this PR is merged. Thanks!

pkg/snapstore/backup_copier.go
tchinmai7: Will move to the copier pacakge, there's a import cycle here (copier imports snapstore) so will have to introduce an interface of sorts to decouple the two

cmd/snapshot.go
tchinmai7: that's fair. This could be isolated just to the server mode

tchinmai7:
@renormalize @ishan16696 I have pushed up a new set of changes that should drastically simplify this. I still need to write tests, but i'd appreciate some feedback on this approach. This uses existing primitives from the copier, and garbage collector. I also have to figure out how to emit metrics separately for the backup endpoint from the garbage collector, but i'll do that once y'all tell me if this makes sense

pkg/server/backuprestoreserver.go
```diff
@@ -234,6 +235,24 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 					b.logger.Fatalf("failed to create snapstore from configured storage provider: %v", err)
 				}
 
+				if b.config.BackupSyncEnabled && b.config.SecondarySnapstoreConfig != nil {
+					b.logger.Infof("Starting backup copier..")
+					secondary, err := snapstore.GetSnapstore(b.config.SecondarySnapstoreConfig)
+					if err != nil {
+						b.logger.Fatalf("failed to create secondary snapstore from configured storage provider: %v", err)
+					}
+					backupssr, err := snapshotter.NewSnapshotter(b.logger, b.config.SnapshotterConfig, ss, b.config.EtcdConnectionConfig, b.config.CompressionConfig, b.config.HealthConfig, b.config.SecondarySnapstoreConfig)
```
tchinmai7: the ssr isn't used other than to start the garbagecollector

pkg/snapshot/copier/copier.go
```diff
@@ -247,3 +250,57 @@ func (c *Copier) doWaitForFinalSnapshot(ctx context.Context, interval time.Durat
...
+	defer func() {
+		c.mu.Lock()
+		c.running = false
+		c.mu.Unlock()
+	}()
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	// start an initial sync first
+	if err := c.CopyBackups(ctx); err != nil {
+		c.logger.Errorf("could not copy backups: %v", err)
```
renormalize: Nit: This might help disambiguate where the `CopyBackups()` failed: the first time, or during the steady state `for` loop.
Not really a strong opinion.

```suggestion
		c.logger.Errorf("could not peform the initial copy of backups: %v", err)
```

pkg/server/backuprestoreserver.go
```diff
@@ -234,6 +235,24 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+					if err != nil {
+						b.logger.Fatalf("failed to create snapshot backup copier: %v", err)
+					}
+
+					cp := copier.NewCopier(b.logger, ss, secondary, -1, -1, 10, false, 0)
+					if err := cp.SyncBackups(ctx, 30*time.Second); err != nil {
```
renormalize: 30 seconds seems a bit too frequent as a hard-coded value, perhaps this could be increased?
I'm also open to making this user configuration, so users can have syncs once an hour for example, or at a 30 second time period. Do you foresee a need to configure this in your use-case in the future?

Assuming this was hard-coded since the PR is marked draft currently.

renormalize (commented):
Thanks for making all the requested changes @tchinmai7! The PR looks much better now!

The flags don't seem to be working as expected, the binary panics like so:

```
[I] ~/go/src/github.com/gardener/etcd-backup-restore (pr/tchinmai7/917)
❯ ./bin/etcdbrctl server --help
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x105597f08]

goroutine 1 [running]:
github.com/gardener/etcd-backup-restore/pkg/types.(*SnapstoreConfig).addFlags(0x0, 0x1400015de00, {0x106e7dbea, 0xa})
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/pkg/types/snapstore.go:220 +0x28
github.com/gardener/etcd-backup-restore/pkg/types.(*SnapstoreConfig).AddSecondaryFlags(...)
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/pkg/types/snapstore.go:216
github.com/gardener/etcd-backup-restore/pkg/server.(*BackupRestoreComponentConfig).AddFlags(0x1400041d420, 0x1400015de00)
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/pkg/server/init.go:50 +0xe4
github.com/gardener/etcd-backup-restore/cmd.(*serverOptions).addFlags(0x140002e23f0, 0x1400015de00)
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/cmd/options.go:55 +0xac
github.com/gardener/etcd-backup-restore/cmd.NewServerCommand({0x107bb5a70, 0x14000755810})
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/cmd/server.go:49 +0xf4
github.com/gardener/etcd-backup-restore/cmd.NewBackupRestoreCommand({0x107bb5a70, 0x14000755810})
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/cmd/root.go:33 +0x100
main.main()
        /Users/renormalize/go/src/github.com/gardener/etcd-backup-restore/main.go:28 +0x4c
```

You can mark it open for review and move it out of draft once you fix the above bug and get the flags working as expected. I've also added a couple nits - the implementation with respect to the copier looks good.

You can also rename the PR to reflect what it does now. Before we merge, I'll bring in feature flag changes into master, so you can rebase your PR and incorporate this as a flagged feature. That should be trivial.

Thanks, and sorry for the slow review!

pkg/server/backuprestoreserver.go
tchinmai7: yes, this is hard-coded for a quick change. I'll make this configurable, and pick a more sensible default (1 hr sounds good to me)

pkg/server/types.go
```diff
@@ -22,12 +25,15 @@ type BackupRestoreComponentConfig struct {
...
 	SnapstoreConfig          *brtypes.SnapstoreConfig          `json:"snapstoreConfig,omitempty"`
+	SecondarySnapstoreConfig *brtypes.SnapstoreConfig          `json:"secondarySnapstoreConfig,omitempty"`
 	CompressionConfig        *compressor.CompressionConfig     `json:"compressionConfig,omitempty"`
 	RestorationConfig        *brtypes.RestorationConfig        `json:"restorationConfig,omitempty"`
 	HealthConfig             *brtypes.HealthConfig             `json:"healthConfig,omitempty"`
 	LeaderElectionConfig     *brtypes.Config                   `json:"leaderElectionConfig,omitempty"`
 	ExponentialBackoffConfig *brtypes.ExponentialBackoffConfig `json:"exponentialBackoffConfig,omitempty"`
 	DefragmentationSchedule  string                            `json:"defragmentationSchedule"`
+	BackupSyncEnabled        bool                              `json:"backupSyncEnabled,omitempty"`
+	BackupSyncTimeout        time.Duration                     `json:"backupSyncTimeout,omitempty"`
```
renormalize: Not sure if `BackupSyncTimeout` is semantically correct. How about `BackupSyncPeriod`? Thoughts?

```suggestion
	BackupSyncPeriod        time.Duration                     `json:"backupSyncPeriod,omitempty"`
```

pkg/server/types.go
```diff
@@ -5,13 +5,16 @@
...
+	"time"
+
 	"github.com/gardener/etcd-backup-restore/pkg/compressor"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
 )
 
 const (
 	defaultServerPort              = 8080
 	defaultDefragmentationSchedule = "0 0 */3 * *"
+	defaultBackupSyncTimeout       = 1 * time.Hour
```
renormalize:
```suggestion
	defaultBackupSyncPeriod       = 1 * time.Hour
```

ishan16696:
@tchinmai7 can we now move the PR from draft to read for review as it's being reviewed 😅 ?

pkg/server/types.go
ishan16696: what about `SecondayBackupSyncPeriod` ? as it's for secondary backup and `BackupSyncPeriod` can be ambiguous as people might think it's for primary backup. wdyt ?

pkg/server/types.go
ishan16696: same as above

ishan16696:
Why you have removed a document. IMO a simple document it's still very useful from user's consumption perspective which explains this feature and it's usage. we can also mention that doc link in release notes. wdyt @tchinmai7 ?

pkg/server/init.go
```diff
@@ -45,10 +48,12 @@ func (c *BackupRestoreComponentConfig) AddFlags(fs *flag.FlagSet) {
 	c.HealthConfig.AddFlags(fs)
 	c.LeaderElectionConfig.AddFlags(fs)
 	c.ExponentialBackoffConfig.AddFlags(fs)
-
+	c.SecondarySnapstoreConfig.AddSecondaryFlags(fs)
 	// Miscellaneous
 	fs.StringVar(&c.DefragmentationSchedule, "defragmentation-schedule", c.DefragmentationSchedule, "schedule to defragment etcd data directory")
+	fs.DurationVar(&c.BackupSyncTimeout, "backup-sync-timeout", c.BackupSyncTimeout, "timeout for periodic backup sync operations (defaults to 1h)")
```
ishan16696:
```suggestion
	fs.DurationVar(&c.BackupSyncTimeout, "backup-sync-timeout", c.BackupSyncTimeout, "timeout for periodic backup sync operations.")
```
It is already mentioned by cobra lib.
Right now, it's like this:
```
--backup-sync-timeout duration         timeout for periodic backup sync operations (defaults to 1h) (default 1h0m0s)
```

pkg/server/backuprestoreserver.go
```diff
@@ -234,6 +235,24 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 					b.logger.Fatalf("failed to create snapstore from configured storage provider: %v", err)
 				}
 
+				if b.config.BackupSyncEnabled {
+					b.logger.Infof("Starting backup copier..")
```
ishan16696:
```suggestion
					b.logger.Infof("Starting periodic backup copier..")
```

pkg/server/init.go
```diff
@@ -45,10 +48,12 @@ func (c *BackupRestoreComponentConfig) AddFlags(fs *flag.FlagSet) {
...
 	fs.BoolVar(&c.UseEtcdWrapper, "use-etcd-wrapper", c.UseEtcdWrapper, "to enable backup-restore to use etcd-wrapper related functionality. Note: enable this flag only if etcd-wrapper is deployed.")
+	fs.BoolVar(&c.BackupSyncEnabled, "backup-sync-enabled", c.BackupSyncEnabled, "enable backup-sync feature")
```
ishan16696:
```suggestion
	fs.BoolVar(&c.BackupSyncEnabled, "backup-sync-enabled", c.BackupSyncEnabled, "enable secondary backup-sync feature")
```

pkg/server/init.go
```diff
@@ -17,20 +17,23 @@ import (
...
+		SecondarySnapstoreConfig: snapstore.NewSnapstoreConfig(),
 		CompressionConfig:        compressor.NewCompressorConfig(),
 		RestorationConfig:        brtypes.NewRestorationConfig(),
 		DefragmentationSchedule:  defaultDefragmentationSchedule,
+		BackupSyncTimeout:        defaultBackupSyncTimeout,
 		HealthConfig:             brtypes.NewHealthConfig(),
 		LeaderElectionConfig:     brtypes.NewLeaderElectionConfig(),
 		ExponentialBackoffConfig: brtypes.NewExponentialBackOffConfig(),
 		UseEtcdWrapper:           usageOfEtcdWrapperEnabled,
+		BackupSyncEnabled:        false,
```
ishan16696: IMO, all secondary backup config should be under 1 config.
```go
// SecondarySnapstoreConfig defines the configuration for secondary snapshot store.
type SecondarySnapstoreConfig struct {
	StoreConfig       *SnapstoreConfig
	BackupSyncEnabled bool          `json:"backupSyncEnabled,omitempty"`
	BackupSyncTimeout time.Duration `json:"backupSyncTimeout,omitempty"`
}
```

Flags can be added like this:
```go
// AddFlags adds the flags to flagset.
func (c *SecondarySnapstoreConfig) AddFlags(fs *flag.FlagSet) {
	c.StoreConfig.addFlags(fs, "secondary-")
	fs.BoolVar(&c.BackupSyncEnabled, "backup-sync-enabled", c.BackupSyncEnabled, "enable backup-sync feature")
	fs.DurationVar(&c.BackupSyncTimeout, "backup-sync-timeout", c.BackupSyncTimeout, "timeout for periodic backup sync operations (defaults to 1h)")
}
```

tchinmai7:
@ishan16696 @renormalize I've addressed comments, and pushed up changes. Thank you so much for taking the time to review this.

Tested this on a kind cluster, verified copying works
```
 time="2025-12-04T00:07:43Z" level=info msg="Copying backups ..." actor=copier                                                                                                                                                               │
│ time="2025-12-04T00:07:43Z" level=info msg="Getting source backups..." actor=copier                                                                                                                                                         │
│ time="2025-12-04T00:07:43Z" level=info msg="Bucket versioning is not found to be enabled."                                                                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="GC: Executing garbage collection..." actor=snapshotter                                                                                                                                          │
│ time="2025-12-04T00:07:43Z" level=info msg="Getting destination snapshots..." actor=copier                                                                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Bucket versioning is not found to be enabled."                                                                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Bucket versioning is not found to be enabled."                                                                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="GC: Total number garbage collected chunks: 0" actor=snapshotter                                                                                                                                 │
│ time="2025-12-04T00:07:43Z" level=info msg="GC: Deleting old delta snapshot: Incr-03757270-03758699-1764806266.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Full snapshot Full-00000000-03759344-1764806400.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Full snapshot Full-00000000-03742237-1764802817.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03757270-03758699-1764806266.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03755833-03757269-1764805966.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03754405-03755832-1764805665.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03752983-03754404-1764805365.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03751557-03752982-1764805064.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03750129-03751556-1764804764.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03748793-03750128-1764804463.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03748650-03748792-1764804187.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03747216-03748649-1764804164.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03746527-03747215-1764803864.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03745111-03746526-1764803721.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03743683-03745110-1764803421.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Skipping Incr snapshot Incr-03742238-03743682-1764803120.gz as it already exists" actor=copier                                                                                                  │
│ time="2025-12-04T00:07:43Z" level=info msg="Copying Incr snapshot Incr-03759345-03760141-1764806567.gz..." actor=copier
```
and garbage collection too
```
│ time="2025-12-04T00:07:44Z" level=info msg="Backups copied" actor=copier                                                                                                                                                                    │
│ time="2025-12-04T00:07:44Z" level=info msg="GC: Executing garbage collection..." actor=snapshotter                                                                                                                                          │
│ time="2025-12-04T00:07:45Z" level=info msg="Bucket versioning is not found to be enabled."                                                                                                                                                  │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Total number garbage collected chunks: 0" actor=snapshotter                                                                                                                                 │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03757270-03758699-1764806266.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03755833-03757269-1764805966.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03754405-03755832-1764805665.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03752983-03754404-1764805365.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03751557-03752982-1764805064.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03750129-03751556-1764804764.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:45Z" level=info msg="GC: Deleting old delta snapshot: Incr-03748793-03750128-1764804463.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Deleting old delta snapshot: Incr-03748650-03748792-1764804187.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Deleting old delta snapshot: Incr-03747216-03748649-1764804164.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Deleting old delta snapshot: Incr-03746527-03747215-1764803864.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Deleting old delta snapshot: Incr-03745111-03746526-1764803721.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Deleting old delta snapshot: Incr-03743683-03745110-1764803421.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Deleting old delta snapshot: Incr-03742238-03743682-1764803120.gz" actor=snapshotter                                                                                                        │
│ time="2025-12-04T00:07:46Z" level=info msg="GC: Total number garbage collected snapshots: 13" actor=snapshotter
```

I had to make some adjustments to other areas to make the secondary configuration work properly. Lmk what you think

docs/usage/backup_sync_dual_site.md
```diff
@@ -0,0 +1,306 @@
+# Backup Sync with Dual Site
```
ishan16696, line 1: IMO, Let's move the document to this location: https://github.com/gardener/etcd-backup-restore/tree/master/docs/usage as I'm looking for a doc only from consumption perspective.

docs/operations/backup_sync_dual_site.md
```diff
@@ -0,0 +1,306 @@
...
+
+```yaml
+env:
+- name: AWS_APPLICATION_CREDENTIALS_JSON
+  value: /home/.aws/credentials.json
+- name: SECONDARY_AZURE_APPLICATION_CREDENTIALS_JSON
+  value: /home/.azure/secondary-credentials.json
+```
+
+## How It Works
```
ishan16696: Let's not include the implementation details and I just want to have a doc from users consumption perspective. Please remove this section of "How it works". I think you have given a good idea of how it works in above section, that's should be sufficient.

docs/operations/backup_sync_dual_site.md
```diff
@@ -0,0 +1,306 @@
...
+4. **Manual Sync**: Use the `copy` command for one-time catch-up:
+   ```bash
+   etcdbrctl copy \
+     --storage-provider=S3 \
+     --store-container=primary-bucket \
+     --secondary-storage-provider=ABS \
+     --secondary-store-container=secondary-bucket
+   ```
+
+## Leader Election Considerations
```
ishan16696: please remove this section as well. You can add one line or section regarding garbage collection of secondary backups.

pkg/snapstore/init.go
```diff
@@ -16,3 +18,11 @@ func NewSnapstoreConfig() *brtypes.SnapstoreConfig {
 		TempDir:                 "/tmp",
 	}
 }
+
+func NewSecondarySnapstoreConfig() *brtypes.SecondarySnapstoreConfig {
+	return &brtypes.SecondarySnapstoreConfig{
+		StoreConfig:       NewSnapstoreConfig(),
+		SyncPeriod:        1 * time.Hour,
```
ishan16696:
```suggestion
		SyncPeriod:        brtypes.DefaultSecondayBackupSyncPeriod,
```
you can declare the const here: https://github.com/gardener/etcd-backup-restore/blob/81c4c17376e60731d94469fee5438e3071798036/pkg/types/snapstore.go#L18

pkg/server/types.go
```diff
@@ -5,13 +5,16 @@
...
 	"github.com/gardener/etcd-backup-restore/pkg/compressor"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
 )
 
 const (
-	defaultServerPort              = 8080
-	defaultDefragmentationSchedule = "0 0 */3 * *"
+	defaultServerPort               = 8080
+	defaultDefragmentationSchedule  = "0 0 */3 * *"
+	defaultSecondayBackupSyncPeriod = 1 * time.Hour
```
ishan16696: right now it's not used anywhere, you can remove it once you declare here: 
https://github.com/gardener/etcd-backup-restore/blob/81c4c17376e60731d94469fee5438e3071798036/pkg/types/snapstore.go#L18

ishan16696 (changes requested):
thanks @tchinmai7 for adding usage doc, I have added few comments. Overall PR looks good.

pkg/types/snapstore.go
```diff
@@ -199,6 +202,10 @@ type SnapstoreConfig struct {
 	IsSource bool `json:"isSource,omitempty"`
 	// IsEmulatorEnabled indicates whether a storage emulator is being used for the snapstore.
 	IsEmulatorEnabled bool `json:"isEmulatorEnabled,omitempty"`
+
```
renormalize: Nit: Since the fields of this struct do not have any newlines between them.
```suggestion
```

pkg/types/snapstore.go
```diff
@@ -254,3 +261,27 @@ func (c *SnapstoreConfig) MergeWith(other *SnapstoreConfig) {
...
+type SecondarySnapstoreConfig struct {
+	StoreConfig       *SnapstoreConfig
+	BackupSyncEnabled bool          `json:"backupSyncEnabled,omitempty"`
+	SyncPeriod        time.Duration `json:"syncPeriod,omitempty"`
+}
+
+func (c *SecondarySnapstoreConfig) AddFlags(fs *flag.FlagSet) {
+	c.StoreConfig.addFlags(fs, "secondary-")
+	fs.BoolVar(&c.BackupSyncEnabled, "secondary-backup-sync-enabled", c.BackupSyncEnabled, "enable secondary backup-sync feature")
```
renormalize, line 273: As mentioned above, can we mark this alpha?

```suggestion
	fs.BoolVar(&c.BackupSyncEnabled, "secondary-backup-sync-enabled", c.BackupSyncEnabled, "enable secondary backup-sync feature [alpha]")
```

renormalize (approved):
Thanks!

Regarding feature-flagging: I've taken a look at how we've implemented feature flags in etcd-druid, and mirroring that here will be an unnecessary amount of work at this moment in my opinion. 

I think marking it as experimental/alpha in the release notes and in the flag description would be sufficient in my opinion.

/lgtm

docs/usage/backup_sync_dual_site.md
```diff
@@ -0,0 +1,238 @@
...
+INFO[0002] Getting destination snapshots...
+INFO[0003] Copying Full snapshot v2/Backup-1234567890/Full-00000000-00000001-1234567890...
+INFO[0004] Backups copied
+```
+
+### Verification Steps
+
+1. **Check Primary Backups**:
+   ```bash
+   etcdbrctl snapshot list --storage-provider=S3 --store-container=primary-bucket
```
ishan16696: I think this verification step should be storage provider api call not `etcdbrctl snapshot` subcommand.
example for s3:
```
aws s3 ls s3://<bucket_name> --recursive
```

docs/usage/backup_sync_dual_site.md
```diff
@@ -0,0 +1,238 @@
...
+   ```
+
+2. **Check Secondary Backups**:
+   ```bash
+   etcdbrctl snapshot list --storage-provider=ABS --store-container=secondary-bucket
```
ishan16696: same as above

ishan16696:
can you also update the example config file [here](https://github.com/gardener/etcd-backup-restore/blob/master/example/00-backup-restore-server-config.yaml) with something like this: 

```yaml
# secondarySnapstoreConfig:
#   StoreConfig:
#     container: secondary-bucket
#     prefix: "etcd-test"
#     provider: Local
#     tempDir: /tmp
#   backupSyncEnabled: true
#   syncPeriod: 5m
```

ishan16696:
I started to test this with this config with sync period of `1m` but what I saw is sync period is `syncPeriod: 3600000000000`, I think it's taking default value in milliseconds. can you also please take a look at this ?
```
./bin/etcdbrctl server \
     --storage-provider=Local \
     --store-container=primary-bucket \
     --secondary-storage-provider=Local \
     --secondary-store-container=secondary-bucket \
     --secondary-backup-sync-enabled=true \ --secondary-backup-sync-period=1m
```

pkg/types/snapstore.go
```diff
@@ -254,3 +260,27 @@ func (c *SnapstoreConfig) MergeWith(other *SnapstoreConfig) {
 		c.TempDir = other.TempDir
 	}
 }
+
+type SecondarySnapstoreConfig struct {
+	StoreConfig       *SnapstoreConfig
+	BackupSyncEnabled bool          `json:"backupSyncEnabled,omitempty"`
+	SyncPeriod        time.Duration `json:"syncPeriod,omitempty"`
```
ishan16696:
```suggestion
	SyncPeriod        wrappers.Duration `json:"syncPeriod,omitempty"`
```

pkg/snapstore/init.go
```diff
@@ -16,3 +16,11 @@ func NewSnapstoreConfig() *brtypes.SnapstoreConfig {
 		TempDir:                 "/tmp",
 	}
 }
+
+func NewSecondarySnapstoreConfig() *brtypes.SecondarySnapstoreConfig {
+	return &brtypes.SecondarySnapstoreConfig{
+		StoreConfig:       NewSnapstoreConfig(),
+		SyncPeriod:        brtypes.DefaultSecondaryBackupSyncPeriod,
```
ishan16696:
```suggestion
		SyncPeriod:        wrappers.Duration{Duration: brtypes.DefaultSecondaryBackupSyncPeriod},
```

pkg/types/snapstore.go
```diff
@@ -254,3 +260,27 @@ func (c *SnapstoreConfig) MergeWith(other *SnapstoreConfig) {
...
+}
+
+func (c *SecondarySnapstoreConfig) AddFlags(fs *flag.FlagSet) {
+	c.StoreConfig.addFlags(fs, "secondary-")
+	fs.BoolVar(&c.BackupSyncEnabled, "secondary-backup-sync-enabled", c.BackupSyncEnabled, "enable secondary backup-sync feature [alpha]")
+	fs.DurationVar(&c.SyncPeriod, "secondary-backup-sync-period", c.SyncPeriod, "period for periodic backup sync operations")
```
ishan16696:
```suggestion
	fs.DurationVar(&c.SyncPeriod.Duration, "secondary-backup-sync-period", c.SyncPeriod.Duration, "period for periodic backup sync operations")
```

pkg/server/backuprestoreserver.go
```diff
@@ -234,6 +235,24 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
...
+					if err != nil {
+						b.logger.Fatalf("failed to create secondary snapstore from configured storage provider: %v", err)
+					}
+					backupssr, err := snapshotter.NewSnapshotter(b.logger, b.config.SnapshotterConfig, ss, b.config.EtcdConnectionConfig, b.config.CompressionConfig, b.config.HealthConfig, b.config.SecondarySnapstoreConfig.StoreConfig)
+					if err != nil {
+						b.logger.Fatalf("failed to create snapshot backup copier: %v", err)
+					}
+
+					cp := copier.NewCopier(b.logger, ss, secondary, -1, -1, 10, false, 0)
+					if err := cp.SyncBackups(ctx, b.config.SecondarySnapstoreConfig.SyncPeriod); err != nil {
```
ishan16696:
```suggestion
					if err := cp.SyncBackups(ctx, b.config.SecondarySnapstoreConfig.SyncPeriod.Duration); err != nil {
```

ishan16696 (changes requested):
I tested this PR, things are working fine, it just it requires few changes for this https://github.com/gardener/etcd-backup-restore/pull/917#issuecomment-3659483235
```
INFO[0245] Copying backups ...                           actor=copier
INFO[0245] Getting source backups...                     actor=copier
INFO[0245] Getting destination snapshots...              actor=copier
INFO[0245] Skipping Full snapshot Full-00000000-00000001-1765875003.gz as it already exists  actor=copier
INFO[0245] Skipping Incr snapshot Incr-00000008-00000008-1765876885.gz as it already exists  actor=copier
INFO[0245] Skipping Incr snapshot Incr-00000006-00000007-1765876845.gz as it already exists  actor=copier
INFO[0245] Skipping Incr snapshot Incr-00000004-00000005-1765876805.gz as it already exists  actor=copier
INFO[0245] Skipping Incr snapshot Incr-00000003-00000003-1765876785.gz as it already exists  actor=copier
INFO[0245] Skipping Incr snapshot Incr-00000002-00000002-1765875023.gz as it already exists  actor=copier
INFO[0245] Backups copied
```

docs/usage/backup_sync_dual_site.md
tchinmai7: I was going for a provider agnostic method - do you think providing a sample for just aws is enough?

tchinmai7:
@ishan16696 thanks for your suggestions - the wrapper change fixed it
```
./bin/etcdbrctl server \
     --storage-provider=Local \
     --store-container=primary-bucket \
     --secondary-storage-provider=Local \
     --secondary-store-container=secondary-bucket \
     --secondary-backup-sync-enabled=true \
     --secondary-backup-sync-period=120m
INFO[0000] etcd-backup-restore Version: v0.39.0-dev     
INFO[0000] Git SHA: 1eeb03ad                            
INFO[0000] Go Version: go1.25.4                         
INFO[0000] Go OS/Arch: darwin/arm64                     
<truncated>
secondarySnapstoreConfig:
  StoreConfig:
    container: secondary-bucket
    envPrefix: SECONDARY_
    maxParallelChunkUploads: 5
    minChunkSize: 5242880
    prefix: v2
    provider: Local
    tempDir: /tmp
  backupSyncEnabled: true
  syncPeriod: 2h0m0s
<truncated>
  ```

File: docs/usage/backup_sync_dual_site.md [hunk 37]
```diff
@@ -0,0 +1,241 @@
...
+```
+
+### Verification Steps
+The examples assume AWS S3 is the backend storage provider. Replace the commands with the appropriate ones for your provider
+
+1. **Check Primary Backups**:
+
+```bash
+aws s3 ls s3://<primary_bucket_name> --recursive
+aws s3 ls s3://<primary_bucket_name> --recursive
```
ishan16696: only 1 line should be there

example/00-backup-restore-server-config.yaml
```diff
@@ -33,6 +33,18 @@ snapstoreConfig:
   maxParallelChunkUploads: 5
   tempDir: "/tmp"
 
+# secondarySnapstoreConfig:
+#   StoreConfig:
+#     container: secondary-bucket
+#     envPrefix: SECONDARY_
+#     maxParallelChunkUploads: 5
+#     minChunkSize: 5242880
+#     prefix: v2
```
ishan16696:
```suggestion
#     prefix: "etcd-test"
```
just to be consistent with snapStore config example.

example/00-backup-restore-server-config.yaml
```diff
@@ -33,6 +33,18 @@ snapstoreConfig:
   maxParallelChunkUploads: 5
   tempDir: "/tmp"
 
+# secondarySnapstoreConfig:
+#   StoreConfig:
+#     container: secondary-bucket
+#     envPrefix: SECONDARY_
```
ishan16696, line 39: you can remove this envPrefix from example.

docs/usage/backup_sync_dual_site.md
ishan16696: ok, may you can add this line: "Similarly, backups present in other storage provider's bucket like Azure, GCS, Swift etc can be listed using their respective API calls or via their dashboards". wdyt ?

example/00-backup-restore-server-config.yaml
tchinmai7: wanted to keep this so users know how to provide config distinct to the secondary endpoint - thoughts?

ishan16696: ok, cool ... no strong opinion.

ishan16696 (approved):

PR #936, opened 2025-11-13
SSE-C key rotation

### LINKED ISSUE #934: Support for rotation of SSE-C keys.

/area backup
/kind api-change

I'm proposing to add support for reading backups from S3 using multiple keys while keeping one key as the active one for writing. This would allow rotation of these keys which is currently not possible.

**Why is this needed**:
Currently etcb support using SSE-C key while uploading and downloading the backups to S3, but this support just single key. That causes impossibility to remove the SSE-C encryption or changing the key once it was used. I'm proposing to add support for using multiple keys for reading (since there is no way to find out which key was used for upload, we would try one by one) and using the first key for writing. The proposal is to replace the original file with SSE-C key with json structure.

I'm working at Akamai (formerly Linode) on our internal implementation and I'll soon have a patch which we would be able to upstream, so I'd like to gather feedback if there would be a desire for that, since the SSE-C support is already there.

This is the example of encryption config file which would replace the sseCustomerKey:
```json
{
  "algorithm": "AES256",
  "disableEncryptionForWriting": false,
  "keys": [
    {
      "id": "primary-key-2024",
      "value": "base64-encoded-32-byte-key"
    },
    {
      "id": "backup-key-2023",
      "value": "another-base64-encoded-32-byte-key"
    }
  ]
}
```

akaokunc:

/area backup
/kind api-change

Fixes #934 

**Release note**:

```other operator

```


ghost:
@akaokunc Labels area/todo, kind/todo do not exist.

ghost:
@akaokunc Labels area/todo, kind/todo do not exist.

ghost:
@renormalize Labels area/todo, kind/todo do not exist.

ghost:
@akaokunc Command `/remove-lifecycle` is not known.

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,290 @@
...
+    {
+      "id": "key-2024-01", 
+      "value": "old-base64-key-data-here"
+    }
+  ]
+}
+```
+
+## Flags and Configuration
+SSE-C is configured by providing json file `sseCustomerKeyConf` in AWS_SSE_CUSTOMER_KEY_AWS_APPLICATION_CREDENTIALS_JSON path.
```
renormalize: The path seems off? It's just `AWS_APPLICATION_CREDENTIALS_JSON` path, right?

docs/usage/s3_serverside_snapshot_encryption.md
```diff
@@ -0,0 +1,290 @@
...
+  - **`id`** (string, required): Unique identifier for the key (used for tracking and logging).
+  - **`value`** (string, required): Base64-encoded 32-byte (256-bit) encryption key.
+
+### Multiple Key Support
+
+The system supports multiple SSE-C keys to enable key rotation and backward compatibility:
+
+1. **Writing**: Always uses the **first key** in the array (unless encryption is disabled).
+2. **Reading**: Attempts all keys in order until one successfully decrypts the backup.
+3. **Key Rotation**: Add new key as the first element, keep old keys for reading existing backups.
```
renormalize, line 42: What about key expiration? Would like to the document to talk about that as well.

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,290 @@
...
+
+- When SSE-C keys are configured, all backup operations will use SSE-C encryption by default
+- Set `disableEncryptionForWriting: true` to stop encrypting new backups while maintaining ability to read existing encrypted backups
+- Multiple keys enable seamless key rotation without backup service interruption
+
+## HTTP API
+
+### Get Encryption Status
+
+Returns the current encryption status of all known backup files. This can be out of date if the app didn't yet wrote/read all the files in the store. In such cases you can start asynchronous scan by the `POST /snapshot/scan` endpoint described below.
```
renormalize: Nit: grammar

```suggestion
Returns the current encryption status of all known backup files. This can be out of date if the app didn't yet write/read all the files in the store. In such cases you can start asynchronous scan by the `POST /snapshot/scan` endpoint described below.
```

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,290 @@
...
+
+**Endpoint:** `GET /snapshot/encstatus`
+
+**Requirements:** 
+- Only available on leader instances
```
renormalize: Non-leader instances of etcd-backup-restore forward all requests to the leader for `/snapshot` endpoints as can be seen in the other methods. I don't think this should be a requirement as such. This would also help usability.

See: https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2672038405

pkg/server/httpAPI.go
```diff
@@ -725,3 +736,97 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+
+// Update serveSnapshotsEncryptionStatus to return structured JSON
+func (h *HTTPHandler) serveSnapshotsEncryptionStatus(rw http.ResponseWriter, req *http.Request) {
+	h.checkAndSetSecurityHeaders(rw)
+	h.Logger.Info("Received request for SSE-C key usage status.")
+	if h.Snapshotter == nil || len(h.StorageProvider) == 0 {
+		h.Logger.Errorf("This request can only be served by leader.")
+		rw.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
```
renormalize, line 798: All methods exposed for SSE-C encryption on the backup-restore server could use the `delegateReqToLeader()` helper method to actually delegate the request to the leading backup-restore member. This not only makes the implementation easier, but also improves usability for consumers - they would not have to find out which member is the leader.

pkg/server/httpAPI.go
```diff
@@ -725,3 +736,97 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+
+	rw.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(rw).Encode(response); err != nil {
+		h.Logger.Errorf("Failed to encode SSE key usage: %v", err)
+		rw.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+}
+
+func (h *HTTPHandler) serveSnapshotsScan(rw http.ResponseWriter, req *http.Request) {
```
renormalize, line 821: https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2672038405

pkg/server/backuprestoreserver.go
```diff
@@ -251,6 +251,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 			if b.config.HealthConfig.EtcdMemberGCEnabled {
 				go membergarbagecollector.RunMemberGarbageCollectorPeriodically(leCtx, b.config.HealthConfig, b.logger, b.config.EtcdConnectionConfig)
 			}
+			handler.OnStartedLeading()
```
renormalize: We wouldn't need this given the context in: https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2672038405

pkg/server/backuprestoreserver.go
```diff
@@ -260,6 +261,7 @@ func (b *BackupRestoreServer) runServer(ctx context.Context, restoreOpts *brtype
 				// TODO @ishan16696: For Multi-node etcd HTTP status need to be set to `StatusServiceUnavailable` only when backup-restore is in "StateUnknown".
 				handler.SetStatus(http.StatusServiceUnavailable)
 			}
+			handler.OnStoppedLeading()
```
renormalize, line 286: We wouldn't need this given the context in: https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2672038405

pkg/server/httpAPI.go
```diff
@@ -725,3 +736,97 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+	store, err := snapstore.GetSnapstore(h.SnapstoreConfig)
+	if err != nil {
+		h.Logger.Errorf("Unable to create snapstore for re-encryption: %v", err)
+		return
+	}
+	s3store, ok := store.(*snapstore.S3SnapStore)
+	if !ok {
+		h.Logger.Errorf("Re-encryption is only supported for S3SnapStore")
+		return
+	}
```
renormalize: nit: A type switch would be preferred by me stylistically here instead of the type assertion here. It would be easier to extend to other providers with the type switch in my opinion.

pkg/server/httpAPI.go
```diff
@@ -139,6 +147,9 @@ func (h *HTTPHandler) RegisterHandler() {
 	mux.HandleFunc("/snapshot/full", h.serveFullSnapshotTrigger)
 	mux.HandleFunc("/snapshot/delta", h.serveDeltaSnapshotTrigger)
 	mux.HandleFunc("/snapshot/latest", h.serveLatestSnapshotMetadata)
+	mux.HandleFunc("/snapshot/reencrypt", h.serveSnapshotsReencrypt)
+	mux.HandleFunc("/snapshot/encstatus", h.serveSnapshotsEncryptionStatus)
+	mux.HandleFunc("/snapshot/scan", h.serveSnapshotsScan)
```
renormalize, line 155: After this PR, only S3 would serve these endpoints. The maintainers and I are trying to figure out if we should expose this endpoint as suggested, or if we should introduce provider specific path like `/s3` somewhere in the path to indicate that this is a provider specific feature. We've not reach any conclusion yet, but I'm letting you know that there's some confusion here still.

renormalize (changes requested):
Thanks for raising the PR @akaokunc!

Firstly, I'd like to apologize for this PR being stuck for so long.

I've not completed my review yet (I'm yet to take a thorough look at the provider specific changes), and I've added my initial comments here.

I think the feature would be overall helpful for consumers of etcd-backup-restore using S3 compatible providers, and lays the groundwork for similar features being contributed to in other providers. Again, thanks!

pkg/server/backuprestoreserver.go
akaokunc: I believe that in this case it would be needed even if we delegate the requests to leader. It is due to the need to wipe the cached information about encryption status which needs to happen whenever we lose leadership. But yes, forwarding the requests is a good idea and I'm going forward with implementing it.

pkg/server/httpAPI.go
akaokunc: I think that it is a good idea to separate it into the /s3 path. Please let me know if you guys land on some common ground and I will update it.

docs/usage/s3_serverside_snapshot_encryption.md
akaokunc: Added description of desired operation below.

docs/usage/snapshot_encryption.md
akaokunc: Makes sense. Implemented and updated docs.

pkg/server/backuprestoreserver.go
akaokunc: Makes sense. Implemented and updated docs.

pkg/server/httpAPI.go
akaokunc: Makes sense. Implemented and updated docs.

docs/usage/snapshot_encryption.md
akaokunc: Fixed

docs/usage/snapshot_encryption.md
akaokunc: fixed, thanks!

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
+# SSE-C (Server-Side Encryption with Customer-Provided Keys) Support
+
+ETCD backup-restore supports AWS S3 Server-Side Encryption with Customer-Provided Keys (SSE-C) for securing backup snapshots. This allows you to maintain full control over encryption keys while leveraging S3's server-side encryption capabilities.
```
renormalize:
```suggestion
etcd-backup-restore supports AWS S3 Server-Side Encryption with Customer-Provided Keys (SSE-C) for securing backup snapshots. This allows you to maintain full control over encryption keys while leveraging S3's server-side encryption capabilities.
```

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Configuration
+
+### Secret Format
+
+SSE-C configuration is provided via Kubernetes secret with the following JSON structure:
```
renormalize:
```suggestion
SSE-C configuration is provided via Kubernetes secret with the following JSON structure (as a part of the credential secret provided to etcd-backup-restore):
```

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
...
+      "id": "backup-key-2023",
+      "value": "another-base64-encoded-32-byte-key"
+    }
+  ]
+}
+```
+
+#### Field Descriptions
+
+- **`algorithm`** (string, required): SSE-C algorithm to use. Currently only `"AES256"` is supported.
```
renormalize:
```suggestion
- **`algorithm`** (string, required): SSE-C algorithm to use. Currently, only `"AES256"` is supported.
```

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
...
+    },
+    {
+      "id": "key-2024-01", 
+      "value": "old-base64-key-data-here"
+    }
+  ]
+}
+```
+
+3. **Key Expiration**: You can remove keys from the list, etcd-backup-restore will automatically reload the file as if you were adding keys. But beware, you should never remove keys which are still in use. It is recommended to rotate the keys by some time-based schema, which should respect, that older keys might be needed. For example if you have oldest full snapshot 4 weeks old, then you should never delete keys which were added within that period. It can happen, that you get in situation, that you would need to rotate the keys earlier. For such case there is a possibility to reencrypt all objects in the snapstore to make sure that all stored snapshots are encrypted by the most recent key. After succesfull reencryption it is ok, to remove the deprecated key. If in doubt, see the [Get Encryption Status](#get-encryption-status).
```
renormalize:
```suggestion
3. **Key Expiration**: You can remove keys from the list; etcd-backup-restore will automatically reload the file as it does when adding keys. However, never remove keys that are still in use. Rotate keys on a time-based schedule that accounts for the possibility that older keys are still needed. For example, if the oldest full snapshot is 4 weeks old, do not delete keys added within that period. If you must rotate keys earlier, re-encrypt all objects in the snapstore to ensure all snapshots are encrypted with the most recent key. After a successful re-encryption, it is safe to remove the deprecated key. If in doubt, see the [Get Encryption Status](#get-encryption-status).
```

docs/usage/snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Flags and Configuration
+SSE-C is configured by providing json file `sseCustomerKeyConf` in AWS_APPLICATION_CREDENTIALS_JSON path.
```
renormalize:
```suggestion
## Flags and Configuration

SSE-C is configured under `sseCustomerKeyConf` as a part of the credential file at the AWS_APPLICATION_CREDENTIALS_JSON path.
```

docs/usage/s3_serverside_snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
...
+    ]
+  }'
+```
+
+2. Configure the backup-restore deployment:
+
+```yaml
+env:
+- name: AWS_SSE_CUSTOMER_KEY_AWS_APPLICATION_CREDENTIALS_JSON
+# this can be part of the secret with AWS S3 bucket credentials or separated,
```
renormalize, line 178: How strongly do you feel about only supporting two files? I personally think just one file might make things simpler for consumers and developers, and so far all credentials to etcd-backup-restore are passed in a singular file; I would like to keep it that way unless there's a technical reason for wanting to seperate it into two files, in which case let's weigh the pros and cons.

docs/usage/s3_serverside_snapshot_encryption.md
```diff
@@ -0,0 +1,275 @@
+# SSE-C (Server-Side Encryption with Customer-Provided Keys) Support
```
renormalize, line 1: Can you rename the file to `s3_serverside_snapshot_encryption.md`. `snapshot_encryption.md` might be misleading since it might mean encryption of snapshots pre-upload for all providers, which is not the case here.

pkg/server/httpAPI.go
```diff
@@ -92,6 +92,20 @@ type HTTPHandler struct {
 	AckState                  uint32
 	EnableTLS                 bool
 	EnableProfiling           bool
+	isLeader                  bool
```
renormalize: I see value in adding this field, however, I would refrain from adding it right now and using the `h.Snapshotter == nil` pattern (as can be seen in the first few lines of `serveFullSnapshotTrigger`) that is being used in the methods implemented for the handler.

I don't want different ways of finding if the current process is the leading backup-restore or not - it makes the code inconsistent method to method. Either we refactor the other methods of the handler (which I would like to avoid at the moment), or we follow the current (albeit not as easy to read) pattern.

pkg/server/httpAPI.go
```diff
@@ -725,3 +742,108 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+			       h.Logger.Errorf("Snapshot scanning is only supported for S3SnapStore")
+			       return
+		       }
+	       }()
+	       rw.WriteHeader(http.StatusAccepted)
+	       rw.Write([]byte("Snapshot scan started.\n"))
+}
+
+// Add this method to your HTTPHandler or wherever you handle leadership changes
+func (h *HTTPHandler) OnLeadershipChanged(isLeader bool) {
```
renormalize: Given https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2686504785, perhaps we can just have a `OnStoppedLeading()` method since we don't really do anything when leadership is gained. This can be called in the server to ensure the key usage data is flushed.

renormalize (commented):
Added a few more comments. Thanks for addressing the previous comments. I will finally look at the S3 changes after this. Thanks!

pkg/server/httpAPI.go
```diff
@@ -725,3 +742,108 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+	       case *snapstore.S3SnapStore:
+		       go func() {
+			       s.ReencryptAllSnapshots(h.Logger)
+		       }()
+		       rw.WriteHeader(http.StatusAccepted)
+		       rw.Write([]byte("Re-encryption of all snapshots started.\n"))
+	       default:
+		       h.Logger.Errorf("Re-encryption is only supported for S3SnapStore")
+		       return
+	       }
```
renormalize: The indent in this block does not follow `gofmt` rules. You can format your changes with the `make format` target. You can also run `make check` to format and test your changes all in one go.

pkg/server/httpAPI.go
```diff
@@ -725,3 +742,108 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+				       return
+			       }
+			       h.Logger.Info("Snapshot scan completed successfully.")
+		       default:
+			       h.Logger.Errorf("Snapshot scanning is only supported for S3SnapStore")
+			       return
+		       }
+	       }()
+	       rw.WriteHeader(http.StatusAccepted)
+	       rw.Write([]byte("Snapshot scan started.\n"))
```
renormalize: https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2689093579

pkg/server/httpAPI.go
```diff
@@ -725,3 +742,108 @@ func IsBackupRestoreHealthy(backupRestoreURL string, TLSEnabled bool, rootCA str
...
+		       h.Logger.Info("Lost leadership, clearing SSE key usage cache")
+		       if h.Snapshotter != nil && h.Snapshotter.GetStore() != nil {
+			       switch s := h.Snapshotter.GetStore().(type) {
+			       case *snapstore.S3SnapStore:
+				       s.ClearSSEKeyUsageData()
+			       }
+		       }
+	       } else {
+		       h.Logger.Info("Gained leadership")
+	       }
```
renormalize: https://github.com/gardener/etcd-backup-restore/pull/936#discussion_r2689093579

renormalize (changes requested):

docs/usage/s3_serverside_snapshot_encryption.md
akaokunc: Done, good suggestion!

pkg/server/httpAPI.go
akaokunc: ACK, refactored this part.

pkg/server/httpAPI.go
akaokunc: Thanks for feedback. I must say that it is not that straightforward, but I do respect the intent not to do unrelated changes to the PR, so adopted that approach in my PR and left it for possible future refactor.

pkg/server/httpAPI.go
akaokunc: Did you guys end up on some common ground already?

docs/usage/s3_serverside_snapshot_encryption.md
akaokunc: This was mainly thought as a way to easily support rollout / rollback of SSE-C. If it would be in a single secret we would simply drop the whole secret while rolling back. But if it is preferred to have it just in a single file, then I can update the implementation and we will deal with possible rollback differently, we could likely keep it there anyway.

PR #952, opened 2025-12-08
Introduce the `--store-endpoint-override` flag to pass storage endpoint configuration, instead of using the credential secret

### LINKED ISSUE #943: Introduce capability to override cloud provider bucket's endpoint
/area control-plane
/kind enhancement

etcd-backup-restore configuration should not be passed through the credential secret. Configuration to override the endpoint must be passed as a flag.

For example, #691 added option the storage API endpoint for GCS to be overridden, however this is configured via the credentials secret, instead such configurations should be made either via the [config](https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/etcd-configmap.yaml) ConfigMap or via command flags.

**Why is this needed**:
Segregates credentials from configurations.

renormalize:

/area backup compliance robustness
/kind enhancement

This PR introduces changes to pass storage endpoint configuration through CLI flags, instead of information passed in the credential secret.

Passing endpoint information through credential files is deprecated, and this handling will be completely removed in v0.42.0.

This PR also removes all code in etcd-backup-restore that makes it aware that the storage backend is an emulator and not real infrastructure. etcd-backup-restore must not be aware of this information ideally.

Fixes #943 

**Release note**:

```other operator
Storage endpoint configuration must now be passed through the `--store-endpoint-override` flag instead of a field in the credential secret. The older method of passing endpoint through credential files is now deprecated.
```

### DIFF
--- a/docs/deployment/getting_started.md
+++ b/docs/deployment/getting_started.md
@@ -18,18 +18,18 @@ The procedure to provide credentials to access the cloud provider object store v
       1. Web Identity Role: `roleARN` and `tokenPath` must be set.
    2. For `S3-compatible providers` such as MinIO, `s3ForcePathStyle`, `insecureSkipVerify`, `trustedCaCert`, `requestChecksumCalculation` and `responseChecksumValidation`, can also be made available in an above file to configure the S3 client to communicate to a non-AWS provider.
    3. To enable Server-Side Encryption using Customer Managed Keys for `S3-compatible providers`, use `sseCustomerKey` and `sseCustomerAlgorithm` in the credentials file above. For example, `sseCustomerAlgorithm` could be set to `AES256`, and correspondingly the `sseCustomerKey` is set to a valid AES-256 key.
-   4. To override the endpoint of the S3 compatible API, use the `--store-endpoint-override` flag and pass the corresponding URL. This can be used for S3 compatible providers like Ceph, MinIO, etc., or while using the localstack S3 emulator.
+   4. To override the endpoint of the S3 compatible API, use the `--store-endpoint-override` flag and pass the corresponding URL. This can be used for S3 compatible providers like Ceph, MinIO, etc., or while using the localstack S3 emulator. For example, if the localstack emulator is running on your host along with etcd-backup-restore, `--store-endpoint-override=http://localhost:4566` should be passed.
 
 * For `Google Cloud Storage`:
    1. GCS supports two alternative authentication options:
       1. Static credentials, they should be provided via a file in the `~/.gcp` folder named `service-account-file.json`.This file must have field named `type` with value `service_account`.
       1. Workload Identity credentials, they should be provided via a file in the `~/.gcp` folder named `credentialsConfig` which have field named `type` with value `external_account`. Additionally, it is expected in the `~/.gcp` folder files named `projectID` and `token` to be provided, they contain the GCP project ID and the JWT respectively.
    2. The service account json or the credentials config file should be provided, and the path to it should be made available as environment variable `GOOGLE_APPLICATION_CREDENTIALS`.
-   3. If using a storage API [endpoint override](https://pkg.go.dev/cloud.google.com/go#hdr-Endpoint_Override), such as a [regional endpoint](https://cloud.google.com/storage/docs/regional-endpoints) or a local GCS emulator endpoint, the endpoint must be passed through the `--store-endpoint-override` flag.
+   3. If using a storage API [endpoint override](https://pkg.go.dev/cloud.google.com/go#hdr-Endpoint_Override), such as a [regional endpoint](https://cloud.google.com/storage/docs/regional-endpoints) or a local GCS emulator endpoint, the endpoint must be passed through the `--store-endpoint-override` flag. For example, if the fake-gcs-emulator is running on your host along with etcd-backup-restore, `--store-endpoint-override=http://localhost:8000/storage/v1/` should be passed.
 
 * For `Azure Blob storage`:
    1. The JSON secret file should be provided, and the file path should be made available as an environment variable: `AZURE_APPLICATION_CREDENTIALS`.
-   2. The Azure Blob Storage domain can be overridden by passing the overriding endpoint through the `--store-endpoint-override` flag. This is to be used if the endpoint of your Azure API is not the default (like Azure China or Azure Government), or while using the Azurite emulator.
+   2. The Azure Blob Storage domain can be overridden by passing the overriding endpoint through the `--store-endpoint-override` flag. This is to be used if the endpoint of your Azure API is not the default (like Azure China or Azure Government), or while using the Azurite emulator. For example, if Azurite is running on your host along with etcd-backup-restore, `--store-endpoint-override=http://localhost:10000/${STORAGE_ACCOUNT}` should be passed.
 
 * For `Openstack Swift`:
   1. The secret file should be provided, and the file path should be made available as an environment variable: `OPENSTACK_APPLICATION_CREDENTIALS`.


pkg/snapstore/abs_snapstore.go
```diff
@@ -110,22 +106,14 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
...
-	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain, emulatorEnabled)
-	if err != nil {
-		return nil, fmt.Errorf("failed to construct the blob service URL with error: %w", err)
+	// Construct the ABS Container endpoint.
+	// TODO: @renormalize support for passing Domain through the credential file must be removed in v0.42.0.
+	domain := ptr.Deref(absCreds.Domain, brtypes.AzureBlobStorageGlobalDomain)
+	containerURL := fmt.Sprintf("https://%s.%s/%s", absCreds.StorageAccount, domain, config.Container)
+	// endpoint specified as a CLI flag takes precedence over configuration passed in the credential file.
+	if config.EndpointOverride != "" {
+		containerURL = config.EndpointOverride
```
CaptainIRS: This seems to lead to `"failed to get properties of the container etcdbr-e2e-test-2a3b8912a3556cbeaf7de88108d21290091094e3 with error: Get \"https://devstoreaccount1.localhost:10000/etcdbr-e2e-test-2a3b8912a3556cbeaf7de88108d21290091094e3?restype=container\": dial tcp: lookup devstoreaccount1.localhost: no such host"` when running the e2e tests with the Azure provider.

I think the proper way to solve this is to add a `Values.backup.endpointOverride` to the helm chart and use that to setup the flags in the StatefulSet. This would be populated in `integrationcluster_suite_test.go` with values populated in `test/e2e/utils.go`'s `testProvider.endpoint` (to something like `fmt.Sprintf("http://%s/%s", getEnvOrFallback(envAzuriteDomain, brtypes.AzureBlobStorageGlobalDomain), azureStorageAccount)`)

If there is any other easier way, feel free to ignore this approach.

pkg/snapstore/gcs_snapstore.go
```diff
@@ -92,51 +83,59 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
-			return nil, fmt.Errorf("emulator enabled, but `storageAPIEndpoint` not provided")
-		}
 	}
 
-	var chunkDirSuffix string
-	if emulatorConfig.enabled {
-		opts, err = emulatorConfig.configureClient(opts)
+	// `http` communication is disallowed by GCS APIs, so we assume an emulator is being used when communication is over `http`.
+	// The GCS SDK requires envGoogleStorageEmulatorHost to be set to the emulator's endpoint.
+	if strings.Contains(endpoint, "http://") {
```
CaptainIRS: Nit: A better way to do this is to use `url.Parse` and compare the `Scheme` field from the obtained `net.URL` object

CaptainIRS (changes requested):
@renormalize Thanks for the PR! I've added a couple of comments.

Also, one thing to note is that since we're removing all logic that depends on the `emulatorEnabled` secret field in the tests, this could also be removed from the helm charts and the e2e test utils wherever this is being set.

ishan16696:
Please update the  example [here](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/04-google-cloud-storage-secret.yaml#L11) and [here](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/00-azure-blob-storage-secret.yaml#L11). Also update the [doc](https://github.com/gardener/etcd-backup-restore/blob/master/docs/deployment/getting_started.md#passing-credentials).

pkg/snapstore/abs_snapstore.go
```diff
@@ -110,22 +106,14 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
...
-	}
-
-	blobServiceURL, err := ConstructBlobServiceURL(absCreds.StorageAccount, domain, emulatorEnabled)
-	if err != nil {
-		return nil, fmt.Errorf("failed to construct the blob service URL with error: %w", err)
+	// Construct the ABS Container endpoint.
+	// TODO: @renormalize support for passing Domain through the credential file must be removed in v0.42.0.
+	domain := ptr.Deref(absCreds.Domain, brtypes.AzureBlobStorageGlobalDomain)
+	containerURL := fmt.Sprintf("https://%s.%s/%s", absCreds.StorageAccount, domain, config.Container)
+	// endpoint specified as a CLI flag takes precedence over configuration passed in the credential file.
```
ishan16696: Is it make sense to take both ? or atleast we should we should print a warning mentioning that "credential files is now deprecated, refer to this doc."

pkg/snapstore/gcs_snapstore.go
```diff
@@ -92,51 +83,59 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
-	var opts []option.ClientOption // no need to explicitly set store credentials here since the Google SDK picks it up from the standard environment variable
-	var emulatorConfig gcsEmulatorConfig
-	emulatorConfig.enabled = config.IsEmulatorEnabled || isEmulatorEnabled(config)
-	endpoint, err := getGCSStorageAPIEndpoint(config)
+	// TODO: @renormalize support for passing Endpoint through the credential file must be removed in v0.42.0.
+	endpoint, err := getGCSStorageAPIEndpointFromFile(config)
 	if err != nil {
 		return nil, err
 	}
+	// endpoint specified as a CLI flag takes precedence over configuration passed in the credential file.
```
ishan16696: same as above

pkg/snapstore/s3_snapstore.go
```diff
@@ -206,6 +210,7 @@ func awsCredentialsFromConfig(awsConfig *awsCredentials) ([]func(*awsconfig.Load
 	}
 	cfgOpts = append(cfgOpts, awsconfig.WithCredentialsProvider(aws.NewCredentialsCache(credentialsProvider)))
 
+	// TODO: @renormalize support for passing Endpoint through the credential file must be removed in v0.42.0.
```
ishan16696, line 213: same as above , also update the example [here](https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/01-aws-s3-storage-secret.yaml#L13) as most of our community users use S3 compatible providers. Can you mention this in release note may be like this:
```
Storage endpoint configuration for all storage providers must be passed through the `--store-endpoint-override` flag instead of a field in the credential secret file. Please refer to this [doc](... doc of passing credentials). The older method of passing endpoint through credential files is now deprecated.
```

pkg/snapstore/snapshot.go
```diff
@@ -116,10 +116,6 @@ func ParseSnapshot(snapPath string) (*brtypes.Snapshot, error) {
 	//parse creation time as well as parse the Snapshot compression suffix
 	lastNameToken := strings.Split(tokens[3], "/")
 	timeWithSnapSuffix := strings.Split(lastNameToken[0], ".")
-	// Check & remove if the last token is a chunk directory. The chunkDirSuffix is set by only GCS snapstore when using a emulator for testing.
-	if fmt.Sprintf(".%s", timeWithSnapSuffix[len(timeWithSnapSuffix)-1]) == brtypes.ChunkDirSuffix {
-		timeWithSnapSuffix = timeWithSnapSuffix[:len(timeWithSnapSuffix)-1]
-	}
```
ishan16696, line 122: any reason to remove this ?

pkg/types/snapstore.go
```diff
@@ -222,6 +219,7 @@ func (c *SnapstoreConfig) AddSourceFlags(fs *flag.FlagSet) {
 func (c *SnapstoreConfig) addFlags(fs *flag.FlagSet, parameterPrefix string) {
 	fs.StringVar(&c.Provider, parameterPrefix+"storage-provider", c.Provider, "snapshot storage provider")
 	fs.StringVar(&c.Container, parameterPrefix+"store-container", c.Container, "container which will be used as snapstore")
+	fs.StringVar(&c.EndpointOverride, parameterPrefix+"store-endpoint-override", c.EndpointOverride, "endpoint that will be used to override the default of the container that is used as snapstore")
```
ishan16696: what do you mean by "default of the container" ? not every provider use terminology `container`, better to use some generic terminology. Can we rephrase this with something like this: 
```suggestion
	fs.StringVar(&c.EndpointOverride, parameterPrefix+"store-endpoint-override", c.EndpointOverride, "endpoint that will be used to override the default endpoint of storage provider.")
```

pkg/snapstore/abs_snapstore.go
renormalize: I did this is to ensure there is backwards compatibility between releases for consumers. If an override is specified in the credential file, it will continue to work as it has; but the `store-endpoint-override` would be preferred over the `domain` specified in the credential file.

This support would also allow to easily revert the release in case any unwanted behavior is found.

pkg/snapstore/snapshot.go
renormalize: This was removed since `ChunkDirSuffix` was introduced in #697 to support the fake-gcs-emulator with etcd-backup-restore. The emulator has some limitations while using disk as the storage backend and needed these suffixes to be added. I have switched to using memory as the backend for the emulator which does not have the same limitation, and therefore removed the `ChunkDirSuffix` since it is no longer needed in the codebase. It also simplifies the code and makes it more production-relevant by removing code that only assists in using emulators.

pkg/snapstore/abs_snapstore.go
renormalize: Addressed in [94026cc](https://github.com/gardener/etcd-backup-restore/pull/952/commits/94026cc2f5303ecfce06c0281abf51f28aeb5162). Printing a warning log.

Also updated documentation in [3d2aec4](https://github.com/gardener/etcd-backup-restore/pull/952/commits/3d2aec43568692e65ace644b787f2b53364bccda).

pkg/snapstore/gcs_snapstore.go
renormalize: Addressed in [94026cc](https://github.com/gardener/etcd-backup-restore/pull/952/commits/94026cc2f5303ecfce06c0281abf51f28aeb5162). Printing a warning log.

pkg/types/snapstore.go
renormalize: Addressed in [94026cc](https://github.com/gardener/etcd-backup-restore/pull/952/commits/94026cc2f5303ecfce06c0281abf51f28aeb5162).

pkg/snapstore/s3_snapstore.go
renormalize: Also updated documentation in [3d2aec4](https://github.com/gardener/etcd-backup-restore/pull/952/commits/3d2aec43568692e65ace644b787f2b53364bccda).

pkg/snapstore/gcs_snapstore.go
renormalize: Addressed in [94026cc](https://github.com/gardener/etcd-backup-restore/pull/952/commits/94026cc2f5303ecfce06c0281abf51f28aeb5162).

pkg/snapstore/snapshot.go
ishan16696: thanks for giving me the context and explaining it.

docs/deployment/getting_started.md
```diff
@@ -16,19 +16,20 @@ The procedure to provide credentials to access the cloud provider object store v
...
-   2. For `S3-compatible providers` such as MinIO, `endpoint`, `s3ForcePathStyle`, `insecureSkipVerify`, `trustedCaCert`, `requestChecksumCalculation` and `responseChecksumValidation`, can also be made available in an above file to configure the S3 client to communicate to a non-AWS provider.
+   2. For `S3-compatible providers` such as MinIO, `s3ForcePathStyle`, `insecureSkipVerify`, `trustedCaCert`, `requestChecksumCalculation` and `responseChecksumValidation`, can also be made available in an above file to configure the S3 client to communicate to a non-AWS provider.
    3. To enable Server-Side Encryption using Customer Managed Keys for `S3-compatible providers`, use `sseCustomerKey` and `sseCustomerAlgorithm` in the credentials file above. For example, `sseCustomerAlgorithm` could be set to `AES256`, and correspondingly the `sseCustomerKey` is set to a valid AES-256 key.
+   4. To override the endpoint of the S3 compatible API, use the `--store-endpoint-override` flag and pass the corresponding URL. This should be used for S3 compatible providers like Ceph, MinIO, etc., or while using the localstack S3 emulator.
```
ishan16696:
```suggestion
   4. To override the endpoint of the S3 compatible API, use the `--store-endpoint-override` flag and pass the corresponding URL. This can be used for S3 compatible providers like Ceph, MinIO, etc., or while using the localstack S3 emulator.
```

pkg/snapstore/abs_snapstore.go
```diff
@@ -110,22 +106,19 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		return nil, fmt.Errorf("failed to create sharedKeyCredential: %w", err)
 	}
 
-	emulatorEnabled := config.IsEmulatorEnabled || absCreds.EmulatorEnabled
+	// Construct the ABS Container endpoint.
+	// TODO: @renormalize support for passing Domain through the credential file must be removed in v0.42.0.
 	domain := brtypes.AzureBlobStorageGlobalDomain
 	if absCreds.Domain != nil {
+		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the --store-endpoint-override flag instead.")
```
ishan16696:
```suggestion
		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the `--store-endpoint-override` flag instead.")
```

pkg/snapstore/gcs_snapstore.go
```diff
@@ -92,39 +83,23 @@ func NewGCSSnapStore(config *brtypes.SnapstoreConfig) (*GCSSnapStore, error) {
...
-		opts = append(opts, option.WithEndpoint(endpoint))
-		if emulatorConfig.enabled {
-			emulatorConfig.endpoint = endpoint
-		}
-	} else {
-		// if emulator is enabled, but custom storage API endpoint for the emulator is not provided, throw error
-		if emulatorConfig.enabled {
-			return nil, fmt.Errorf("emulator enabled, but `storageAPIEndpoint` not provided")
-		}
+		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the --store-endpoint-override flag instead.")
```
ishan16696:
```suggestion
		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the `--store-endpoint-override` flag instead.")
```

pkg/snapstore/s3_snapstore.go
```diff
@@ -206,7 +210,9 @@ func awsCredentialsFromConfig(awsConfig *awsCredentials) ([]func(*awsconfig.Load
...
 	if awsConfig.Endpoint != nil {
+		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the --store-endpoint-override flag instead.")
```
ishan16696:
```suggestion
		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the `--store-endpoint-override` flag instead.")
```

pkg/snapstore/abs_snapstore.go
```diff
@@ -110,22 +106,19 @@ func NewABSSnapStore(config *brtypes.SnapstoreConfig) (*ABSSnapStore, error) {
 		return nil, fmt.Errorf("failed to create sharedKeyCredential: %w", err)
 	}
 
-	emulatorEnabled := config.IsEmulatorEnabled || absCreds.EmulatorEnabled
+	// Construct the ABS Container endpoint.
+	// TODO: @renormalize support for passing Domain through the credential file must be removed in v0.42.0.
```
shreyas-s-rao, line 111: Can you please create an issue for this with milestone v0.42.0, so this is not missed? Applicable to all 3 providers.

pkg/snapstore/s3_snapstore.go
```diff
@@ -90,6 +90,10 @@ func NewS3SnapStore(config *brtypes.SnapstoreConfig) (*S3SnapStore, error) {
 		return nil, err
 	}
 
+	if config.EndpointOverride != "" {
+		cfgOpts = append(cfgOpts, awsconfig.WithBaseEndpoint(config.EndpointOverride))
+	}
+
```
shreyas-s-rao, line 96: This should override the check on L214, which means this should be run after L214 in such a way that this takes precedence over the `if awsConfig.Endpoint != nil` check if that turns out to be true. Basically, we should avoid adding `awsconfig.WithBaseEndpoint()` twice to `cfgOpts` in the case where user uses both old and new ways of setting the endpoint override.

renormalize: We first set the option in L214 like so:

```go
	if awsConfig.Endpoint != nil {
		logrus.Warnf("Passing endpoint override through the credential file is now deprecated. Please use the `--store-endpoint-override` flag instead.")
		cfgOpts = append(cfgOpts, awsconfig.WithBaseEndpoint(*awsConfig.Endpoint))
	}
```

and again in L93 like so:

```go
	if config.EndpointOverride != "" {
		cfgOpts = append(cfgOpts, awsconfig.WithBaseEndpoint(config.EndpointOverride))
	}
```

The `cfgOpts` are all closures which are called when the config is loaded in L97:

```go
	cfg, err := awsconfig.LoadDefaultConfig(context.TODO(), cfgOpts...)
```

The consequence of this is that the option passed in L214 is loaded into `LoadOptions.BaseEndpoint`, and then the option passed in L93 is loaded into `LoadOptions.BaseEndpoint`, the latter overriding the former.

If both are set, the newer method is considered and I think that is good enough behavior that we expose to the user. Being more defensive with the code to check if both methods are present just complicates the code in my opinion.

pkg/snapstore/s3_snapstore.go
renormalize: Addressed in [8d48f85](https://github.com/gardener/etcd-backup-restore/pull/952/commits/8d48f85fad0434df8fdc6cb989c815174af90db2).

pkg/snapstore/gcs_snapstore.go
renormalize: Addressed in [8d48f85](https://github.com/gardener/etcd-backup-restore/pull/952/commits/8d48f85fad0434df8fdc6cb989c815174af90db2).

pkg/snapstore/abs_snapstore.go
renormalize: Addressed in [8d48f85](https://github.com/gardener/etcd-backup-restore/pull/952/commits/8d48f85fad0434df8fdc6cb989c815174af90db2).

docs/deployment/getting_started.md
renormalize: Addressed in [8d48f85](https://github.com/gardener/etcd-backup-restore/pull/952/commits/8d48f85fad0434df8fdc6cb989c815174af90db2).

pkg/snapstore/abs_snapstore.go
renormalize: Will address this at the end of the PR.

pkg/types/snapstore.go
```diff
@@ -222,6 +219,7 @@ func (c *SnapstoreConfig) AddSourceFlags(fs *flag.FlagSet) {
 func (c *SnapstoreConfig) addFlags(fs *flag.FlagSet, parameterPrefix string) {
 	fs.StringVar(&c.Provider, parameterPrefix+"storage-provider", c.Provider, "snapshot storage provider")
 	fs.StringVar(&c.Container, parameterPrefix+"store-container", c.Container, "container which will be used as snapstore")
+	fs.StringVar(&c.EndpointOverride, parameterPrefix+"store-endpoint-override", c.EndpointOverride, "endpoint that will be used to override the default endpoint of the storage provider")
```
ishan16696, line 222: I have one doubt:
how this `--store-endpoint-override` will work in case of copy backup jobs ?
I think we also need to make it configurable for source bucket in case of copy-backup, you can do it in a separate PR to keep the efforts simple as right now we are not dropping the support for passing the endpoint via secret file.

pkg/snapstore/s3_snapstore.go
shreyas-s-rao: Sure, I wasn't aware of how the option overriding would work in this case. Thanks for the explanation.

pkg/types/snapstore.go
renormalize: Will check to see if any further changes are necessary. I hadn't considered this, thanks for pointing it out! I'll raise a different PR to bring in changes if any are required for this.

CaptainIRS: I don't think any changes are required. We'll have an additional flag called `--source-store-endpoint-override` (thanks to `parameterPrefix`), which will correctly populate the source SnapStore.

renormalize: That's correct, thanks for checking @CaptainIRS!

pkg/snapstore/abs_snapstore.go
renormalize: Created #970 for this.

pkg/snapstore/abs_snapstore.go
renormalize: Addressed in [fa9387e](https://github.com/gardener/etcd-backup-restore/pull/952/commits/fa9387e2c2c161b351ad381847763784e303022f).

hack/config/aws_config.sh
```diff
@@ -30,6 +30,6 @@ echo -n "${AWS_SECRET_ACCESS_KEY}" > "${AWS_APPLICATION_CREDENTIALS}/secretAcces
 echo -n "${AWS_DEFAULT_REGION}"    > "${AWS_APPLICATION_CREDENTIALS}/region"
 
 if [[ "${USE_LOCALSTACK}" == "true" ]]; then
-  echo -n "${AWS_ENDPOINT_URL_S3}"   > "${AWS_APPLICATION_CREDENTIALS}/endpoint"
+  export EMULATOR_URL="http://localhost:4566"
```
CaptainIRS, line 33: The `EMULATOR_URL` being set and propagated across the e2e test suite seems to be used only in checks like `if emulatorUrl != ""` or similar in scripts. In essence, the string value itself doesn't seem to be used anywhere. So, could this instead be simplified into an `IS_EMULATOR` variable set in `run-e2e-tests.sh`? Or maybe I'm missing something.

renormalize: The value is used in `test/e2e/integrationcluster_suite_test.go` and `test/e2e/backup_test.go` in `getSnapstore()`. There are `store.List()` calls which would need the endpoint of the bucket to be configured correctly to have the `List` calls succeed.

Since these tests run on the host machine, we need to configure endpoint override to be `http://localhost:*`, which is a different override from what the etcd-backup-restore container needs which runs on the kind cluster which would be `http://<service-name>.*`.

So no, I do not think it can be a boolean.

CaptainIRS: Oh okay, I missed that, sorry. Thanks for the explanation.

PR #954, opened 2025-12-09
Disable auth in embedded etcd

### LINKED ISSUE #950: Embedded etcd does not work when base snaphot has etcd auth enabled

/area backup
/kind bug

**What happened**:

Any operation that requires starting an embedded etcd does not work when base snapshot has etcd auth enabled.

When restoring a backup with incremental snapshots, output is something like this:

```
$ go run main.go restore --storage-provider=Local --store-container=/path/to/backup
INFO[0000] Finding latest set of snapshot to recover from... 
INFO[0000] Restoring from base snapshot: Full-00000000-8391520845-1763683200  actor=restorer
INFO[0000] successfully fetched data of base snapshot in 0.553876178 seconds  actor=restorer
INFO[0001] Starting embedded etcd server...              actor=restorer
INFO[0002] Embedded server is ready to listen client at: 127.0.0.1:43351  actor=restorer
INFO[0002] Applying delta snapshots...                   actor=restorer
INFO[0002] Applying first delta snapshot Incr-8391520846-8391652657-1763684913  actor=restorer
INFO[0002] successfully fetched data of delta snapshot in 0.093521816 seconds  actor=restorer
FATA[0004] Failed to restore snapshot: failed to get etcd latest revision: etcdserver: user name is empty 
```

Embedded etcd should handle etcd authentication properly.

- Enable etcd auth on an etcd cluster (`etcdctl auth enable`)
- Backup the etcd cluster using this project, make sure there are incremental snapshots too.
- Try to run any operation which requires starting the embedded etcd server.

- We "fixed" this issue by manually disabling auth after starting the embedded etcd server:

  ```go
  e.Server.AuthStore().AuthDisable()
  ```

  It then needs to be re-enabled manually.

**Environment**:

- etcd-backup-restore version: v0.40.0

Tomy2e:

/area backup
/kind bug

This PR temporarily disables etcd authentication after starting the embedded etcd (if needed). It's enabled again before stopping the embedded etcd server.

Fixes #950

**Release note**:

```bugfix operator
Fixed the restoration failures that could occur when etcd backups was taken from etcd which has auth enabled.
```

### DIFF
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -223,7 +223,8 @@ func StartEmbeddedEtcd(logger *logrus.Entry, ro *brtypes.RestoreOptions) (*Embed
 	case <-e.Server.ReadyNotify():
 		logger.Infof("Embedded server is ready to listen client at: %s", e.Clients[0].Addr())
 	case <-time.After(60 * time.Second):
-		e.Close() // trigger a shutdown
+		e.Server.Stop() // trigger a shutdown
+		e.Close()
 		return nil, fmt.Errorf("server took too long to start")
 	}
 


Tomy2e:
Hi @ishan16696, my fix is ready for review :pray:

It looks like gitguardian secret scanner doesn't like the password I hardcoded in my test: https://github.com/gardener/etcd-backup-restore/blob/b60e1c8a74d85bc5825e4de40b97c89f56b1c6a5/pkg/snapshot/restorer/restorer_test.go#L1101
Do not hesitate to let me know if there is something I can do to fix that.

pkg/initializer/validator/datavalidator.go
```diff
@@ -371,10 +371,7 @@ func (d *DataValidator) checkFullRevisionConsistency(dataDir string, latestSnaps
 		d.Logger.Infof("unable to start embedded etcd: %v", err)
 		return DataDirectoryCorrupt, err
 	}
-	defer func() {
-		e.Server.Stop()
-		e.Close()
-	}()
+	defer e.Close()
```
ishan16696: same as above

pkg/compactor/compactor.go
```diff
@@ -91,10 +91,7 @@ func (cp *Compactor) Compact(ctx context.Context, opts *brtypes.CompactOptions)
 		}
 	}
 
-	defer func() {
-		embeddedEtcd.Server.Stop()
-		embeddedEtcd.Close()
-	}()
+	defer embeddedEtcd.Close()
```
ishan16696, line 94: I think it's better to stop the server first then close it for garcefull termination even if you directly call the close here, it will internally stop the server first before closing it , so it's doesn't make any difference.
https://github.com/etcd-io/etcd/blob/c123b3ea3db33636011c304d3bfb337d3ded84cf/embed/etcd.go#L417-L420

pkg/snapshot/restorer/restorer.go
```diff
@@ -68,15 +67,14 @@ func (r *Restorer) RestoreAndStopEtcd(ro brtypes.RestoreOptions, m member.Contro
 	embeddedEtcd, err := r.Restore(ro, m)
 	defer func() {
 		if embeddedEtcd != nil {
-			embeddedEtcd.Server.Stop()
```
ishan16696, line 71: same as above

pkg/initializer/validator/datavalidator.go
ishan16696, line 374: same as above

pkg/miscellaneous/miscellaneous.go
```diff
@@ -155,8 +156,31 @@ func getStructuredBackupList(snapList brtypes.SnapList) []backup {
...
+type EmbeddedEtcd struct {
+	*embed.Etcd
+
+	// authWasDisabled tracks if auth was disabled during startup and we need to
+	// enable it again before closing the server.
+	authWasDisabled bool
+}
+
+// Close closes the etcd embedded server. If auth was disabled during startup,
+// we enable it again before closing.
```
```suggestion
// Close closes the etcd embedded server. If auth was disabled during startup,
// it will re-enable auth before closing the server.
```

test/utils/utils.go
```diff
@@ -249,16 +285,32 @@ func RunSnapshotter(logger *logrus.Entry, snapstoreConfig brtypes.SnapstoreConfi
...
+
+	// If auth is enabled, we expect a username and password to be provided.
+	// Otherwise, there must be no username or password.
+	if etcd.Server.AuthStore().IsAuthEnabled() {
+		if username == "" || password == "" {
+			return fmt.Errorf("a username and password must be set when auth is enabled")
+		}
+	} else {
+		if username != "" || password != "" {
+			return fmt.Errorf("a username or password should not be set when auth is disabled")
```
ishan16696:
```suggestion
			return fmt.Errorf("a username or password should not be set as etcd auth found to be disabled")
```

test/utils/utils.go
```diff
@@ -249,16 +285,32 @@ func RunSnapshotter(logger *logrus.Entry, snapstoreConfig brtypes.SnapstoreConfi
...
 	if err != nil {
 		return fmt.Errorf("unable to start embedded etcd server: %v", err)
 	}
 	defer etcd.Close()
+
+	// If auth is enabled, we expect a username and password to be provided.
+	// Otherwise, there must be no username or password.
+	if etcd.Server.AuthStore().IsAuthEnabled() {
+		if username == "" || password == "" {
+			return fmt.Errorf("a username and password must be set when auth is enabled")
```
ishan16696:
```suggestion
			return fmt.Errorf("a username and password must be set as etcd auth found to be enabled")
```

test/utils/utils.go
```diff
@@ -249,16 +285,32 @@ func RunSnapshotter(logger *logrus.Entry, snapstoreConfig brtypes.SnapstoreConfi
...
+// A valid username and password must be provided if etcd auth is enabled.
+func CheckDataConsistency(ctx context.Context, dir string, keyTo int, username, password string, logger *logrus.Entry) error {
 	etcd, err := StartEmbeddedEtcd(ctx, dir, logger, DefaultEtcdName, EmbeddedEtcdPortNo)
 	if err != nil {
 		return fmt.Errorf("unable to start embedded etcd server: %v", err)
 	}
 	defer etcd.Close()
+
+	// If auth is enabled, we expect a username and password to be provided.
+	// Otherwise, there must be no username or password.
```
ishan16696:
```suggestion
	// If auth is enabled then username and password must be provided.
	// Otherwise, no username or password should be provided.
```

test/utils/utils.go
```diff
@@ -182,9 +185,39 @@ func PopulateEtcd(ctx context.Context, logger *logrus.Entry, endpoints []string,
...
 	defer wg.Done()
-	PopulateEtcd(ctx, logger, endpoints, 0, math.MaxInt64, resp)
+	PopulateEtcd(ctx, logger, endpoints, username, password, 0, math.MaxInt64, resp)
+}
+
+// EnableEtcdAuth does the following on an existing etcd cluster that has auth disabled:
+//   - Create a new user using the provided name and password.
+//   - Grant "root" role to the created user.
+//   - Enable auth in etcd.
+func EnableEtcdAuth(ctx context.Context, endpoints []string, username, password string) error {
```
ishan16696: this function `EnableEtcdAuth` is doing more than what name is suggesting, may be we can rename this which can better tell us what it's doing. example: `CreateRootUserAndEnableAuth` or something like this.

pkg/snapshot/restorer/restorer_test.go
```diff
@@ -1067,6 +1067,129 @@ var _ = Describe("Running Restorer when both v1 and v2 directory structures are
...
+		maxFetchers             uint   = 6
+		maxCallSendMsgSize             = 2 * 1024 * 1024 //2Mib
+		maxRequestBytes                = 2 * 1024 * 1024 //2Mib
+		maxTxnOps                      = 2 * 1024
+		embeddedEtcdQuotaBytes  int64  = 8 * 1024 * 1024 * 1024
+		autoCompactionMode      string = "periodic"
+		autoCompactionRetention string = "0"
+		embeddedEtcdPortNo      string = "9089"
+		rootUsername                   = "root"
+		rootPassword                   = "toor"
```
ishan16696: I'm also not quite sure about how to avoid GitGuardian detection of password, may be you can try with wrong spelling of password eg: `psswrd` 🤷

Tomy2e: It worked with `rootPssword`, thank you

pkg/compactor/compactor.go
Tomy2e: OK I added a call to `Server.Stop()` in method [EmbeddedEtcd.Close()](https://github.com/gardener/etcd-backup-restore/blob/e509318510205afc7a8b85e865e620d61e57ea51/pkg/miscellaneous/miscellaneous.go#L170) to preserve the existing behavior:
https://github.com/gardener/etcd-backup-restore/blob/e509318510205afc7a8b85e865e620d61e57ea51/pkg/miscellaneous/miscellaneous.go#L177

Calling `embeddedEtcd.Server.Stop()` should no longer be done before `embeddedEtcd.Close()` because the program will crash if we attempt to use `e.Etcd.Server.AuthStore().AuthEnable()` after the server is stopped:

```
"level":"fatal","caller":"backend/backend.go:637","msg":"failed to begin tx","error":"database not open"
```

test/utils/utils.go
Tomy2e: I renamed this function with your suggested name `CreateRootUserAndEnableAuth`

pkg/compactor/compactor.go
ishan16696: ok, got it ... make sense.

PR #971, opened 2026-01-22
Fix the deadlock issue in snapshotter.

### LINKED ISSUE #968: Deadlock in snapshotter
/area backup
/kind bug

**What happened**:
In one of our landscape it has been observed that backup-restore leader got stuck after closing the snapshottter and there is no log until it got backup-restore restarted at `23:50`, seems like there is a deadlock happened.
```
2026-01-09 23:50:04 | {"log":"Go OS/Arch: linux/amd64","severity":"INFO"}
2026-01-09 23:50:04 | {"log":"Go Version: go1.25.3","severity":"INFO"}
2026-01-09 23:50:04 | {"log":"Git SHA: 15c8e86","severity":"INFO"}
2026-01-09 23:50:04 | {"log":"etcd-backup-restore Version: v0.40.0","severity":"INFO"}
2026-01-09 23:49:52 | {"log":"HTTPS server closed gracefully.","severity":"INFO"}
2026-01-09 23:49:52 | {"log":"Stopping handleSsrStopRequest handler...","severity":"INFO"}
2026-01-09 23:49:52 | {"log":"handleSsrStopRequest: context canceled","severity":"INFO"}
2026-01-09 23:49:52 | {"log":"Shutting down LeaderElection...","severity":"INFO"}
2026-01-09 23:49:52 | {"log":"Stopped member lease renewal timer","severity":"INFO"}
2026-01-09 23:49:52 | {"log":"Closing defragmentor.","severity":"INFO"}
2026-01-09 14:16:48 | {"log":"Closing the full snapshot lease renewal","severity":"INFO"}
2026-01-09 14:16:48 | {"log":"Closing the Snapshotter...","severity":"INFO"}
2026-01-09 14:16:48 | {"log":"Renewed delta snapshot lease at time 2026-01-09 14:16:48.950461053 +0000 UTC m=+8660.998386745","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Stopping the FullSnapshot lease update","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Full snapshot lease's spec.renewTime is updated to the full snapshot taken time i.e 2026-01-09 14:09:10.188833408 +0000 UTC","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Renewed full snapshot lease with revision 10057291012 at time 2026-01-09 14:16:44.310818577 +0000 UTC m=+8656.358744268","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Triggering out of schedule full snapshot...","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Resetting delta snapshot to run after 5m0s.","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Stopping delta snapshot...","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"No events received to save snapshot. Skipping delta snapshot.","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Taking delta snapshot for time: 2026-01-09 14:16:44.267396683 +0000 UTC","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Will take next full snapshot at time: 2026-01-10 01:42:00 +0000 UTC","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Resetting full snapshot to run after 11h25m15.732674421s","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Stopping full snapshot...","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Applied watch on etcd from revision: 10057291013","severity":"INFO"}
2026-01-09 14:16:44 | {"log":"Successfully saved full snapshot at: Full-00000000-10057291012-1767967750.gz","severity":"INFO"}
2026-01-09 14:16:41 | {"log":"Total time to save Full snapshot: 619.589724 seconds.","severity":"INFO"}
2026-01-09 14:16:40 | {"log":"Composite object uploaded successfully.","severity":"INFO"}
2026-01-09 14:16:40 | {"log":"All chunk uploaded successfully. Uploading composite object.","severity":"INFO"}
```

**RCA**:
It has happened because of this:
1. Assume [snapshotEventHandler](https://github.com/gardener/etcd-backup-restore/blob/38d19c64caf687a5727719607ea06f5f83d4f161/pkg/snapshot/snapshotter/snapshotter.go#L201C13-L201C33) got some error before returning the error it's will execute the defer call here: https://github.com/gardener/etcd-backup-restore/blob/38d19c64caf687a5727719607ea06f5f83d4f161/pkg/snapshot/snapshotter/snapshotter.go#L171
2. Inside `stop function`, it will close lease update handler as well as try to mark the snapshotter as `inactive` and to do that it tries to take the lock.
https://github.com/gardener/etcd-backup-restore/blob/38d19c64caf687a5727719607ea06f5f83d4f161/pkg/snapshot/snapshotter/snapshotter.go#L259
4. Now, assume a `out-of-scheduled full snapshot` is triggered at same time, then it also tries to take the lock and send the information to channel to trigger the snapshot, assume it has taken the lock first before the `stop actor` (mentioned in point 2) and then tries to send information to channel but there is no snapshot event handler is active to read that channel information, hence it will cause a dealock.
https://github.com/gardener/etcd-backup-restore/blob/38d19c64caf687a5727719607ea06f5f83d4f161/pkg/snapshot/snapshotter/snapshotter.go#L206

It can be reproduced it using above RCA.

It very rare scenario

ishan16696:
/area backup
/kind bug

This PR fixes the deadlock in snapshotter. Please check RCA of this issue #968  for more details on this.

Fixes #968 

**Release note**:

```bugfix operator
Fix the deadlock issue in snapshotter.
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -98,7 +98,7 @@ type Snapshotter struct {
 	events                       []byte
 	PrevDeltaSnapshots           brtypes.SnapList
 	lastEventRevision            int64
-	SsrState                     brtypes.SnapshotterState
+	SnapshotterStateActive       bool
 	PrevFullSnapshotSucceeded    bool
 }
 
@@ -150,7 +150,7 @@ func NewSnapshotter(logger *logrus.Entry, config *brtypes.SnapshotterConfig, sto
 		PrevSnapshot:              prevSnapshot,
 		PrevFullSnapshot:          fullSnap,
 		PrevDeltaSnapshots:        deltaSnapList,
-		SsrState:                  brtypes.SnapshotterInactive,
+		SnapshotterStateActive:    !brtypes.SnapshotterActive,
 		SsrStateMutex:             &sync.RWMutex{},
 		fullSnapshotReqCh:         make(chan bool),
 		deltaSnapshotReqCh:        make(chan struct{}),
@@ -205,7 +205,7 @@ func (ssr *Snapshotter) Run(stopCh <-chan struct{}, startWithFullSnapshot bool) 
 // trigger full snapshot externally out of regular schedule.
 func (ssr *Snapshotter) TriggerFullSnapshot(_ context.Context, isFinal bool) (*brtypes.Snapshot, error) {
 
-	if !ssr.GetSnapshotterState() {
+	if !ssr.IsSnapshotterStateActive() {
 		return nil, fmt.Errorf("snapshotter is not active")
 	}
 	ssr.logger.Info("Triggering out of schedule full snapshot...")
@@ -218,7 +218,7 @@ func (ssr *Snapshotter) TriggerFullSnapshot(_ context.Context, isFinal bool) (*b
 // trigger delta snapshot externally out of regular schedule.
 func (ssr *Snapshotter) TriggerDeltaSnapshot() (*brtypes.Snapshot, error) {
 
-	if !ssr.GetSnapshotterState() {
+	if !ssr.IsSnapshotterStateActive() {
 		return nil, fmt.Errorf("snapshotter is not active")
 	}
 	if ssr.config.DeltaSnapshotPeriod.Duration < brtypes.DeltaSnapshotIntervalThreshold {
@@ -252,21 +252,21 @@ func (ssr *Snapshotter) stop(fullSnapshotLeaseStopCh chan struct{}) {
 func (ssr *Snapshotter) SetSnapshotterInactive() {
 	ssr.SsrStateMutex.Lock()
 	defer ssr.SsrStateMutex.Unlock()
-	ssr.SsrState = brtypes.SnapshotterInactive
+	ssr.SnapshotterStateActive = (!brtypes.SnapshotterActive)
 }
 
 // SetSnapshotterActive set the snapshotter state to active.
 func (ssr *Snapshotter) SetSnapshotterActive() {
 	ssr.SsrStateMutex.Lock()
 	defer ssr.SsrStateMutex.Unlock()
-	ssr.SsrState = brtypes.SnapshotterActive
+	ssr.SnapshotterStateActive = brtypes.SnapshotterActive
 }
 
-// GetSnapshotterState returns the snapshotter state.
-func (ssr *Snapshotter) GetSnapshotterState() brtypes.SnapshotterState {
+// IsSnapshotterStateActive checks the snapshotter state active or not.
+func (ssr *Snapshotter) IsSnapshotterStateActive() bool {
 	ssr.SsrStateMutex.RLock()
 	defer ssr.SsrStateMutex.RUnlock()
-	return ssr.SsrState
+	return ssr.SnapshotterStateActive
 }
 
 func (ssr *Snapshotter) closeEtcdClient() {
--- a/pkg/types/snapshotter.go
+++ b/pkg/types/snapshotter.go
@@ -24,10 +24,8 @@ const (
 	// DefaultMaxBackups is default number of maximum backups for limit based garbage collection policy.
 	DefaultMaxBackups = 7
 
-	// SnapshotterInactive is set when the snapshotter has not started taking snapshots.
-	SnapshotterInactive SnapshotterState = false
 	// SnapshotterActive is set when the snapshotter has started taking snapshots.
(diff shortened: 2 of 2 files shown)


pkg/types/snapshotter.go
```diff
@@ -25,9 +25,9 @@ const (
 	DefaultMaxBackups = 7
 
 	// SnapshotterInactive is set when the snapshotter has not started taking snapshots.
-	SnapshotterInactive SnapshotterState = 0
+	SnapshotterInactive SnapshotterState = false
 	// SnapshotterActive is set when the snapshotter has started taking snapshots.
-	SnapshotterActive SnapshotterState = 1
+	SnapshotterActive SnapshotterState = true
```
anveshreddy18: The snapshotter state is just a boolean, but I see some complexity here w.r.t having a new type that is just an alias to bool, and doesn't serve much purpose. As per the [comment](https://github.com/gardener/etcd-backup-restore/pull/971#discussion_r2757803349) I think we can simplify it all by just removing these constants which are just a wrapper on top of `true` & `false`, I don't see much value honestly.

pkg/types/snapshotter.go
```diff
@@ -44,7 +44,7 @@ const (
 )
 
 // SnapshotterState denotes the state the snapshotter would be in.
-type SnapshotterState int
+type SnapshotterState bool
```
anveshreddy18: As per the [comment](https://github.com/gardener/etcd-backup-restore/pull/971#discussion_r2757803349) I feel this can just be removed altogether.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -93,7 +93,7 @@ type Snapshotter struct {
 	watchCh                      clientv3.WatchChan
 	etcdWatchClient              *clientv3.Watcher
 	cancelWatch                  context.CancelFunc
-	SsrStateMutex                *sync.Mutex
+	SsrStateMutex                *sync.RWMutex
 	config                       *brtypes.SnapshotterConfig
 	events                       []byte
 	PrevDeltaSnapshots           brtypes.SnapList
```
anveshreddy18, line 99: In the `Snapshotter` struct, I feel we can just have the field as `SnapshotterActive bool` instead of the current nested & complicated `SsrState brtypes.SnapshotterState`, this eliminates two things:
1. Need of having `SnapshotterState` which is just an alias for bool type.
2. The need for maintaining the exact opposite constants `SnapshotterInactive` & `SnapshotterActive`, they would've made sense if we were having different possible states, but here it's only `true` or `false`, so we don't need to maintain all these complications.

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -268,6 +262,13 @@ func (ssr *Snapshotter) SetSnapshotterActive() {
 	ssr.SsrState = brtypes.SnapshotterActive
 }
 
+// GetSnapshotterState returns the snapshotter state.
+func (ssr *Snapshotter) GetSnapshotterState() brtypes.SnapshotterState {
+	ssr.SsrStateMutex.RLock()
+	defer ssr.SsrStateMutex.RUnlock()
+	return ssr.SsrState
+}
```
anveshreddy18, line 270: In continuition to the [comment](https://github.com/gardener/etcd-backup-restore/pull/971#discussion_r2757803349), we can just call this function `IsSnapshotterActive`, two reasons for that:
1. Getters are not idiomatic in go, just a standard convention to not have `Get` as prefix in function names, ref: https://go.dev/doc/effective_go#Getters
2. It also simplifies the caller logic, currently it's:
```
if !ssr.GetSnapshotterState() {
    return nil, fmt.Errorf("snapshotter is not active")
}
```
Which is a bit confusing as it doesn't tell what exactly is expected out of the function unless one dives into it, which makes it kind of hard to read.

Instead if we had:
```
if !ssr.IsSnapshotterActive() {
    return nil, fmt.Errorf("snapshotter is not active")
}
```
Which is more readable & a better way to name the func imo.

anveshreddy18 (changes requested):
Thanks for the PR @ishan16696, I know you've fixed the previously written code, but I have couple of comments which simplies the code a lot, would like to know your thoughts on these, pls see below. Thanks

PR #985, opened 2026-02-12
Fix the memory leak issue by closing the etcd watch client.

### LINKED ISSUE #928: Memory usage of backup-restore spiked.
/area robustness
/kind enhancement

Memory optimization which can act as a safeguard in case of frequent leadership change in etcd cluster.

**Why is this needed**:

It has been observed that when etcd cluster experiences frequent leadership changes (due to n/w issue or infrastructure issue), there is a corresponding change in the backup-restore leadership as well and we have observed that this lead to spike in memory usage within backup-restore container, as shown below:

Image

Furthermore, to demonstrate that this is not a memory leak issue, please check the memory usage over a 7-day period. The memory spikes occurred only during periods when the etcd leadership was unstable.

Image

ishan16696:
/area backup
/kind bug

This PR fixes the memory leak issue by closing the etcd watch client.

Fixes #928 

please check this RCA for more details: https://github.com/gardener/etcd-backup-restore/issues/928#issuecomment-3860332706

**Release note**:

```bugfix operator
Memory leak issue has been fixed by ensuring the etcd watch client is properly closed.
```

### DIFF
--- a/pkg/snapshot/snapshotter/snapshotter.go
+++ b/pkg/snapshot/snapshotter/snapshotter.go
@@ -269,7 +269,10 @@ func (ssr *Snapshotter) IsSnapshotterStateActive() bool {
 	return ssr.SnapshotterStateActive
 }
 
+// closeEtcdClient closes the Etcd watch client.
 func (ssr *Snapshotter) closeEtcdClient() {
+	ssr.logger.Info("Closing the etcd watch client.")
+
 	if ssr.cancelWatch != nil {
 		ssr.cancelWatch()
 		ssr.cancelWatch = nil
@@ -280,7 +283,7 @@ func (ssr *Snapshotter) closeEtcdClient() {
 
 	if ssr.etcdWatchClient != nil {
 		if err := (*ssr.etcdWatchClient).Close(); err != nil {
-			ssr.logger.Warnf("Error while closing etcd watch client connection, %v", err)
+			ssr.logger.Fatalf("unable to close the etcd watch client: %v", err)
 		}
 		ssr.etcdWatchClient = nil
 	}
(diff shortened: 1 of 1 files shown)


ishan16696:
After the fix, I used the same [benchmark tool](https://github.com/etcd-io/etcd/tree/main/tools/benchmark) to put the dummy data in etcd (same as [RCA](https://github.com/gardener/etcd-backup-restore/issues/928#issuecomment-3860332706)) and monitored the heap size at various timestamps , the heap dump is look like this, as we can see everything is now in `KBs`.

Screenshot 2026-02-12 at 2 47 58 PM

pkg/snapshot/snapshotter/snapshotter.go
```diff
@@ -280,7 +283,7 @@ func (ssr *Snapshotter) closeEtcdClient() {
```
anveshreddy18, line 283: It's untouched by your PR but in this `CloseEtcdClient()` method, the condition check is not needed
Current:
```go
if ssr.watchCh != nil {
	ssr.watchCh = nil
}
```
Suggested:
```go
ssr.watchCh = nil
```

anveshreddy18 (approved):
Thanks for the PR @ishan16696 for catching the issue and fixing it.

Just have one nit which you can address if you're making any other changes, otherwise we can skip that part.

Other than that LGTM!.

pkg/snapshot/snapshotter/snapshotter.go
ishan16696: I have a plan to remove/refactor some of the legacy code which is not much of a use now, I will include this change in that follow-up PR which I will open after this PR if you're fine with it.

anveshreddy18: yeah that's fine, thanks.

CaptainIRS (approved):
@ishan16696 thanks for the PR! I've verified that the memory leak is gone (Repro code: https://github.com/CaptainIRS/etcd-backup-restore/commit/411e01b19352f31c3e1ed47970f380cf3ff3a5ef).

| Before | After |
| --- | --- |
| image | image |

PR #986, opened 2026-02-16
Upgrade etcd version to 3.5.27

Shreyas-s14:
**How to categorize this PR?**

/area control-plane
/kind task

**What this PR does / why we need it**:
This PR upgrades the etcd version from `3.4.34` to etcd `3.5.26`
**Which issue(s) this PR fixes**:
Fixes Partially [Etcd-Druid Umbrella issue](https://github.com/gardener/etcd-druid/issues/445)

**Special notes for your reviewer**:
The e2e test script currently has a custom etcd-wrapper image. Once [the upgrade PR](https://github.com/gardener/etcd-wrapper/pull/80) is merged, this will be changed to the new version of etcd-wrapper. 
**Release note**:

```other operator
Upgrade etcd version to 3.5.27
```

PR #988, opened 2026-02-19
Added the `next-cluster-version-compatible` flag for embed etcd for compaction.

Shreyas-s14:

/area backup
/kind task

This PR adds the `next-cluster-version-compatible` flag to the embedded etcd creation during the process of compaction. This is a necessity in the current version of etcd (3.4.34) since the etcd version is planned to be upgraded in #986 .
Having this PR enables the downgrade path in case the rollout of the upgrade isn't successful.

Partially Fixes [Etcd-Upgrade Epic](https://github.com/gardener/etcd-druid/issues/445)

**Release note**:

```other operator
added the next-cluster-version-compatible flag for embed etcd and helm chart
```

### DIFF
--- a/chart/etcd-backup-restore/templates/etcd-configmap.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-configmap.yaml
@@ -88,6 +88,9 @@ data:
     {{- end }}
     {{- end }}
 
+    # Enable compatibility with next etcd cluster version
+    next-cluster-version-compatible: {{ .Values.backup.nextClusterVersionCompatible }}
+
 {{- if .Values.tls.etcd }}
     client-transport-security:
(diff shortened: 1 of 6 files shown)


anveshreddy18 (changes requested):
Thanks for the PR @Shreyas-s14 , mostly looks good but there are a few places in charts & tests where this flag is missing, can we add there too.

1. In the `etcd-configmap.yaml` file in the charts
2. In the example configmap file at `example/01-etcd-config.yaml` because people might refer to this to know what fields are essential
3. In `pkg/miscellaneous/testdata/valid_config.yaml` file
4. And in `test/perf/regression/resources/etcd/configmap.yaml`

CaptainIRS (approved):

PR #991, opened 2026-02-20
Disable fsync in the embedded etcd

Tomy2e:

/area performance
/kind enhancement

This PR ~~adds the `--unsafe-no-fsync` flag to~~ disables fsync in the embedded etcd which is used for applying delta snapshots.

The goal is to drastically reduce the time needed to restore a backup with heavy delta snapshots. Despite the `unsafe` prefix in the flag, there should be no data loss when the `etcdbrctl` command ends without error. 

Here are my measures:

Backup files:

|Name|Size|
|---|---|
|Full-00000000-8391520845-1763683200|344M|
|Incr-8391520846-8391652657-1763684913|232M|

Time to restore:
|`--unsafe-no-fsync`|Duration|
|---|---|
|false|2m24.394s|
|true|0m22.227s|

Logs (some lines were stripped for better readability):

```console
$ time etcdbrctl restore --storage-provider=Local --store-container=path/to/backup
INFO[0000] Finding latest set of snapshot to recover from... 
INFO[0000] Creating temporary directory default.restoration.tmp for persisting full and delta snapshots locally.  actor=restorer
INFO[0000] Restoring from base snapshot: Full-00000000-8391520845-1763683200  actor=restorer
INFO[0000] Fetched the snapshot from the object store in 0.361176521 seconds  actor=restorer
INFO[0000] Successfully fetched and saved data of the base snapshot in 0.361176521 seconds  actor=restorer
INFO[0000] Successfully restored from base snapshot: Full-00000000-8391520845-1763683200  actor=restorer
INFO[0000] Attempting to apply 1 delta snapshots for restoration.  actor=restorer
INFO[0000] Starting an embedded etcd server...           actor=restorer
INFO[0001] Embedded server is ready to listen client at: 127.0.0.1:39021  actor=restorer
INFO[0001] Applying delta snapshots...                   actor=restorer
INFO[0001] Applying first delta snapshot Incr-8391520846-8391652657-1763684913  actor=restorer
INFO[0002] successfully read the data of delta snapshot in 0.451580816 seconds  actor=restorer
INFO[0003] Applying first delta snapshot Incr-8391520846-8391652657-1763684913  actor=restorer
INFO[0144] Successfully restored the etcd data directory. 

real	2m24.394s
user	0m40.313s
sys	0m29.337s
```

```console
$ time etcdbrctl restore --storage-provider=Local --store-container=path/to/backup --unsafe-no-fsync
INFO[0000] Finding latest set of snapshot to recover from... 
INFO[0000] Creating temporary directory default.restoration.tmp for persisting full and delta snapshots locally.  actor=restorer
INFO[0000] Restoring from base snapshot: Full-00000000-8391520845-1763683200  actor=restorer
INFO[0000] Fetched the snapshot from the object store in 0.08836313 seconds  actor=restorer
INFO[0000] Successfully fetched and saved data of the base snapshot in 0.08836313 seconds  actor=restorer
INFO[0000] Successfully restored from base snapshot: Full-00000000-8391520845-1763683200  actor=restorer
INFO[0000] Attempting to apply 1 delta snapshots for restoration.  actor=restorer
INFO[0000] Starting an embedded etcd server...           actor=restorer
INFO[0001] Embedded server is ready to listen client at: 127.0.0.1:33309  actor=restorer
INFO[0001] Applying delta snapshots...                   actor=restorer
INFO[0001] Applying first delta snapshot Incr-8391520846-8391652657-1763684913  actor=restorer
INFO[0002] successfully read the data of delta snapshot in 0.261979689 seconds  actor=restorer
INFO[0003] Applying first delta snapshot Incr-8391520846-8391652657-1763684913  actor=restorer
INFO[0022] Successfully restored the etcd data directory. 

real	0m22.227s
user	0m26.358s
sys	0m13.448s
```

~~It's currently disabled by default, should we consider enabling this by default in the future?~~

**Release note**:

```feature operator
Disable fsync in the embedded etcd during restoration to reduce restoration time.
```

### DIFF
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -235,6 +235,7 @@ func StartEmbeddedEtcd(logger *logrus.Entry, ro *brtypes.RestoreOptions) (*Embed
 	cfg.MaxTxnOps = ro.Config.MaxTxnOps
 	cfg.AutoCompactionMode = ro.Config.AutoCompactionMode
 	cfg.AutoCompactionRetention = ro.Config.AutoCompactionRetention
+	cfg.UnsafeNoFsync = true
 	cfg.Logger = "zap"
 	e, err := embed.StartEtcd(cfg)
 	if err != nil {
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -73,6 +73,7 @@ func StartEmbeddedEtcd(ctx context.Context, etcdDir string, logger *logrus.Entry
 	cfg.AdvertisePeerUrls = []url.URL{*apurl}
 	cfg.AdvertiseClientUrls = []url.URL{*acurl}
 	cfg.InitialCluster = cfg.InitialClusterFromName(cfg.Name)
+	cfg.UnsafeNoFsync = true
 	cfg.Logger = "zap"
 	cfg.AutoCompactionMode = "periodic"
 	cfg.AutoCompactionRetention = "0"


CaptainIRS:
@Tomy2e thanks for the PR! This seems to be a nice optimization to reduce the restoration/snapshot compaction times. The worst-case restoration time in a heavily loaded cluster reduces from ~45m to ~6.5m (~9x improvement). I think we should enable this by default in the embedded etcd for restoration and snapshot compaction though and maybe have a flag to disable this optimization if required (cc @ishan16696 for his thoughts).

### Analysis - performance perspective:
I simulated a typical heavily loaded cluster in our landscapes that had ~1000 delta snapshots and 40M events a day (just 1M events are loaded here since restoration will involve just applying 1M events max due to snapshot compaction):
```sh
benchmark --clients 10 --conns 10 put --val-size 2500 --key-space-size 1000000 --total 1000000 --rate 100 --endpoints=http://etcd-main-client:2379
```
The DB was populated with 4.3GB of data and the time to perform a restoration dropped from 45:18m without this optimization to 6:33m.

**Q: Does this optimization make the snapshot compaction feature redundant?** No. We have a worst-case restoration time of 6:33m because the number of events to be applied is capped to 1000000. If snapshot compaction wasn't there, we'd have to apply the full 40M events in this cluster, which would take roughly 4h, which is > 4x the restoration time we currently have.

### Analysis - safety perspective:

<details>
  <summary>Analysis of the impact of `unsafe-no-fsync` on the WAL/.snap/.snap.db/db files of etcd w.r.t restoration and snapshot compaction (after etcd 3.5.27 upgrade)</summary>

(Assisted by Claude)
 ### 1. WAL files — affected

`WAL.sync()` checks `w.unsafeNoSync` and returns early without calling `Fdatasync()` when the flag is set:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/wal/wal.go#L789-L798

This applies to all WAL operations, including:

- Regular entry saves (via `Save()`)
- Snapshot marker records — `WAL.SaveSnapshot()` calls `sync()` at line 971:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/wal/wal.go#L953-L972

- **Close** — `WAL.Close()` calls `sync()` at line 878, which still skips fdatasync:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/wal/wal.go#L867-L892

Data is flushed to the kernel buffer via `encoder.flush()` but not guaranteed on disk.

### 2. `.snap` files (raft snapshot metadata) — NOT affected

`Snapshotter.save()` writes `.snap` files via `pioutil.WriteAndSyncFile()` at line 91. This unconditionally calls `Fsync()` — the `UnsafeNoFsync` flag has no effect here:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/etcdserver/api/snap/snapshotter.go#L75-L105

### 3. `.snap.db` files (BoltDB database copies) — NOT affected, and irrelevant for single-node

`Snapshotter.SaveDBFrom()` unconditionally calls `fileutil.Fsync(f)` at line 47:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/etcdserver/api/snap/db.go#L36-L75

However, `SaveDBFrom` is only called from the raft HTTP snapshot receiver — a follower receiving a snapshot from a leader:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp/http.go#L266-L268

In our single-node embedded etcd, **this code path is never hit**, so no `.snap.db` files are created during restoration/compaction.

### 4. `db` file (the live BoltDB backend) — affected

This is the main area of impact. The backend sets `bopts.NoSync` and `bopts.NoGrowSync` from `UnsafeNoFsync`:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/mvcc/backend/backend.go#L195-L196

With `NoSync=true`, BoltDB's `writeMeta()` skips `fdatasync()` on every transaction commit — data pages and meta pages are written to the OS page cache but never explicitly persisted to disk:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/bbolt/tx.go#L555-L559

On shutdown, `backend.Close()` triggers a final `CommitAndStop()` which commits the last BoltDB transaction. Normally (`NoSync=false`) this is where the `db` file gets its final fsync via `writeMeta()` → `fdatasync()`. With `UnsafeNoFsync`, this final commit skips the fsync too:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/mvcc/backend/backend.go#L446-L450

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/mvcc/backend/backend.go#L434-L437

Note that BoltDB's `close()` itself never fsyncs regardless of the flag — it simply unmaps memory and closes the file descriptor:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/bbolt/db.go#L659-L699

So the entire shutdown durability guarantee for the `db` file rests on the `Tx.Commit()` inside `CommitAndStop()`, which is exactly what `UnsafeNoFsync` disables.

---

## Safety assessment

### Restoration

The concern: if the `db` file or WAL aren't fsync'd when the embedded etcd shuts down, could the real etcd fail to start from the resulting data directory?

**Normal operation (no machine crash):** Safe. After the embedded etcd's `close()` syscall returns, all written data is in the kernel page cache and visible to any subsequent `open()` of the same files. The real etcd will see consistent data.

**Machine crash (power loss, kernel panic) between embedded etcd shutdown and OS page flush:** The `db` file and WAL could be incomplete or corrupt. However, the data directory validator that runs on every pod startup will detect this and trigger a re-restoration from the snapshot store. The validator performs multiple checks before etcd is ever started:

1. Verifies the data directory structure exists (`member/`, `member/snap/`, `member/wal/`)
2. Opens the BoltDB `db` file read-only and reads the latest revision — a corrupt `db` file will fail here
3. Runs a full BoltDB page-level consistency check via `tx.Check()`
4. Verifies WAL integrity via `wal.Verify()`
5. Compares the etcd revision against the latest snapshot revision in the store

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/pkg/initializer/validator/datavalidator.go#L62-L82

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/pkg/initializer/validator/datavalidator.go#L191-L222

If any check fails, the initializer calls `restoreCorruptData()` which fetches the latest snapshots from the store and performs a full restoration:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/pkg/initializer/initializer.go#L106-L126

The real etcd is never started with a corrupt data directory — the worst case is re-doing the restoration on the next pod startup.

### Snapshot compaction

The concern: could the compacted full snapshot that gets uploaded to the store be corrupted or incomplete?

**No.** The compacted snapshot is taken via the etcd Maintenance `Snapshot` gRPC API **while the embedded etcd is still running**. Server-side, this calls `Backend().Snapshot()` which opens a BoltDB read-only transaction and streams the data via `WriteTo()`:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc/maintenance.go#L112-L124

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/etcd/server/v3/mvcc/backend/backend.go#L345-L353

`WriteTo()` reads data pages from the mmap'd file handle and generates fresh meta pages from the in-memory transaction state — it never reads from the physical disk:

https://github.com/gardener/etcd-backup-restore/blob/46cddae4d738e457f459066d1c43e10d5898a5f1/vendor/go.etcd.io/bbolt/tx.go#L382-L419

So `UnsafeNoFsync` has no effect on the integrity of the compacted snapshot. After the snapshot is uploaded, the embedded etcd is shut down and its data directory (`/tmp/compaction`) is deleted — nobody ever reads that `db` file from disk.

### Conclusion

Enabling `UnsafeNoFsync` on the embedded etcd is safe for both use cases. During restoration, the data directory is either consumed immediately by the real etcd (same boot cycle, kernel cache guarantees consistency) or is expendable (crash → re-restore from snapshots). During compaction, the output snapshot is read from in-memory state, not from disk, and the data directory is discarded entirely.
</details>

pkg/types/restorer.go
```diff
@@ -105,6 +107,7 @@ func (c *RestorationConfig) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.AutoCompactionMode, "auto-compaction-mode", c.AutoCompactionMode, "mode for auto-compaction: 'periodic' for duration based retention. 'revision' for revision number based retention.")
 	fs.StringVar(&c.AutoCompactionRetention, "auto-compaction-retention", c.AutoCompactionRetention, "Auto-compaction retention length.")
 	fs.BoolVar(&c.NextClusterVersionCompatible, "next-cluster-version-compatible", c.NextClusterVersionCompatible, "enable compatibility with next etcd cluster version (e.g., allow etcd 3.4 to open data from etcd 3.5)")
+	fs.BoolVar(&c.UnsafeNoFsync, "unsafe-no-fsync", c.UnsafeNoFsync, "disables all uses of fsync in the embedded etcd used for applying delta snapshots")
```
ishan16696: why we want to make it configurable if we know that this will improve the performance of restoration ?

ishan16696:
@Tomy2e Please rebase this PR on latest master and resolve the merge conflicts .

ishan16696:
Can also update this config in this function as well.
https://github.com/gardener/etcd-backup-restore/blob/61226be54848068aa05ecc195224438f5be03dd2/test/utils/utils.go#L41

Tomy2e:
Thanks for your reviews, I have updated my PR to always disable fsync in the embedded etcd. I have also removed the flag to configure that, thank you @CaptainIRS for confirming that there is no bad side effect that prevents enabling that by default.

PR #992, opened 2026-02-20
Populate OriginalClusterSize in initialize command

Tomy2e:

/area backup
/kind bug

This fixes a bug that occurred in these conditions:

- Etcd cluster with 3 members
- Running `etcdbrctl initialize` on a a member with lost/corrupt data dir

The bug is that the member with lost/corrupt data dir will start as a single member after running `etcdbrctl initialize`, instead of rejoining the existing cluster.

The root cause is that [EtcdInitializer.restoreInMultiNode()](https://github.com/gardener/etcd-backup-restore/blob/v0.41.0/pkg/initializer/initializer.go#L282) is not called because `OriginalClusterSize` is always set to 0.

This PR fixes that bug by properly setting `OriginalClusterSize` based on the `etcd.conf.yaml` file, replicating what is already done in the `etcdbrctl server` command.

This is a breaking change, `etcdbrctl initialize` will now crash if `etcd.conf.yaml` is missing:

```console
$ etcdbrctl initialize
FATA[0000] failed to read etcd config file: unable to read etcd config file at path: /var/etcd/config/etcd.conf.yaml : open /var/etcd/config/etcd.conf.yaml: no such file or directory  configFile=/var/etcd/config/etcd.conf.yaml
exit status 1
```

**Release note**:

```breaking operator
Populate OriginalClusterSize in initialize command
```

### DIFF
--- a/cmd/initializer.go
+++ b/cmd/initializer.go
@@ -6,9 +6,11 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/gardener/etcd-backup-restore/pkg/initializer"
 	"github.com/gardener/etcd-backup-restore/pkg/initializer/validator"
+	"github.com/gardener/etcd-backup-restore/pkg/miscellaneous"
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


ishan16696:
Hi @Tomy2e ,

Can you please open an issue first, describing what kind of problem you're currently facing and what kind of setup you have for consumption of backup-restore, by this way I can help you better.
Anyway, from this PR description I'm assuming you're trying to do the restoration in a multi-node setup and you don't seems to be consuming backup-restore via etcd-operator ([etcd-druid](https://github.com/gardener/etcd-druid)).  Now, I think you wants to do the single member restoration (1 member went down, 2/3 up) in a multi-node setup. I hope I am right till now.

Now, Let me give you some background: Basically backup-restore never meant to work standalone in a HA setup, as it requires to be kubernetes aware and config-map template need to be provided by someone (etcd-druid in this case), and scale-up feature (1->3 replicase) also need etcd-druid to update the config-map template and so on.


Now, coming back to this PR, I'm not very comfortable with this change because now this `initialize` sub-command will have to depend on config-map template (provided by whom ?) and moreover it will try to create [k8s clientSet](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/initializer/initializer.go#L52) and then it will also do lot checks to detect various scenarios such as scale-up case etc.
https://github.com/gardener/etcd-backup-restore/blob/36243e6fca29ac92342cd5ce92f67462b6e57c06/pkg/initializer/initializer.go#L60-L77

Can you open an issue first, we can discuss it there and we can see what we can do, if it's valid case then I would happy to include that change.

Tomy2e:
Hello @ishan16696, thank you for your analysis of this PR.

For the context, I'm trying to integrate backup-restore into a fork of the official [etcd-operator](https://github.com/etcd-io/etcd-operator).

I decided to switch to the `server` subcommand which seems to be better suited for using backup-restore in Kubernetes.

I'm closing this PR as it's no longer needed. Have a nice day!

PR #1000, opened 2026-02-27
Remove deprecated support for passing endpoint via credential secrets

### LINKED ISSUE #970: Remove support to pass backup bucket endpoint information through the credential file(s)

/area backup robustness
/kind cleanup 

Remove support in passing backup bucket endpoint information through the credential file(s).
This applies to all providers that currently utilize `--store-endpoint-override`.

**Why is this needed**:

#952 introduced the `--store-endpoint-override` flag which can be used to override the endpoint at which the storage provider's bucket is hosted. This information was originally passed as a field in the credential file.

Since the credential file should not contain such information (#943), support for this should be removed in `v0.42.0` release of etcd-backup-restore.

Support for this was kept till `v0.42.0` to ensure backwards compatibility after the first release which includes #952, and provides enough time to consumers to adapt their credential files.

anveshreddy18:

/area storage
/kind cleanup

Prior to this change, storage endpoint configuration could be passed through credential secrets (endpoint for S3, domain for Azure, storageAPIEndpoint for GCS). This was deprecated in PR #952  which introduced the `--store-endpoint-override` CLI flag as the recommended approach.

This PR removes the backward compatibility code that was kept for migration purposes:

- S3: Remove `Endpoint` field from `awsCredentials` struct and related handling in `awsCredentialsFromConfig()` and `readAWSCredentialFromDir()`
- Azure: Remove `Domain` field from `absCredentials` struct and related handling in `NewABSSnapStore()` and `readABSCredentialFiles()`
- GCS: Remove `fileNameStorageAPIEndpoint` constant and `getGCSStorageAPIEndpointFromFile()` function
- Chart: Remove deprecated fields from secret template and values.yaml (endpoint, domain, storageAPIEndpoint, emulatorEnabled)

Users must now use the `--store-endpoint-override` flag for custom endpoints (emulators, regional endpoints, S3-compatible providers).

Fixes #970 

**Release note**:

```breaking user
BREAKING: Removed deprecated support for passing storage endpoint via credential secrets for AWS, GCP & Azure. Users must now use the [--store-endpoint-override](https://github.com/gardener/etcd-backup-restore/blob/36243e6fca29ac92342cd5ce92f67462b6e57c06/docs/deployment/getting_started.md?plain=1#L21) flag for custom endpoints (emulators, S3-compatible providers, regional endpoints). The following credential secret fields are no longer supported: `endpoint` (S3), `domain`/`emulatorEnabled` (Azure), `storageAPIEndpoint`/`emulatorEnabled` (GCS).
```

### DIFF
--- a/chart/etcd-backup-restore/templates/etcd-backup-secret.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-backup-secret.yaml
@@ -21,30 +21,15 @@ data:
   {{- if .Values.backup.s3.s3ForcePathStyle }}
   s3ForcePathStyle: {{ .Values.backup.s3.s3ForcePathStyle | b64enc}}
   {{- end }}
-  {{- if .Values.backup.s3.endpoint }}
-  endpoint: {{ .Values.backup.s3.endpoint | b64enc }}
-  {{- end }}
 {{- else if eq .Values.backup.storageProvider "ABS" }}
   storageAccount: {{ .Values.backup.abs.storageAccount | b64enc }}
   storageKey : {{ .Values.backup.abs.storageKey | b64enc }}
-  {{- if .Values.backup.abs.emulatorEnabled }}
-  emulatorEnabled: {{ .Values.backup.abs.emulatorEnabled | b64enc}}
-  {{- end }}
-  {{- if .Values.backup.abs.domain }}
-  domain: {{ .Values.backup.abs.domain | b64enc}}
-  {{- end }}
 {{- else if eq .Values.backup.storageProvider "GCS" }}
   {{- if .Values.backup.tokenAuth.enabled }}
   credentialsConfig: {{ include "gcs.credentialsConfig" . | fromYaml | toJson | b64enc }}
   projectID: {{ .Values.backup.gcs.projectID }}
   {{- else }}
   serviceaccount.json : {{ .Values.backup.gcs.serviceAccountJson | b64enc }}
-  {{- end }}
-  {{- if .Values.backup.gcs.storageAPIEndpoint }}
-  storageAPIEndpoint: {{ .Values.backup.gcs.storageAPIEndpoint | b64enc}}
-  {{- end }}
-  {{- if .Values.backup.gcs.emulatorEnabled }}
-  emulatorEnabled: {{ .Values.backup.gcs.emulatorEnabled | b64enc}}
   {{- end }}
 {{- else if eq .Values.backup.storageProvider "Swift" }}
   authURL: {{ .Values.backup.swift.authURL | b64enc }}
--- a/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
+++ b/chart/etcd-backup-restore/templates/etcd-statefulset.yaml
@@ -211,14 +211,6 @@ spec:
   {{- if eq .Values.backup.storageProvider "S3" }}
         - name: "AWS_APPLICATION_CREDENTIALS"
           value: "/var/etcd-backup"
-    {{- if .Values.backup.s3.endpoint }}
-        - name: "AWS_ENDPOINT_URL_S3"
-          valueFrom:
-            secretKeyRef:
-              name: {{ .Release.Name }}-etcd-backup
-              key: "endpoint"
-              optional: true
-    {{- end }}
   {{- else if eq .Values.backup.storageProvider "ABS" }}
         - name: "AZURE_APPLICATION_CREDENTIALS"
           value: "/var/etcd-backup"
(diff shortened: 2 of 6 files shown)


ishan16696:
We have introduced the [--store-endpoint-override](https://github.com/gardener/etcd-backup-restore/pull/952) in `v0.41.0`, so do we want to drop this feature of passing the endpoint information via the credential file in this release i.e `v0.43.0` or atleast wait for couple more releases (may be next `0.44.0`) as it's breaking change ?
IMO, we can include this PR in next release `v0.44.0` as general rule of thumb, we can drop the support for passing endpoint via credential file after 2 releases of deprecating it.
wdyt @anveshreddy18 @renormalize ?

anveshreddy18:
@ishan16696 sure that is fine by me, the reason for raising this PR early is that the deprecation `TODO` comments mentioned the version `0.42.0` for removal, we thought maybe it was discussed -- but yeah, we can absolutely wait and include it in `0.44.0` release.

If you've already taken a look at the PR and if it looks good, maybe you can approve it and add a `do-not-merge` label to avoid duplicate effort later. Wdyt?

ishan16696:
> If you've already taken a look at the PR and if it looks good, maybe you can approve it and add a do-not-merge label to avoid duplicate effort later. Wdyt?

I had taken a look but I guess you still have to rebase the PR after `v0.43.0` release just to be on safer side.

/hold

ishan16696:
overall look good to me, just have one question:
what about this endpoint used in openshift-container: https://github.com/gardener/etcd-backup-restore/blob/master/example/storage-provider-secrets/05-openshift-container-storage-secret.yaml#L10 .... this remains the same ?
cc @renormalize

CaptainIRS (changes requested):
@anveshreddy18 could we remove references to `AWS_ENDPOINT_URL_S3` in the codebase since we don't populate the endpoint key of the backup secret for S3 anymore?

renormalize (commented):
https://github.com/gardener/etcd-backup-restore/pull/1000#issuecomment-5233181618 @ishan16696 raises a good point.

The `generic_s3_snapstore.go` actually does not actually use `config.EndpointOverride` as is done in `NewS3SnapStore`. So all endpoint handling for any provider that depends on the generic provider can only be done through passing the field in the secret. This was a miss in #952.

The generic S3 snapstore must be modified so it can be passed the override endpoint which it will make use of to create the correct S3 client.

anveshreddy18:
Thanks for confirming @renormalize , in that case it looks like we need to open a separate PR to deprecate the endpoint options for ECS/OCS to get them to use the common `--store-endpoint-override` flag and that would require giving users the time to migrate slowly like how we did with the s3/gcs/abs storage.

I feel the fix for that will come in a separate release then? wdyt?
cc: @ishan16696

ishan16696:
yeah, may be we can do it in a separate PR, @renormalize wdyt ?
please open a new issue to track it.

CaptainIRS (changes requested):
As a follow-up to https://github.com/gardener/etcd-backup-restore/pull/1000#pullrequestreview-4894937192 could you please remove the redundant references to `AWS_ENDPOINT_URL_S3` in `aws_config.sh` and `run-e2e-test.sh`? Other than that, looks good to me.

anveshreddy18:
> As a follow-up to [#1000 (review)](https://github.com/gardener/etcd-backup-restore/pull/1000#pullrequestreview-4894937192) could you please remove the redundant references to `AWS_ENDPOINT_URL_S3` in `aws_config.sh` and `run-e2e-test.sh`? Other than that, looks good to me.


@CaptainIRS Those two fields are needed for local testing with `Localstack` since the aws sdk reads this particular `AWS_ENDPOINT_URL_S3` to connect to LocalStack.

Note: I also wanted to test this but didn't have the localstack license, so I instead removed those and pushed the changes => the e2e prow resulted in errors as expected as it was trying to connect to the real provider mistakenly as the external endpoint ref was removed, I then reverted those changes.

CaptainIRS (approved):
Ah okay, I wasn't aware of that. Then, LGTM, thanks for the PR!

anveshreddy18:
> yeah, may be we can do it in a separate PR, @renormalize wdyt ? please open a new issue to track it.

@ishan16696 @renormalize I've raised an issue #1056 to track the changes needed for other providers.

can you approve this if it looks good. Thanks

ishan16696 (approved):
thanks @anveshreddy18 for the PR. LGTM!!

renormalize (approved):
@anveshreddy18 @ishan16696 yes, let's add support to OCS/ECS in a different PR.

Let's stick to this PR for now. Thanks!

PR #1001, opened 2026-02-27
Fetch the latest 4 full snapshots when checking if a `final` snapshot is present during a `copy` operation

plkokanov:

/area control-plane-migration
/kind enhancement

This PR changes the backup `copy` operation to get the last 4 full snapshots and check whether one of them is also a `final` snapshot, instead of only getting the very last full snapshot.

This is necessary because during the `migrate` phase of control plane migration, a race could happen when the `final` snapshot is taken. There is a chance that a full snapshot is taken when `etcd-backup-resotre` is started up (e.g. if etcd is scaled up because the shoot was hibernated), or a full snapshot is taken when the `etcd-backup-restore` is being shut down. The operations to take these snapshots could take longer than the one for the `final` snapshot.
This could lead to the following logs during the `copy` operation:
```
time="2026-02-19T15:10:35Z" level=info msg="Waiting for final full snapshot..." actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Timed out waiting for final full snapshot" actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Copying backups ..." actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Getting source backups..." actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Getting destination snapshots..." actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Skipping Full snapshot Full-00000000-00000222-1769026200.gz as it already exists" actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Skipping Full snapshot Full-00000000-00000222-1769026191.gz.final as it already exists" actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Skipping Full snapshot Full-00000000-00000001-1769026109.gz as it already exists" actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Backups copied" actor=copier
time="2026-02-19T15:15:35Z" level=info msg="Shutting down..."
```

Notice that the `final` full snapshot is present and even already copied: `Skipping Full snapshot Full-00000000-00000222-1769026191.gz.final as it already exists`. However, the `copy` operation times out while trying to find out whether the `final` snapshot exists: `Timed out waiting for final full snapshot` 

We chose to take the last 4 full snapshots to be "certain" that one of them is a `final` snapshot, if it was actually taken.

Fixes #

**Release note**:

```other operator
The `copy` operation now fetches the latest 4 full snapshots when checking if a `final` full snapshot is present.
```

### DIFF
--- a/pkg/miscellaneous/miscellaneous.go
+++ b/pkg/miscellaneous/miscellaneous.go
@@ -78,10 +78,9 @@ func GetNLatestFullSnapshots(store brtypes.SnapStore, n int) (brtypes.SnapList, 
 
 	var fullSnapshotList brtypes.SnapList
 	for index := len(snapList); index > 0 && len(fullSnapshotList) < n; index-- {
-		if snapList[index-1].Kind != brtypes.SnapshotKindFull {
-			continue
+		if snapList[index-1].Kind == brtypes.SnapshotKindFull {
+			fullSnapshotList = append(fullSnapshotList, snapList[index-1])
 		}
-		fullSnapshotList = append(fullSnapshotList, snapList[index-1])
 	}
 
 	return fullSnapshotList, nil


pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,37 +64,55 @@ type advertiseURLsConfig struct {
 	InitialAdvertisePeerURLs map[string][]string `json:"initial-advertise-peer-urls"`
 }
 
-// GetLatestFullSnapshotAndDeltaSnapList returns the latest snapshot
-func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Snapshot, brtypes.SnapList, error) {
+// GetNLatestFullSnapshotsAndDeltaSnapList returns the latest N full snapshots and delta snapshots from the store.
+func GetNLatestFullSnapshotsAndDeltaSnapList(store brtypes.SnapStore, n int) (brtypes.SnapList, brtypes.SnapList, error) {
+	if n <= 0 {
+		return nil, nil, nil
+	}
```
ishan16696: I think we should return error here
```suggestion
	if n <= 0 {
		return nil, nil, fmt.Errorf("...")
	}
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,37 +64,55 @@ type advertiseURLsConfig struct {
...
 		deltaSnapList = append(deltaSnapList, snapList[index-1])
 	}
 
 	sort.Sort(deltaSnapList) // ensures that the delta snapshot list is well formed
+	return fullSnapshotList, deltaSnapList, nil
+}
+
+// GetLatestFullSnapshotAndDeltaSnapList returns the latest snapshot.
+func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Snapshot, brtypes.SnapList, error) {
+	fullSnapList, deltaSnapList, err := GetNLatestFullSnapshotsAndDeltaSnapList(store, 1)
```
ishan16696: we are not handling the error return here

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,37 +64,55 @@ type advertiseURLsConfig struct {
...
+
+	var fullSnapShot *brtypes.Snapshot
+	if len(fullSnapList) > 0 {
+		fullSnapShot = fullSnapList[0]
+	}
```
ishan16696, line 109: Isn't it's better to return the error if `GetNLatestFullSnapshotsAndDeltaSnapList` return empty fullSnapshot List ?

ishan16696 (changes requested):
few nits

pkg/miscellaneous/miscellaneous.go
plkokanov: Nice catch! I added the code for the metrics and forgot to move the error handling before it.

pkg/miscellaneous/miscellaneous.go
plkokanov: I think not returning an error keeps the old behaviour, e.g. if there were no full snapshots.

pkg/miscellaneous/miscellaneous.go
plkokanov: Do you mean only if n is less than 0, or also return an error if n is 0?

ishan16696: it depends how we want to interpretate if `n==0`.
I'm thinking like this :
```go
if n<0 {
  return the err
}

if n==0 {
  // it means caller don't want fullSnapshotList
  // may be print a warning 🤷 
}

```
wdyt ?

pkg/miscellaneous/miscellaneous.go
ishan16696: ok, fine ... no strong opinion on this.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,37 +64,63 @@ type advertiseURLsConfig struct {
...
-	for index := len(snapList); index > 0; index-- {
+	for index := len(snapList); index > 0 && len(fullSnapshotList) < n; index-- {
 		if snapList[index-1].IsChunk {
 			continue
 		}
 		if snapList[index-1].Kind == brtypes.SnapshotKindFull {
-			fullSnapshot = snapList[index-1]
-			break
+			fullSnapshotList = append(fullSnapshotList, snapList[index-1])
+			continue
```
ishan16696: this logic doesn't seems to be correct to me, the condition: `index > 0 && len(fullSnapshotList) < n` doesn't seems to be correct as the loop is breaking if `len(fullSnapshotList) > n` becomes true, so it might possible we might to not appending all the delta snapshot in its list.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,37 +64,63 @@ type advertiseURLsConfig struct {
...
-func GetLatestFullSnapshotAndDeltaSnapList(store brtypes.SnapStore) (*brtypes.Snapshot, brtypes.SnapList, error) {
+// GetNLatestFullSnapshotsAndDeltaSnapList returns the latest N full snapshots and delta snapshots from the store.
+// N must be non-negative.
+func GetNLatestFullSnapshotsAndDeltaSnapList(log *logrus.Entry, store brtypes.SnapStore, n int) (brtypes.SnapList, brtypes.SnapList, error) {
+	if n < 0 {
+		return nil, nil, fmt.Errorf("invalid value for n: %d, n must be non-negative", n)
+	}
+	if n == 0 {
+		log.Warn("n is 0, returning empty snapshot lists")
+		return nil, nil, nil
```
ishan16696: function name suggest `n` is for LatestFullSnapshot not for delta snapshot but here we are returning empty delta snapshot list as well.
should we simply return error for this well instead of complicating this ?

pkg/miscellaneous/miscellaneous.go
ishan16696: IMO, there should be 2 loops one for collecting the all deltaSnapshot and one for `n` full snapshot. wdyt ?

plkokanov: > this logic doesn't seems to be correct to me, the condition: index > 0 && len(fullSnapshotList) < n doesn't seems to be correct as the loop is breaking if len(fullSnapshotList) > n becomes true, so it might possible we might to not appending all the delta snapshot in its list.

Isn't this exactly the same as what was happening before, where we break on the first full snapshot that we find?

Anyway, after thinking more on this, maybe it's better to just introduce a new function `GetNLatestFullSnapshots` and not change the already existing. Wdyt?

ishan16696: >  Isn't this exactly the same as what was happening before, where we break on the first full snapshot that we find?

yes, I think what's happening is snapshot list was sorted in timestamp, we get 1 last/latest full snapshot and delta snapshots which were taken after that fullSnapshot, but after this logic change, if we try to get `n` full snapshot then we're continuing the iteration, it means we might append more delta-snapshots in deltaSnapshotList which may not be require.

> Anyway, after thinking more on this, maybe it's better to just introduce a new function GetNLatestFullSnapshots and not change the already existing. Wdyt?

yeah, let's do that, it will simplify the logic and testing will be easy.

plkokanov: > yeah, let's do that, it will simplify the logic and testing will be easy.

Done

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,7 +64,30 @@ type advertiseURLsConfig struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var fullSnapshotList brtypes.SnapList
+	for index := len(snapList); index > 0 && len(fullSnapshotList) < n; index-- {
+		if snapList[index-1].Kind != brtypes.SnapshotKindFull {
+			continue
+		}
+		fullSnapshotList = append(fullSnapshotList, snapList[index-1])
```
```suggestion
		if snapList[index-1].Kind == brtypes.SnapshotKindFull {
			fullSnapshotList = append(fullSnapshotList, snapList[index-1])
		}
```

ishan16696 (approved):

Issue #896, opened 2025-07-30
[Enhancement] Decouple the etcd member name from the pod name


acumino:
**How to categorize this issue?**

/area control-plane-migration high-availability ops-productivity performance robustness scalability usability
/kind enhancement

**What would you like to be added**:
Currently, the [etcd member name](https://github.com/gardener/etcd-backup-restore/blob/eaf511e602c6ca339974d55e1f86c1df22ac0680/pkg/server/httpAPI.go#L439) is the same as the pod name of the member.
Because of this, in scenarios like [Live CPM](https://github.com/gardener/gardener/issues/10686), where we want to have multiple etcd members with the same pod name, the new member cannot join the cluster.

Therefore, we want the etcd member name to be generated independently of the pod name.

**Why is this needed**:
The supports use cases like [Live CPM](https://github.com/gardener/gardener/issues/10686)

ghost:
@acumino No more than 5 labels permitted, but 7 labels were given.

acumino:
/area control-plane-migration ops-productivity performance robustness scalability

acumino:
cc @unmarshall @shreyas-s-rao @seshachalam-yv

ghost:
@acumino Command `/remove-lifecycle` is not known.

acumino:
@seshachalam-yv and I had a discussion about it.

We can add a new field in etcd called `memberPrefix`, and the member names of that etcd will be calculated based on the prefix.
For example, if the `memberPrefix` is `test` and etcd replicas is 3, member names will be
```
- test-0
- test-1
- test-2
```

But the pods names will still be based on the etcd cr names.

@gardener/etcd-druid-maintainers

anveshreddy18:
/assign @unmarshall @ishan16696 

As discussed in the today's Etcd sync, this is no longer a required task for self hosted control planes, can you provide the comments in the context of CPM live migration.

ghost:
@seshachalam-yv Command `/remove-lifecycle` is not known.

CaptainIRS:
@acumino, the idea of adding a `memberPrefix` to the etcds sounds good to me, but I just want to clarify one aspect of the proposal: How do we onboard existing etcd clusters which currently do not have a `memberPrefix` to these new member names? etcd doesn't provide a mechanism to update the name of the etcd dynamically, the only way to do it is to issue a `member remove` with the old member ID and then issue a `member add` with the new name.

CaptainIRS:
[...]

PR #1030, opened 2026-04-27
Adapt `etcd-backup-restore` to support `member-name-prefix` for etcd cluster members

acumino:

/area control-plane-migration delivery open-source
/kind api-change enhancement epic

Adapts etcd-backup-restore to the `member-name-prefix feature` from [etcd-druid#1309](https://github.com/gardener/etcd-druid/pull/1309). When set in the etcd config file, each member name becomes `<prefix>-<pod-name>` instead of just the pod name. Backward compatible `-` prefix is optional.

Part of https://github.com/gardener/etcd-backup-restore/issues/896 https://github.com/gardener/etcd-druid/issues/1237 https://github.com/gardener/gardener/issues/10686

**Release note**:

```feature operator
etcd member names now support an optional `member-name-prefix` field in the etcd config file. When set, member names take the form `<prefix>-<pod-name>` instead of just the pod name.

```

### DIFF
--- a/pkg/member/member_control_test.go
+++ b/pkg/member/member_control_test.go
@@ -28,9 +28,11 @@ var _ = Describe("Membercontrol", func() {
 		ctrl                 *gomock.Controller
 		factory              *mockfactory.MockFactory
 		cl                   *mockfactory.MockClusterCloser
+		memberNamePrefix     string
 	)
 
 	BeforeEach(func() {
+		memberNamePrefix = ""
 		etcdConnectionConfig = brtypes.NewEtcdConnectionConfig()
 		etcdConnectionConfig.Endpoints = []string{etcd.Clients[0].Addr().String()}
 		etcdConnectionConfig.ConnectionTimeout.Duration = 30 * time.Second
@@ -43,6 +45,15 @@ var _ = Describe("Membercontrol", func() {
 		ctrl = gomock.NewController(GinkgoT())
 		factory = mockfactory.NewMockFactory(ctrl)
 		cl = mockfactory.NewMockClusterCloser(ctrl)
+	})
+
+	JustBeforeEach(func() {
+		urlKey := podName
+		prefixLine := ""
+		if memberNamePrefix != "" {
+			urlKey = memberNamePrefix + "-" + podName
+			prefixLine = "\nmember-name-prefix: " + memberNamePrefix
+		}
 
 		outfile := "/tmp/etcd.conf.yaml"
 		etcdConfigYaml := `# Human-readable name for this member.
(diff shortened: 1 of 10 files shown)


pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,6 +64,43 @@ type advertiseURLsConfig struct {
...
+		}
+	}
+	return "", nil
+}
+
+// ComputeMemberName constructs the member name from an optional prefix and pod name.
+// If prefix is non-empty, the member name is "<prefix>-<podName>", otherwise it is just <podName>.
+func ComputeMemberName(memberNamePrefix, podName string) string {
+	if memberNamePrefix != "" {
+		return memberNamePrefix + "-" + podName
```
```suggestion
		return fmt.Sprintf("%s-%s", memberNamePrefix, podName)
```

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,6 +64,43 @@ type advertiseURLsConfig struct {
...
+	}
+	return podName
+}
+
+// GetMemberName reads the member-name-prefix from the config file and computes
```
ishan16696, line 90: I see sometimes you're using `prefix` term, sometimes you're using `member-name-prefix` term, I would suggest just use `member-name-prefix` at all places.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -64,6 +64,43 @@ type advertiseURLsConfig struct {
...
+	if prefix, ok := config["member-name-prefix"]; ok {
+		if p, ok := prefix.(string); ok {
+			return p, nil
+		}
+	}
+	return "", nil
+}
+
+// ComputeMemberName constructs the member name from an optional prefix and pod name.
+// If prefix is non-empty, the member name is "<prefix>-<podName>", otherwise it is just <podName>.
```
```suggestion
// ComputeMemberName constructs the member name from a pod name and from an optional member-name-prefix (if any).
// If member-name-prefix is non-empty, then member name would be "<memberPrefix>-<podName>", otherwise it would be just <podName>.
```

ishan16696 (changes requested):
few nits from my side, also rebase the PR on latest master.

acumino:
Can we add it to `v0.43.0`?

ishan16696:
Please update the release note with more details, may be like this:
```
etcd cluster member names can now be prefixed via optional `member-name-prefix` field in the etcd config file. When set, member names take the form `<prefix>-<pod-name>`. If not set then no effect, member name would be same i.e `<pod-name>`.
```
Update this example file: https://github.com/gardener/etcd-backup-restore/blob/master/example/01-etcd-config.yaml and also in helm charts: https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/etcd-configmap.yaml#L12

/hold

acumino:
> Please update the release note with more details, may be like this:
> 
> ```
> etcd cluster member names can now be prefixed via optional `member-name-prefix` field in the etcd config file. When set, member names take the form `<prefix>-<pod-name>`. If not set then no effect, member name would be same i.e `<pod-name>`.
> ```
> 
> Update this example file: https://github.com/gardener/etcd-backup-restore/blob/master/example/01-etcd-config.yaml and also in helm charts: https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/etcd-configmap.yaml#L12
> 
> /hold

I have updated the configs 323adcf PTAL.

pkg/member/member_control_test.go
```diff
@@ -53,21 +64,20 @@ snapshot-count: 75000
...
 initial-advertise-peer-urls:
-  ` + podName + `:
+  ` + urlKey + `:
     - http://` + etcd.Peers[0].Addr().String() + `
 initial-cluster: etcd1=http://0.0.0.0:2380
 initial-cluster-token: new
 initial-cluster-state: new
 auto-compaction-mode: periodic
-auto-compaction-retention: 30m`
+auto-compaction-retention: 30m` + prefixLine
```
ishan16696, line 76: why you've added `prefixLine` to `auto-compaction-retention` ?

ishan16696 (changes requested):
overall looks good, just nit.

pkg/member/member_control_test.go
acumino: It's not appended to `auto-compaction-retention`, but is appended to the config. Here basically `member-name-prefix` is being added at the end of the config if present.

ishan16696: ok, got it ... my bad.

ishan16696 (approved):
LGTM!!

PR #1039, opened 2026-07-15
Enhance schedule/cron defrag

ishan16696:
/area high-availability
/kind enhancement

This PR enchances the schedule defragmentation to run only if it's beneficial:
It checks whether defragmentation can be skipped for a given etcd cluster.
- Can't be skipped:  when either TotalDBSize or available free space is greater than threshold.
  1. If TotalDBSize > 3GB
  2. If FreeSpaceAvailable > 1GB (configurable via flag `--freespace-threshold` )
- Can be skipped: when the TotalDBSize and available free space are below threshold, meaning defragmentation would have little benefit.

This PR also introduce defragConfig for all defrag related configuration.

Fixes #

Please note `DefragThreshold (3GiB DB-size check)`  is still a hardcoded constant in `miscellaneous.go`, it will be addressed in this issue https://github.com/gardener/etcd-backup-restore/issues/556.

**Release note**:

```noteworthy user
Schedule/cron etcd defragmentation can be auto skipped if TotalDBSize < 3GB and FreeSpaceAvailable < 1GB (configurable via CLI flag `--freespace-threshold` )
```

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -48,13 +48,13 @@ storing snapshots on various cloud storage providers as well as local disk locat
 				logger.Fatalf("Failed to create snapshotter: %v", err)
 			}
 
-			defragSchedule, err := cron.ParseStandard(opts.defragmentationSchedule)
+			defragSchedule, err := cron.ParseStandard(opts.defragConfig.DefragmentationSchedule)
 			if err != nil {
 				logger.Fatalf("failed to parse defragmentation schedule: %v", err)
 				return
 			}
 
-			go defragmentor.DefragDataPeriodically(ctx, opts.etcdConnectionConfig, defragSchedule, ssr.TriggerFullSnapshot, logger)
+			go defragmentor.DefragDataPeriodically(ctx, opts.etcdConnectionConfig, opts.defragConfig, defragSchedule, ssr.TriggerFullSnapshot, logger)
 
 			go ssr.RunGarbageCollector(ctx.Done())
 			if err := ssr.Run(ctx.Done(), true); err != nil {
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -13,7 +13,6 @@ import (
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
 
 	"github.com/go-logr/logr"
-	cron "github.com/robfig/cron/v3"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	runtimelog "sigs.k8s.io/controller-runtime/pkg/log"
@@ -48,13 +47,7 @@ storing snapshots on various cloud storage providers as well as local disk locat
 				logger.Fatalf("Failed to create snapshotter: %v", err)
 			}
 
-			defragSchedule, err := cron.ParseStandard(opts.defragConfig.DefragmentationSchedule)
-			if err != nil {
-				logger.Fatalf("failed to parse defragmentation schedule: %v", err)
-				return
-			}
-
-			go defragmentor.DefragDataPeriodically(ctx, opts.etcdConnectionConfig, opts.defragConfig, defragSchedule, ssr.TriggerFullSnapshot, logger)
+			go defragmentor.DefragDataPeriodically(ctx, opts.etcdConnectionConfig, opts.defragConfig, ssr.TriggerFullSnapshot, logger)
 
 			go ssr.RunGarbageCollector(ctx.Done())
 			if err := ssr.Run(ctx.Done(), true); err != nil {
--- a/pkg/defragmentor/defrag.go
+++ b/pkg/defragmentor/defrag.go
@@ -37,6 +37,9 @@ func NewDefragmentorJob(ctx context.Context, etcdConnectionConfig *brtypes.EtcdC
 	}
 }
 
+// Run executes one defragmentation cycle.
+//   - First checks whether defragmentation can be skipped based on DB-size and free-space thresholds.
+//   - If not, then waits until all etcd cluster members are healthy before defragmenting.
 func (d *defragmentorJob) Run() {
 	clientFactory := etcdutil.NewFactory(*d.etcdConnectionConfig)
 
@@ -45,6 +48,14 @@ func (d *defragmentorJob) Run() {
 		d.logger.Warnf("failed to create etcd maintenance client")
 	}
 	defer clientMaintenance.Close()
+
+	canSkip, err := miscellaneous.CanDefragSkip(d.ctx, clientMaintenance, d.etcdConnectionConfig)
+	if err != nil {
+		d.logger.Warnf("failed to check etcd defrag skip conditions, proceeding with etcd defragmentation: %v", err)
+	} else if canSkip {
+		d.logger.Infof("skipping the schedule defrag for etcd cluster as Total DBSize is below %dGB threshold and available free-space is below %dGB threshold", miscellaneous.DefragThreshold/1024/1024/1024, miscellaneous.FreeSpaceThreshold/1024/1024/1024)
+		return
+	}
 
 	client, err := clientFactory.NewCluster()
 	if err != nil {
@@ -75,7 +86,7 @@ waitLoop:
 			}
 
 			if isClusterHealthy {
-				d.logger.Infof("Starting the defragmentation as all members of etcd cluster are in healthy state")
+				d.logger.Infof("Starting the schedule defragmentation as all members of etcd cluster are in healthy state")
 				err = etcdutil.DefragmentData(d.ctx, clientMaintenance, client, etcdEndpoints, d.etcdConnectionConfig.DefragTimeout.Duration, d.logger)
 				if err != nil {
 					d.logger.Warnf("failed to defrag data with error: %v", err)
--- a/pkg/defragmentor/defrag.go
+++ b/pkg/defragmentor/defrag.go
@@ -23,15 +23,17 @@ type CallbackFunc func(ctx context.Context, isFinal bool) (*brtypes.Snapshot, er
 type defragmentorJob struct {
 	ctx                  context.Context
 	etcdConnectionConfig *brtypes.EtcdConnectionConfig
+	defragConfig         *brtypes.DefragConfig
 	logger               *logrus.Entry
 	callback             CallbackFunc
 }
 
 // NewDefragmentorJob returns the new defragmentor job.
-func NewDefragmentorJob(ctx context.Context, etcdConnectionConfig *brtypes.EtcdConnectionConfig, logger *logrus.Entry, callback CallbackFunc) cron.Job {
+func NewDefragmentorJob(ctx context.Context, etcdConnectionConfig *brtypes.EtcdConnectionConfig, defragConfig *brtypes.DefragConfig, logger *logrus.Entry, callback CallbackFunc) cron.Job {
 	return &defragmentorJob{
 		ctx:                  ctx,
 		etcdConnectionConfig: etcdConnectionConfig,
+		defragConfig:         defragConfig,
 		logger:               logger.WithField("job", "defragmentor"),
 		callback:             callback,
 	}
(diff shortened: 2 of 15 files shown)


pkg/types/defragmentation.go
```diff
@@ -0,0 +1,66 @@
...
+	fs.DurationVar(&c.DefragTimeout.Duration, "etcd-defrag-timeout", c.DefragTimeout.Duration, "timeout duration for etcd defrag call")
+}
+
+// Validate validates the DefragConfig fields.
+func (c *DefragConfig) Validate() error {
+	if c.DefragTimeout.Duration <= 0 {
+		return fmt.Errorf("etcd defrag timeout should be greater than zero")
+	}
+	if _, err := cron.ParseStandard(c.DefragmentationSchedule); err != nil {
+		return err
```
anveshreddy18:
```suggestion
		return fmt.Errorf("error parsing defragmentation schedule : %w", err)
```

the `err` comes from the `func (p Parser) Parse(spec string) (Schedule, error)` from the `github.com/robfig/cron/v3/parser.go` file and that has non-descriptive errors like `empty spec string` & `provided bad location`, etc... which are not descriptive enough, the assumption I think was that the caller would handle the specific error case with more explanation. So I feel here we should add a error prefix explaining where the error is coming from, otherwise the logs wouldn't make much sense. Wdyt?

pkg/types/defragmentation.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/gardener/etcd-backup-restore/pkg/wrappers"
+	"github.com/robfig/cron/v3"
+
+	flag "github.com/spf13/pflag"
+)
+
+const (
+	// DefaultDefragSchedule is the default cron schedule for defragmentation (midnight every 3 days).
+	DefaultDefragSchedule = "0 0 */3 * *"
+	// DefaultDefragConnectionTimeout defines default timeout duration for ETCD defrag call.
```
anveshreddy18:
```suggestion
	// DefaultDefragConnectionTimeout defines the default timeout duration for a single etcd defrag call.
```

pkg/types/defragmentation.go
```diff
@@ -0,0 +1,66 @@
...
+		FreespaceThreshold:      DefaultFreeSpaceThreshold,
+		DefragTimeout:           wrappers.Duration{Duration: DefaultDefragConnectionTimeout},
+	}
+}
+
+// AddFlags adds defragmentation-related flags to the given flag set.
+func (c *DefragConfig) AddFlags(fs *flag.FlagSet) {
+	fs.Int64Var(&c.FreespaceThreshold, "freespace-threshold", c.FreespaceThreshold, "minimum free-space in database only above which schedule defragmentation is triggered.")
+	fs.StringVar(&c.DefragmentationSchedule, "defragmentation-schedule", c.DefragmentationSchedule, "cron schedule to defragment etcd data directory")
+	fs.DurationVar(&c.DefragTimeout.Duration, "etcd-defrag-timeout", c.DefragTimeout.Duration, "timeout duration for etcd defrag call")
```
anveshreddy18:
```suggestion
	fs.DurationVar(&c.DefragTimeout.Duration, "etcd-defrag-timeout", c.DefragTimeout.Duration, "timeout duration for a single etcd defrag call, incase of multiple etcd instances configure this accordingly")
```

Since the text here is the only actual explanation of the API surface from the backup-restore side, I think we need to explicitly mention that the timeout is for a single member and for multiple members, it needs to be scaled accordingly.

pkg/defragmentor/defrag.go
```diff
@@ -75,8 +88,8 @@ waitLoop:
 			}
 
 			if isClusterHealthy {
-				d.logger.Infof("Starting the defragmentation as all members of etcd cluster are in healthy state")
-				err = etcdutil.DefragmentData(d.ctx, clientMaintenance, client, etcdEndpoints, d.etcdConnectionConfig.DefragTimeout.Duration, d.logger)
+				d.logger.Infof("Starting the schedule defragmentation as all members of etcd cluster are in healthy state")
```
anveshreddy18:
```suggestion
				d.logger.Infof("Starting the scheduled defragmentation since all members of the etcd cluster are in a healthy state")
```
nit

cmd/snapshot.go
```diff
@@ -48,13 +48,13 @@ storing snapshots on various cloud storage providers as well as local disk locat
...
 
-			defragSchedule, err := cron.ParseStandard(opts.defragmentationSchedule)
+			defragSchedule, err := cron.ParseStandard(opts.defragConfig.DefragmentationSchedule)
 			if err != nil {
 				logger.Fatalf("failed to parse defragmentation schedule: %v", err)
 				return
 			}
 
-			go defragmentor.DefragDataPeriodically(ctx, opts.etcdConnectionConfig, defragSchedule, ssr.TriggerFullSnapshot, logger)
+			go defragmentor.DefragDataPeriodically(ctx, opts.etcdConnectionConfig, opts.defragConfig, defragSchedule, ssr.TriggerFullSnapshot, logger)
```
anveshreddy18: Now that we're passing the entire defragConfig to the function, this `defragSchedule` need not be computed here and can instead be moved inside the `DefragDataPeriodically` func because of two reasons:
1. the func `DefragDataPeriodically` now accepts the `defragConfig` param which contains the schedule, now we're duplicating information by passing `defragSchedule` also.
2. If we place it in the `DefragDataPeriodically`, it follows the single responsibiity principle as it is this func's responsibility to compute the cron.Schedule object out of the cron expression, we can remove duplicate computation that is happening before every call to this function.

pkg/defragmentor/defrag.go
```diff
@@ -93,8 +106,8 @@ waitLoop:
 }
 
 // DefragDataPeriodically defragments the data directory of each etcd member.
-func DefragDataPeriodically(ctx context.Context, etcdConnectionConfig *brtypes.EtcdConnectionConfig, defragmentationSchedule cron.Schedule, callback CallbackFunc, logger *logrus.Entry) {
-	defragmentorJob := NewDefragmentorJob(ctx, etcdConnectionConfig, logger, callback)
+func DefragDataPeriodically(ctx context.Context, etcdConnectionConfig *brtypes.EtcdConnectionConfig, defragConfig *brtypes.DefragConfig, defragmentationSchedule cron.Schedule, callback CallbackFunc, logger *logrus.Entry) {
+	defragmentorJob := NewDefragmentorJob(ctx, etcdConnectionConfig, defragConfig, logger, callback)
 	// TODO: Sync logrus logger to cron logger
 	jobRunner := cron.New(cron.WithChain(cron.SkipIfStillRunning(cron.DefaultLogger)))
 	jobRunner.Schedule(defragmentationSchedule, defragmentorJob)
```
anveshreddy18:
```suggestion
func DefragDataPeriodically(ctx context.Context, etcdConnectionConfig *brtypes.EtcdConnectionConfig, defragConfig *brtypes.DefragConfig, callback CallbackFunc, logger *logrus.Entry) {
	defragmentorJob := NewDefragmentorJob(ctx, etcdConnectionConfig, defragConfig, logger, callback)
    defragSchedule, err := cron.ParseStandard(defragConfig.DefragmentationSchedule)
    if err != nil {
        logger.Fatalf("failed to parse defragmentation schedule: %v", err)
        return
    }
	// TODO: Sync logrus logger to cron logger
	jobRunner := cron.New(cron.WithChain(cron.SkipIfStillRunning(cron.DefaultLogger)))
	jobRunner.Schedule(defragSchedule, defragmentorJob)
```

As mentioned in the previous [comment](https://github.com/gardener/etcd-backup-restore/pull/1039#discussion_r3657030526), currently we're having a duplicate input parameter which is not required as the defragmentation schedule can be computed inside this `DefragDataPeriodically` itself using the new parameter `defragConfig`.

pkg/defragmentor/defrag.go
```diff
@@ -46,6 +51,14 @@ func (d *defragmentorJob) Run() {
 	}
 	defer clientMaintenance.Close()
 
+	canSkip, err := miscellaneous.CanDefragSkip(d.ctx, clientMaintenance, d.etcdConnectionConfig, d.defragConfig)
```
anveshreddy18: I took a look into `miscellaneous` pkg and found that most of the helper functions in that are supposed to be used by different pkgs so that we don't duplicate them everywhere.

But the `CanDefragSkip` is just specific to defragmentation, I think the right place for this function is in the current `defragmentor` pkg itself, maybe you can create a `helper.go` file for this as it's only helping the defragmentor pkg.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -820,3 +823,27 @@ func getEtcdWrapperEndpoint(etcdEndpoints []string) (string, error) {
 
 	return fmt.Sprintf("%s://%s:%s", etcdURL.Scheme, etcdURL.Hostname(), etcdWrapperPort), nil
 }
+
+// CanDefragSkip checks whether defragmentation can be skipped for a given etcd cluster.
+// It returns false (do not skip) when either TotalDBSize or available free space is greater than threshold.
+// It returns true (can skip) when the TotalDBSize and available free space are below threshold, meaning defragmentation would have
+// little benefit.
+func CanDefragSkip(pCtx context.Context, client etcdClient.MaintenanceCloser, etcdConnectionConfig *brtypes.EtcdConnectionConfig, defragCfg *brtypes.DefragConfig) (bool, error) {
```
anveshreddy18:
```suggestion
func CanSkipDefrag(pCtx context.Context, client etcdClient.MaintenanceCloser, etcdConnectionConfig *brtypes.EtcdConnectionConfig, defragCfg *brtypes.DefragConfig) (bool, error) {
```

Nit: just so that it sounds right.

anveshreddy18 (changes requested):
Thanks for the PR, PTAL at my comments.

pkg/types/defragmentation.go
ishan16696: I understand what you are trying to convey but it's hard to explain in 1 line for someone who doesn't understand how defrag works in etcd.
I can reworded to this, but still it won't achieve everything what you are trying to convey

```suggestion
	// DefaultDefragConnectionTimeout defines the default timeout duration for etcd's single defrag api call.
```

pkg/types/defragmentation.go
ishan16696: same reason as above 
```suggestion
	fs.DurationVar(&c.DefragTimeout.Duration, "etcd-defrag-timeout", c.DefragTimeout.Duration, "timeout duration for etcd's one defrag api call")
```

ishan16696: > incase of multiple etcd instances configure this accordingly

what do you mean by this, defrag run in rolling fashion, timeout is for one api call

pkg/defragmentor/defrag.go
ishan16696: ok, no strong opinion on this, we can do that as well.

ishan16696: I have taken look on this and as I mentioned I have no objection in moving the function to different place but the problem currently many other function related to defrag are available in this pkg: `pkg/etcdutil` example:
https://github.com/gardener/etcd-backup-restore/blob/debd95bd77352280dada2a9a94efa9d1d20bbb78/pkg/etcdutil/etcdutil.go#L189

https://github.com/gardener/etcd-backup-restore/blob/debd95bd77352280dada2a9a94efa9d1d20bbb78/pkg/etcdutil/etcdutil.go#L157

So, either I can move this func `CanDefragSkip` to same pkg/etcdutil or I will move all functions related to defrag to this new file . wdyt ?

cmd/snapshot.go
ishan16696: yes, thanks for pointing this out ... I missed this.

pkg/defragmentor/defrag.go
ishan16696: correct ... made the changes. PTAL.

pkg/types/defragmentation.go
ishan16696: ok, made the changes.

pkg/defragmentor/defrag.go
ishan16696: As discussed I have moved the func `CanDefragSkip` to same pkg/etcdutil. PTAL

pkg/types/defragmentation.go
anveshreddy18: the above wording looks fine to me.

pkg/types/defragmentation.go
anveshreddy18: as discussed... while there is some difference in how the API sounds and the internal implementation as the internal implementation uses the mentioned timeout for a single call to one member, but previously the API sounded like the timeout is for the entire defragmentation operation which is not the case.

But after the comment has been enhanced to mention that the timeout is for a single member, it looks fine to me.

anveshreddy18:
@ishan16696  Looks good but just few nit changes:

The const `defaultDefragmentationSchedule` is now orphaned and can be removed.

https://github.com/ishan16696/etcd-backup-restore/blob/5ea72103a8b8af263567d9435e77cd2649d79b67/pkg/server/types.go#L14

Also the log line https://github.com/ishan16696/etcd-backup-restore/blob/5ea72103a8b8af263567d9435e77cd2649d79b67/pkg/defragmentor/defrag.go#L58 is doing integer division now, so even if we configures the threshold to be `3.25 GB`, then it still shows it as `3GB` and `500MB` is shown as `0GB`. Can we make it a floating point division to be accurate.

ishan16696:
> Also the log line https://github.com/ishan16696/etcd-backup-restore/blob/5ea72103a8b8af263567d9435e77cd2649d79b67/pkg/defragmentor/defrag.go#L58 is doing integer division now, so even if we configures the threshold to be 3.25 GB, then it still shows it as 3GB and 500MB is shown as 0GB. Can we make it a floating point division to be accurate.

yes, I also found it while testing, fixed it now. PTAL.

anveshreddy18 (approved):

anveshreddy18 (approved):
Thanks for making the changes. LGTM!

PR #1040, opened 2026-07-21
Remove StatefulSet-based scale-out detection from initial-cluster-state determination

### LINKED ISSUE #1037: Remove StatefulSet-based scale-out detection from initial-cluster-state determination

/area high-availability
/kind cleanup

Remove the StatefulSet status-based scale-out detection from `/config` `initial-cluster-state` determination.

Specifically:
- Remove `GetInitialClusterStateIfScaleup` from the `getClusterState` decision path.
- Remove the StatefulSet `spec.replicas > status.updatedReplicas` check used to infer scale-out.
- For multi-node clusters, determine `initial-cluster-state` only from learner presence:
  - learner present: `existing`
  - no learner or learner check error: `new`
- Keep single-node clusters returning `new`.

**Why is this needed**:

`/config` is requested only after backup-restore initialization completes, so members that must join an existing cluster should already have been added as learners. StatefulSet status is an indirect rollout signal and can be misleading during bootstrap, unhibernation, or status lag, while learner presence reflects actual etcd membership state.

priyadharshini-151193:
/area high-availability
/kind cleanup

Remove the StatefulSet status-based scale-out detection from /config initial-cluster-state determination.

Specifically:

Remove GetInitialClusterStateIfScaleup from the getClusterState decision path.
Remove the StatefulSet spec.replicas > status.updatedReplicas check used to infer scale-out.
For multi-node clusters, determine initial-cluster-state only from learner presence:
learner present: existing
no learner or learner check error: new
Keep single-node clusters returning new.

Fixes #1037

**Release note**:

```noteworthy user
Remove StatefulSet-based scale-out detection from etcd's initial-cluster-state determination; cluster state for multi-node etcd is derived solely from etcd learner membership. Respective RBACs has also been removed from helm charts.
```


priyadharshini-151193:
Testing Done : 

1. ETCD Cluster was created with one member. The cluster state was reported as "New".
2. The cluster was scaled out to 3. The new members had the state reported as "Existing". 


```
 k logs -n default etcd-test-2  -c backup-restore | grep -i state
- time="2026-07-20T06:03:01Z" level=info msg="The cluster state is existing" actor=backup-restore-server
- k logs -n default etcd-test-1  -c backup-restore | grep -i state
time="2026-07-20T06:03:51Z" level=info msg="The cluster state is existing" actor=backup-restore-server

```

3. One of the existing member was restarted . The cluster state was reported as "New". 

```
k logs -n default etcd-test-2  -c backup-restore | grep -i state
time="2026-07-20T06:10:50Z" level=info msg="The cluster state is new" actor=backup-restore-server
```

pkg/server/httpAPI.go
```diff
@@ -484,38 +484,28 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
-		return "", err
-	}
-
-	if state == nil {
-		// Not a Scale-up scenario.
-		// Either a multi-node bootstrap or a restoration of single member in multi-node.
-		m := member.NewMemberControl(h.EtcdConnectionConfig)
-
-		// check whether a learner is present in the cluster
-		// if a learner is present then return `ClusterStateExisting` else `ClusterStateNew`.
```
ishan16696, line 510: may be don't remove these comments, can you please re-add these comments as it's give context.

pkg/server/httpAPI.go
```diff
@@ -484,38 +484,28 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
-		if present, err := m.IsLearnerPresent(ctx); present && err == nil {
-			return miscellaneous.ClusterStateExisting, nil
-		}
+		h.Logger.Warnf("failed to check learner presence, defaulting cluster state to 'new': %v", err)
 		return miscellaneous.ClusterStateNew, nil
-
 	}
+	if present {
+		return miscellaneous.ClusterStateExisting, nil
```
ishan16696:
```suggestion
		// if a learner is present then return `ClusterStateExisting` because to start a learner
		// clusterState must be "existing".
		return miscellaneous.ClusterStateExisting, nil
```

pkg/server/httpAPI.go
```diff
@@ -484,38 +484,28 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
-	client, err := miscellaneous.GetKubernetesClientSetOrError()
-	if err != nil {
-		h.Logger.Warnf("Failed to create clientset: %v", err)
-		return "", fmt.Errorf("failed to get clusterState: %w", err)
-	}
+	// Multi-node case: rely solely on etcd membership state (learner presence)
+	// to decide whether this member is joining an existing cluster or bootstrapping a new one.
+	// StatefulSet status is intentionally NOT consulted here — /config is only requested
+	// after backup-restore initialization, at which point any joining member should
+	// already have been added as a learner.
```
ishan16696: no need to mention about StatefulSet now, rest seems fine as dependency on `StatefulSet` is being removed, in future it will just confuse people that why statefulset is being mentioned in the comment. wdyt ?

ishan16696 (changes requested):
Overall looks good, few nits from my side.

pkg/miscellaneous/miscellaneous.go
```diff
@@ -530,22 +529,6 @@ func IsBackupBucketEmpty(snapStoreConfig *brtypes.SnapstoreConfig, logger *logru
...
-
-	if *etcdSts.Spec.Replicas > 1 && *etcdSts.Spec.Replicas > etcdSts.Status.UpdatedReplicas {
-		logger.Info("etcd statefulset fields", "replicas:", *etcdSts.Spec.Replicas, "updatedReplicas:", etcdSts.Status.UpdatedReplicas)
-		return ptr.To(ClusterStateExisting), nil
-	}
-	return nil, nil
-}
-
 // GetStatefulSet gets the StatefulSet with the name podName in podNamespace namespace. It will return if there is any error or the StatefulSet is not found.
 func GetStatefulSet(ctx context.Context, clientSet client.Client, podNamespace string, podName string) (*appsv1.StatefulSet, error) {
```
CaptainIRS: Could we get rid of this helper? I think the tests in `Etcd Statefulset` section could also be removed.

CaptainIRS: I think `emptyStatefulSet` could also be removed as a consequence.

CaptainIRS (changes requested):
@priyadharshini-151193 thanks for the PR. I've added a couple of comments. PTAL, thanks.

Also, I think we can remove this file: https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/role.yaml and the corresponding rolebinding since etcd-backup-restore doesn't need to have access to the statefulset object.

CaptainIRS:
Follow-up issue in etcd-druid to update the role: https://github.com/gardener/etcd-druid/issues/1387

ishan16696:
> Also, I think we can remove this file: https://github.com/gardener/etcd-backup-restore/blob/master/chart/etcd-backup-restore/templates/role.yaml 

yes, this file is need to be updated but why it's not containing the RBACs for leases 🤔 ?

/cc @anveshreddy18

ishan16696:
@priyadharshini-151193 please also add a appropriate release note.

CaptainIRS:
> yes, this file is need to be updated but why it's not containing the RBACs for leases 🤔 ?

@ishan16696 I was wondering the same, I think since MemberLeaseRenewalEnabled is false by default:

https://github.com/gardener/etcd-backup-restore/blob/10581eb0a9d02f116f57ef10113366f91a16f0b3/pkg/types/health.go#L43

and in the e2e tests, we don't explicitly enable it via CLI flags, it remains disabled. So, we never encountered any issues with permissions when e2e tests are being run. Though, I think if the helm template is supposed to be consumed by community users, we should add lease permissions in the role.

anveshreddy18:
> yes, this file is need to be updated but why it's not containing the RBACs for leases 🤔 ?

The lease object is not created in the helm charts because we don't need both the `snapshot lease`  & `member lease` since there's no etcd-druid to consume when we're deploying etcd-backup-restore standalone, and in etcdbr tests ( where these charts are used ) we're not renewing the lease for the same reason as Rinish has already pointed out. I've probably followed principle of least privilege then 😅 

But we could create these leases & give the access to them if we want to enable their use in community. Only concern is whether we should do it by default or just comment it out in our charts so that they know we provide the capability and they can use them only if they need 🤔

priyadharshini-151193:
> > yes, this file is need to be updated but why it's not containing the RBACs for leases 🤔 ?
> 
> The lease object is not created in the helm charts because we don't need both the `snapshot lease`  & `member lease` since there's no etcd-druid to consume when we're deploying etcd-backup-restore standalone, and in etcdbr tests ( where these charts are used ) we're not renewing the lease for the same reason as Rinish has already pointed out. I've probably followed principle of least privilege then 😅
> 
> But we could create these leases & give the access to them if we want to enable their use in community. Only concern is whether we should do it by default or just comment it out in our charts so that they know we provide the capability and they can use them only if they need 🤔

In the latest commits have removed the RBAC files.  Will add them back once there is an agreement  on the above discussion. 


Also, have refactored the current code a bit to add test cases ( commit - 3258f1c661b45d32bf583f4760960c509517dbea) . There are trade offs but have to proceed with this way to avoid over engineering.

ishan16696:
@anveshreddy18 can you create a separate issue to follow-up on introduction of lease in helm-charts, I think `snapshot leases` aren't required but `member-leases` can be useful in HA setup. It can be helpful for features: `single-member-restoration` and `clusterID mismatch`.

ishan16696 (approved):
LGTM!!

pkg/server/httpAPI.go
```diff
@@ -483,39 +483,31 @@ func (h *HTTPHandler) serveConfig(rw http.ResponseWriter, req *http.Request) {
...
-		// check whether a learner is present in the cluster
-		// if a learner is present then return `ClusterStateExisting` else `ClusterStateNew`.
-		if present, err := m.IsLearnerPresent(ctx); present && err == nil {
-			return miscellaneous.ClusterStateExisting, nil
-		}
+	// check whether a learner is present in the cluster.
+	// if a learner is present then return `ClusterStateExisting` because to start a learner clusterState must be 'existing'.
+	present, err := m.IsLearnerPresent(ctx)
+	if err != nil {
+		h.Logger.Warnf("failed to check learner presence, defaulting cluster state to 'new': %v", err)
```
CaptainIRS, line 504: I think we could return the error to the caller instead of silently returning "new", WDYT? I understand that we're not handling this in the existing code, but I wonder if that is a bug? (Maybe @ishan16696 has some context)

priyadharshini-151193: @CaptainIRS 
I made the change locally to return error and test it . it was failing in 0->3 bootstrap scenario. 

 During a fresh 0→3 bootstrap, etcd is not yet running when this is called — so a context deadline exceeded error is noted as expected. Defaulting to ClusterStateNew is correct here because new is always the safe boot state when we can't determine membership. If the caller returns 500 Internal Server Error on this error and if the caller treat it as fatal and not a warning ,  the etcd wrapper never receives its config file, etcd never starts, and IsLearnerPresent continues to fail — a deadlock
 
 
The function always returns a valid, actionable state. The error from IsLearnerPresent is fully handled at the point it occurs — logged with context, and a safe default applied. There is nothing for the caller to act on. Propagating the error upward currently causes a duplicate log line and implies the caller has a choice it doesn't have. The signature should be getClusterState(...) string — making the contract explicit: this function always gives you a state to boot etcd with.

Sharing my understanding here. Let me know your thoughts

anveshreddy18:
> @anveshreddy18 can you create a separate issue to follow-up on introduction of lease in helm-charts, I think `snapshot leases` aren't required but `member-leases` can be useful in HA setup. It can be helpful for features: `single-member-restoration` and `clusterID mismatch`.

Sure, have created it at https://github.com/gardener/etcd-backup-restore/issues/1041

pkg/server/httpAPI.go
CaptainIRS: Thanks for looking into this. I think we could add a short comment about this for future reference (so that someone looking at the code doesn't assume this is a bug in the future).

Also, just a nit: For the log statement (`h.Logger.Infof("The cluster state is %v", state)`), I think it would be good to move it to this function and have separate specific log statements before the returns. This is because, if you try deleting a pod (simple restart scenario), when it comes back up, or when you unhibernate a cluster, you'll see the log statement "The cluster state is new", which I feel is a bit misleading. WDYT?

priyadharshini-151193: Thanks @CaptainIRS for your review. In the recent commit , added a comment to indicate the deadlock scenario on 0->3 bootstrap scenario. 

Also updated the logs to highlight why the cluster state was set to a certain value 

Sample logs from testing 


**Single member cluster**


```
k logs -n default etcd-test-0  -c backup-restore | grep -i state
time="2026-07-24T08:24:09Z" level=info msg="The cluster size is 1. The cluster state is new" actor=backup-restore-server
```

**1->3 scale out** 


```
 k logs -n default etcd-test-1  -c backup-restore | grep -i state
time="2026-07-24T08:25:57Z" level=info msg="Learner is present. The cluster state is existing" actor=backup-restore-server

```


**0->3 scale out** 


```
k logs -n default etcd-test-0  -c backup-restore | grep -i state
time="2026-07-24T08:29:00Z" level=warning msg="failed to check learner presence, defaulting cluster state to 'new': error listing members: context deadline exceeded" actor=backup-restore-server
```

```
 k logs -n default etcd-test-1  -c backup-restore | grep -i state
time="2026-07-24T08:29:00Z" level=warning msg="failed to check learner presence, defaulting cluster state to 'new': error listing members: context deadline exceeded" actor=backup-restore-server
```

```
k logs -n default etcd-test-2  -c backup-restore | grep -i state
time="2026-07-24T08:29:00Z" level=warning msg="failed to check learner presence, defaulting cluster state to 'new': error listing members: context deadline exceeded" actor=backup-restore-server
```

priyadharshini-151193: Since we are returning a specific state in every scenario and error is always nil , updated the method signature  of getclusterstate() to only return a string ( state) . This contract  works fine with the current requirements. May be in future , might need refactoring if there is a new requirement. Hence the tradeoff.

CaptainIRS: Thanks for updating the logs.

I've tested out the unhibernation scenario as well:
```
etcd-test-1 backup-restore time="2026-07-24T09:02:45Z" level=warning msg="failed to check learner presence, defaulting cluster state to 'new': error listing members: context deadline exceeded" actor=backup-restore-server
```

The method signature change also looks good to me.

CaptainIRS (approved):
@priyadharshini-151193 thanks for addressing the review comments. LGTM!

Issue #1044, opened 2026-07-29
Restore/compaction fails when full snapshot's actual revision exceeds the first delta's last revision


CaptainIRS:
/area backup
/area disaster-recovery
/kind bug

**What happened**:

Restoration (and compaction, which restores internally) fails at the first delta snapshot:

```
level=fatal msg="Failed to compact snapshot: unable to restore snapshots during compaction: mismatched event revision while applying delta snapshot, expected 3607435063 but applied 3607439350"
```

A full snapshot's filename revision is obtained from a separate `Get` call taken *before* `client.Snapshot()`. Because the two are not atomic, the snapshot's true internal (mvcc) revision can be higher than the revision encoded in its filename. This is a design flaw on our side, not an etcd limitation. `applyFirstDeltaSnapshot` tries to compensate for the resulting overlap between the full snapshot and the first delta, but only when the full snapshot's true revision lands *within* the first delta's range. When the true revision exceeds the first delta's last revision, restoration fails.

Artifacts in the failure:
- Full snapshot filename revision: `3607431063` (file name: `Full-00000000-3607431063-1784752476.gz`)
- First delta `Incr-3607431064-3607435063`, so its last revision is `3607435063`
- Full snapshot's actual internal revision: `3607435351`, confirmed via snapshot status:

```
$ etcdctl snapshot status Full-00000000-3607431063-1784752476 -wtable
+----------+------------+------------+------------+
|   HASH   |  REVISION  | TOTAL KEYS | TOTAL SIZE |
+----------+------------+------------+------------+
| 5fa3d9c4 | 3607435351 |     182845 |     2.7 GB |
+----------+------------+------------+------------+
```

The actual revision (`3607435351`) is 288 beyond the first delta's last revision (`3607435063`).

Trace with the actual revision `3607435351`:

The skip guard is not taken, because it checks for exact equality (`3607435351 == 3607435063` is false):

https://github.com/gardener/etcd-backup-restore/blob/debd95bd77352280dada2a9a94efa9d1d20bbb78/pkg/snapshot/restorer/restorer.go#L511

The trim loop looks for the first event whose ModRevision is greater than the store's current revision (`3607435351`). Every event in the delta has a ModRevision of at most `3607435063`, which is less than `3607435351`, so the condition never holds and the loop never breaks:

[...]

PR #1047, opened 2026-07-29
Fix restoration failures + verification condition when full snapshot and next delta have overlapping events.

anveshreddy18:

/area backup
/area robustness
/kind bug

This PR fixes a restoration failure issue that showed up during our regular compaction job occurrence which caught a particular issue where the applied revision on the etcd was coming out to be higher than the expected revision which caused failures in restoration.

The issue is also covered in the issue opened at #1044 , basically it's because of the non-atomicity between the get call that records the full snapshot's revision and the actual `Maintenance.Snapshot()` call that streams the db while taking a full snapshot, so by the time we note down the revision `R` , the snapshot db would have actually increased and streamed more events to the snapshot, so it might contain `R+k` revisions in actual but the full snapshot recorded revision in it's name contains only `R` like `Full-0-R-<timestamp>`. 

This PR also improves the verification of restoration by fixing an issue with error handling while restoring where currently the error is as shown below:

https://github.com/gardener/etcd-backup-restore/blob/debd95bd77352280dada2a9a94efa9d1d20bbb78/pkg/snapshot/restorer/restorer.go#L590-L592

But the problem with that is it expects the `etcdRevision/curRevision` of the etcd db to be always the same as the `LastRevision` of the most recently applied delta snapshot, if not, it throws a validation error to fail the restoration. But the verification needs to be improved because the following revisions are possible:

Snapshotting
  ════════════

    full            in store  →  Full-0-R-T₀
    ●───────────────●                 (recorded end revision = R)
    0            R + K  ← actual db content reaches R+K, but only R is recorded

    Delta-1 (D₁)    in store  →  Incr-(R+1)-R₁-T₁
    ●───────────────●                 no. of events = R₁ - R
    R+1            R₁

    Delta-2 (D₂)    in store  →  Incr-(R₁+1)-R₂-T₂
    ●───────────────●                 no. of events = R₂ - R₁
    R₁+1           R₂

  Restoring
  ═════════

    ①  ← full      →  rev = R + K
    ②  ← Delta-1   →  rev = (R + K) + (no. of events in D₁)
                          = R̶ + K + R₁ - R̶
                          = R₁ + K
    ③  ← Delta-2   →  rev = (R₁ + K) + (no. of events in D₂)
                          = R̶₁ + K + R₂ - R̶₁
                          = R₂ + K      ← actual revision after restore

So in conclusion, the actual revision (R2+K) is always >= R2 ( as K>=0), meaning => `etcdRevision >= snap.LastRevision` always. If not, then that's an issue with the verification. The new verification reflects the decision from the above math.

Fixes part of #1044 

**Release note**:

```bugfix operator
fix a restoration failure & verification mechanism that could occur during restoration when the full snapshot's recorded revision lagged the actual data it contained, causing the first delta snapshot to be incorrectly re-applied and the revision verification to fail.
```

### DIFF
--- a/pkg/snapshot/restorer/restorer.go
+++ b/pkg/snapshot/restorer/restorer.go
@@ -508,26 +508,18 @@ func (r *Restorer) applyFirstDeltaSnapshot(clientKV client.KVCloser, snap *brtyp
 	}
 	lastRevision := resp.Header.Revision
 
-	if lastRevision == snap.LastRevision {
-		// there is no need to apply this fist delta snapshot
-		// as it's completely overlaps with full snapshot data.
+	// remove all events that are already present in the restored database as it can be at, or even beyond the first delta's LastRevision
+	newEvents := FilterEventsAfterRevision(events, lastRevision)
+	if len(newEvents) == 0 {
+		// there is nothing to apply.
 		// please refer: https://github.com/gardener/etcd-backup-restore/issues/844
-		r.logger.Infof("First delta snapshot %s found to be completely overlap with full snapshot with db revisions: %d", path.Join(snap.SnapDir, snap.SnapName), lastRevision)
-		r.logger.Info("Skipping this delta snapshot...")
+		r.logger.Infof("First delta snapshot %s completely overlaps with the full snapshot (db revision: %d); skipping it.", path.Join(snap.SnapDir, snap.SnapName), lastRevision)
 		return nil
-	}
-
-	var newRevisionIndex int
-	for index, event := range events {
-		if event.EtcdEvent.Kv.ModRevision > lastRevision {
-			newRevisionIndex = index
-			break
-		}
 	}
 
 	r.logger.Infof("Applying first delta snapshot %s", path.Join(snap.SnapDir, snap.SnapName))
 
-	return applyEventsToEtcd(clientKV, events[newRevisionIndex:])
+	return applyEventsToEtcd(clientKV, newEvents)
 }
 
 func persistRawDeltaSnapshot(rc io.ReadCloser, tempFilePath string) error {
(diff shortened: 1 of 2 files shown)


pkg/snapshot/restorer/restorer.go
```diff
@@ -587,8 +596,11 @@ func verifySnapshotRevision(clientKV client.KVCloser, snap *brtypes.Snapshot) er
 		return fmt.Errorf("failed to connect to etcd KV client: %v", err)
 	}
 	etcdRevision := getResponse.Header.GetRevision()
-	if snap.LastRevision != etcdRevision {
-		return fmt.Errorf("mismatched event revision while applying delta snapshot, expected %d but applied %d ", snap.LastRevision, etcdRevision)
+	// The embedded etcd's revision must be at least the snapshot's LastRevision. It can also be higher when the snapshot fully overlaps data already present in the restored database
+	// eg: a full snapshot whose recorded revision lagged the actual data it contained, so the first delta snapshot is entirely contained in the full snapshot.
+	// So if the etcdRevision is less than than the applied snapshot's LastRevision, it then indicates missing events and is treated as an error.
+	if etcdRevision < snap.LastRevision {
```
CaptainIRS: This should not be `<` because the goal of restoration is to get the correct state after applying N revisions on top of a full snapshot. If we relax the verification rule, it would be impossible to figure out if something went wrong during the restoration process. We will see that restoration was successful, but it would either have duplicate revisions or wrong data.

CaptainIRS (changes requested):
@anveshreddy18 thanks for the PR. I think we need a different approach because, currently in this PR we're just relaxing the criteria for correct restoration in `verifySnapshotRevision`, which will allow duplicate/wrong data to be applied and still report success.

PR #1058, opened 2026-08-13
DEP-08: Prevent removed etcd members from rejoining during scale-in

seshachalam-yv:
/area high-availability
/area robustness
/area control-plane
/kind enhancement

This PR adds an anti-rejoin guard for removed etcd members.

During scale-in of a multi-node etcd cluster, `etcd-druid` removes an etcd member before the corresponding pod/PVC cleanup has fully completed. If that pod restarts with its old data directory, `etcd-backup-restore` can otherwise interpret the state as a scale-out/rejoin case and re-add the removed member as a learner. This can cause a remove/re-add loop. Refer [DEP-08](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md#anti-rejoin-guard)

The guard runs before the learner-add path and checks whether the local member ID is present in etcd's local `members_removed` tombstone bucket. If the member was permanently removed, initialization stops and the member is not re-added.

The local member ID is resolved from:

1. the Kubernetes member Lease
2. a member-id file persisted on the PV by the heartbeat

The member-id file gives the guard a local fallback when the Kubernetes Lease is unavailable, for example in self-hosted or temporary API-unavailable scenarios.

Fixes #
Part of https://github.com/gardener/etcd-druid/pull/1425

This change supports safe scale-in for multi-node etcd clusters only.

**Release note**:
```feature operator
Added a membership removal check to etcd-backup-restore that prevents a scaled-in etcd member from re-adding itself as a learner after etcd-druid removes it during a scale-in operation.
```

### DIFF
--- a/docs/concepts/anti-rejoin-guard.md
+++ b/docs/concepts/anti-rejoin-guard.md
@@ -2,7 +2,7 @@
 
 If etcd cluster is managed by [etcd-druid](https://github.com/gardener/etcd-druid) and when a multi-node etcd cluster is [scaled in](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md), `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
 
-Without a guard, `etcd-backup-restore` interprets the state *"this pod has local etcd data, but its member ID is not present in the live cluster"* as a scale-out case and re-adds the removed member as a learner. `etcd-druid` removes it again, the same pod re-adds itself, and the cluster enters a remove/re-add loop.
+Without a guard, `etcd-backup-restore` interprets the state *"this pod has local etcd data, but its member ID is not present in the live cluster"* as a scale-out case and re-adds the removed member as a learner to the etcd cluster. `etcd-druid` find out that member hasn't been removed yet (still present in cluster), hence it removes the cluster member again, but now again the backup-restore of member pod re-adds the removed member, and the cluster enters a remove/re-add loop.
 
 The anti-rejoin guard closes this loop. On a multi-node cluster it runs on every startup **before** the membership and scale-up checks, whenever the data directory already holds an etcd member tree (`member/`, `wal/`, `snap/`). It runs regardless of whether the member lease is still alive — a stale lease must not be allowed to bypass the tombstone check. If the local member was explicitly removed from the cluster, startup stops with `ErrMemberPermanentlyRemoved` instead of re-adding the member.
 
--- a/docs/concepts/anti-rejoin-guard.md
+++ b/docs/concepts/anti-rejoin-guard.md
@@ -22,17 +22,15 @@ If the resolved local member ID is present in `members_removed`, the cluster has
 
 ```mermaid
 flowchart TD
-    Start[backup-restore init, multi-node] --> Data{Data dir has member/wal/snap tree?}
-    Data -->|No| Fresh[Fresh PVC: skip guard, continue]
+    Start[backup-restore init, multi-node] --> Data{Data dir has valid structure + valid files}
+    Data -->|No| Fresh[No prior data: skip guard, continue]
     Data -->|Yes| Lease{Member lease has ID?}
     Lease -->|Yes| Have[Local member ID]
     Lease -->|No / unavailable| File{member-id file exists?}
     File -->|No| NoID[No ID: treat as fresh join, continue]
     File -->|Yes| Have
-    Have --> DB{boltdb backend exists?}
-    DB -->|No| Failed[Partial deletion: ErrMembershipCheckFailed, fail closed]
-    DB -->|Yes| Open[Open boltdb read-only]
-    Open -->|Open/read fails or corrupt| Failed
+    Have --> Open[Open boltdb read-only]
+    Open -->|Open/read fails or corrupt| Failed[Data inconsistency: ErrMembershipCheckFailed, fail closed]
     Open -->|OK| Removed{Own ID in members_removed?}
     Removed -->|No| Continue[Not removed, continue normal init]
     Removed -->|Yes| Stop[ErrMemberPermanentlyRemoved, stop]
@@ -44,9 +42,9 @@ The check is deliberately conservative and **fails closed** — when it cannot b
 
 - If the data directory state **cannot be determined** (the `member/`, `wal/`, `snap/` structure check itself errors), the guard fails closed.
 - If **no local member ID** can be resolved (neither lease nor member-id file), there is nothing to check, so normal initialization continues.
-- If a local member ID **is** resolved but the **boltdb backend file is missing**, the WAL tree is present but the db is not — a possible partial PVC deletion. The guard cannot safely treat this as a fresh member, so it fails closed.
-- If the boltdb exists but **cannot be opened** (the read-only open is bounded by a lock-acquisition timeout to ride out transient lock contention) or is **corrupt** (bolt panics are recovered and surfaced as errors), the guard fails closed. The process then exits and is restarted by the crash-back-off loop.
+- If a local member ID **is** resolved but the **boltdb backend cannot be opened** (file missing, lock-acquisition timeout, or corrupt — bolt panics are recovered and surfaced as errors), the data directory is in an inconsistent state. The guard cannot safely proceed, so it fails closed. The process then exits and is restarted by the crash-back-off loop.
 - Only the **local member's own ID** is considered. Entries for other removed members are ignored.
+- If the guard passes (the member ID is **not** in `members_removed`), no action is taken and normal initialization continues as-is — the existing membership and restoration flow is unchanged.
 
 Opening boltdb **read-only** ensures the guard cannot corrupt a live backend, and reading only the small `members_removed` bucket keeps runtime and memory overhead negligible. This follows the same access pattern — including recovering from bolt panics on a corrupt backend — already used by `etcd-backup-restore`'s data validator (`getLatestEtcdRevision`).
 
--- a/docs/concepts/anti-rejoin-guard.md
+++ b/docs/concepts/anti-rejoin-guard.md
@@ -1,6 +1,6 @@
 # Anti-rejoin guard
 
-When a multi-node etcd cluster is scaled in, `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
+If etcd cluster is managed by [etcd-druid](https://github.com/gardener/etcd-druid) and when a multi-node etcd cluster is [scaled in](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md), `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
 
 Without a guard, `etcd-backup-restore` interprets the state *"this pod has local etcd data, but its member ID is not present in the live cluster"* as a scale-out case and re-adds the removed member as a learner. `etcd-druid` removes it again, the same pod re-adds itself, and the cluster enters a remove/re-add loop.
 
--- a/docs/concepts/anti-rejoin-guard.md
+++ b/docs/concepts/anti-rejoin-guard.md
@@ -1,0 +1,61 @@
+# Anti-rejoin guard
+
+When a multi-node etcd cluster is scaled in, `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
+
+Without a guard, `etcd-backup-restore` interprets the state *"this pod has local etcd data, but its member ID is not present in the live cluster"* as a scale-out case and re-adds the removed member as a learner. `etcd-druid` removes it again, the same pod re-adds itself, and the cluster enters a remove/re-add loop.
+
+The anti-rejoin guard closes this loop. On a multi-node cluster it runs on every startup **before** the membership and scale-up checks, whenever the data directory already holds an etcd member tree (`member/`, `wal/`, `snap/`). It runs regardless of whether the member lease is still alive — a stale lease must not be allowed to bypass the tombstone check. If the local member was explicitly removed from the cluster, startup stops with `ErrMemberPermanentlyRemoved` instead of re-adding the member.
+
+This feature backs [etcd-druid DEP-08 (Scaling-in a multi-node etcd cluster)](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md).
+
+## How a removed member is detected
+
+etcd records removed member IDs in the local boltdb backend's `members_removed` bucket when `MemberRemove` is applied. This is the local member's tombstone. The guard inspects two on-disk artefacts of the local etcd data directory:
+
+- The **local member ID**, resolved in priority order:
+  - **Member lease** — the heartbeat writes the member ID into the lease holder identity as `<memberID-hex>:<clusterID-hex>:<role>` on every renewal. Reading it is a single cheap API call, so it is tried first.
+  - **Member-id file** (`<data-dir>/member-id`) — the heartbeat also writes the same holder-identity string to this file on every renewal, *before* patching the lease. The write-before-patch ordering ensures the file is always at least as fresh as the lease. The guard falls back to this file when the lease is genuinely absent (`NotFound`); other lease errors (RBAC denial, network timeout, unparseable holder identity) are logged and also fall back to the file, so a transient API blip does not permanently block the member.
+  - If neither source yields an ID (fresh member, no lease yet, no file yet), the guard treats the pod as a fresh join and continues normally.
+- The **`members_removed` bucket** in the boltdb backend (`<data-dir>/member/snap/db`), opened read-only.
+
+If the resolved local member ID is present in `members_removed`, the cluster has explicitly removed this member and the sidecar must not re-add it.
+
+```mermaid
+flowchart TD
+    Start[backup-restore init, multi-node] --> Data{Data dir has member/wal/snap tree?}
+    Data -->|No| Fresh[Fresh PVC: skip guard, continue]
+    Data -->|Yes| Lease{Member lease has ID?}
+    Lease -->|Yes| Have[Local member ID]
+    Lease -->|No / unavailable| File{member-id file exists?}
+    File -->|No| NoID[No ID: treat as fresh join, continue]
+    File -->|Yes| Have
+    Have --> DB{boltdb backend exists?}
+    DB -->|No| Failed[Partial deletion: ErrMembershipCheckFailed, fail closed]
+    DB -->|Yes| Open[Open boltdb read-only]
+    Open -->|Open/read fails or corrupt| Failed
+    Open -->|OK| Removed{Own ID in members_removed?}
+    Removed -->|No| Continue[Not removed, continue normal init]
+    Removed -->|Yes| Stop[ErrMemberPermanentlyRemoved, stop]
+```
+
+## Design decisions
+
+The check is deliberately conservative and **fails closed** — when it cannot be sure a member is safe to re-add, it stops startup rather than risk re-joining a removed member:
+
+- If the data directory state **cannot be determined** (the `member/`, `wal/`, `snap/` structure check itself errors), the guard fails closed.
+- If **no local member ID** can be resolved (neither lease nor member-id file), there is nothing to check, so normal initialization continues.
+- If a local member ID **is** resolved but the **boltdb backend file is missing**, the WAL tree is present but the db is not — a possible partial PVC deletion. The guard cannot safely treat this as a fresh member, so it fails closed.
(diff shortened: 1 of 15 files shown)


docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
+# Anti-rejoin guard
+
+When a multi-node etcd cluster is scaled in, `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
```
ishan16696: can you also provide a DEP link to `scaled-in` and also mention that it's only application when backup-restore is managed by etcd-druid

```suggestion
If etcd cluster is managed by [etcd-druid](https://github.com/gardener/etcd-druid) and when a multi-node etcd cluster is [scaled in](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md), `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
```

pkg/member/antirejoin.go
```diff
@@ -0,0 +1,67 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
```
ishan16696:
```suggestion
// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```

pkg/member/antirejoin.go
```diff
@@ -0,0 +1,67 @@
...
+	// ErrMemberPermanentlyRemoved is returned when the local member ID is found
+	// in etcd's members_removed tombstone bucket.
+	ErrMemberPermanentlyRemoved = errors.New("member was permanently removed from the cluster; refusing to re-add as learner")
+
+	// ErrMembershipCheckFailed is returned when the tombstone could not be
+	// inspected. Fails closed on this error.
+	ErrMembershipCheckFailed = errors.New("failed to inspect local membership tombstone")
+)
+
+const antiRejoinBoltDBTimeout = 10 * time.Second
```
ishan16696, line 27: 10 seconds seems to be very less as db size can be larger which can take more than 10sec.

pkg/member/antirejoin_test.go
```diff
@@ -0,0 +1,116 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
```
ishan16696:
```suggestion
// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```

pkg/etcdutil/dir.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
```
ishan16696:
```suggestion
// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+
+This feature backs [etcd-druid DEP-08 (Scaling-in a multi-node etcd cluster)](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md).
+
+## How a removed member is detected
+
+etcd records removed member IDs in the local boltdb backend's `members_removed` bucket when `MemberRemove` is applied. This is the local member's tombstone. The guard inspects two on-disk artefacts of the local etcd data directory:
+
+- The **local member ID**, resolved in priority order:
+  - **Member lease** — the heartbeat writes the member ID into the lease holder identity as `<memberID-hex>:<clusterID-hex>:<role>` on every renewal. Reading it is a single cheap API call, so it is tried first.
```
ishan16696, line 16: what will happened if lease update is turn off (in-case of backup-restore is consumed via helm-charts or via independently).

ishan16696 (changes requested):

docs/concepts/anti-rejoin-guard.md
seshachalam-yv: The doc already documents the fallback chain clearly at line 17:

1. Member lease — tried first
2. Member-id file (<data-dir>/member-id) — fallback when lease is absent or errors
3. Neither — treated as fresh join, continues normally

pkg/member/antirejoin.go
seshachalam-yv: The `antiRejoinBoltDBTimeout` is not a read/IO timeout — it is the file-lock acquisition timeout passed to bolt.Open. As you can see in the bbolt implementation: https://github.com/gardener/etcd-backup-restore/blob/master/vendor/go.etcd.io/bbolt/db.go#L1223-L1226 — it only controls how long we wait to acquire the file lock if another process already holds it.

The actual read touches only the small `members_removed` bucket, which is negligible regardless of db size.

`IsMemberRemoved` returns an error, `WasPermanentlyRemoved` returns `ErrMembershipCheckFailed`, and the container exits — the `crashloopbackoff` loop then restarts it with exponential backoff, which is the intended behaviour for the fail-closed path.

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
+# Anti-rejoin guard
+
+If etcd cluster is managed by [etcd-druid](https://github.com/gardener/etcd-druid) and when a multi-node etcd cluster is [scaled in](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md), `etcd-druid` removes an etcd member from the cluster before the corresponding StatefulSet pod is terminated. In the short window between the member removal and the pod termination, the pod can still restart with its old data directory.
+
+Without a guard, `etcd-backup-restore` interprets the state *"this pod has local etcd data, but its member ID is not present in the live cluster"* as a scale-out case and re-adds the removed member as a learner. `etcd-druid` removes it again, the same pod re-adds itself, and the cluster enters a remove/re-add loop.
```
ishan16696:
```suggestion
Without a guard, `etcd-backup-restore` interprets the state *"this pod has local etcd data, but its member ID is not present in the live cluster"* as a scale-out case and re-adds the removed member as a learner to the etcd cluster. `etcd-druid` find out that member hasn't been removed yet (still present in cluster), hence it removes the cluster member again, but now again the backup-restore of member pod re-adds the removed member, and the cluster enters a remove/re-add loop.
```

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+
+The anti-rejoin guard closes this loop. On a multi-node cluster it runs on every startup **before** the membership and scale-up checks, whenever the data directory already holds an etcd member tree (`member/`, `wal/`, `snap/`). It runs regardless of whether the member lease is still alive — a stale lease must not be allowed to bypass the tombstone check. If the local member was explicitly removed from the cluster, startup stops with `ErrMemberPermanentlyRemoved` instead of re-adding the member.
```
ishan16696, line 7: If DODs or anyone wants to remove a member and re-adds a member from a cluster (may be because scale-out or bootstrap wasn't successful) then how do remove a member or trigger single member restoration ?
Do we have to remove the data-dir for that ? ... may be please capture this in ops-guide.

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+
+This feature backs [etcd-druid DEP-08 (Scaling-in a multi-node etcd cluster)](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md).
+
+## How a removed member is detected
```
ishan16696, line 11: can you also capture that why we have abandon the wal based approach (mention about the memory spike) ...may be DEP itself, so that we remember it.

pkg/member/identity.go
```diff
@@ -0,0 +1,140 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
```
ishan16696:
```suggestion
// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```

pkg/member/identity_test.go
```diff
@@ -0,0 +1,322 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
```
ishan16696:
```suggestion
// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```

ishan16696 (changes requested):

docs/concepts/anti-rejoin-guard.md
seshachalam-yv: The anti-rejoin guard only runs on multi-node clusters — single-member restoration is completely unaffected.

For the case where an SRE or operator manually removes a member (outside of a scale-in), the guard does not differentiate — it sees the tombstone in members_removed and blocks the pod with ErrMemberPermanentlyRemoved, causing it to crashloop. The recovery path in that case is to scale in (remove the member's PVC/pod) and then scale out again (let etcd-druid provision a fresh member with a clean data directory). Since the guard only triggers when there is prior etcd data on the PV, a fresh PVC bypasses it entirely.

Will add a note about this in the ops-guide section.

docs/concepts/anti-rejoin-guard.md
seshachalam-yv: I will update the DEP itself.

pkg/etcdutil/dir.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package etcdutil
+
+import "path/filepath"
+
+const (
+	// MemberIDFileName is the name of the file that caches the member identity on the PV.
```
ishan16696:
```suggestion
	// MemberIDFileName is the name of the file that persist the etcd cluster member identity on the PV.
```
may be "persist" is more correct word here.

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+  - **Member-id file** (`<data-dir>/member-id`) — the heartbeat also writes the same holder-identity string to this file on every renewal, *before* patching the lease. The write-before-patch ordering ensures the file is always at least as fresh as the lease. The guard falls back to this file when the lease is genuinely absent (`NotFound`); other lease errors (RBAC denial, network timeout, unparseable holder identity) are logged and also fall back to the file, so a transient API blip does not permanently block the member.
+  - If neither source yields an ID (fresh member, no lease yet, no file yet), the guard treats the pod as a fresh join and continues normally.
+- The **`members_removed` bucket** in the boltdb backend (`<data-dir>/member/snap/db`), opened read-only.
+
+If the resolved local member ID is present in `members_removed`, the cluster has explicitly removed this member and the sidecar must not re-add it.
+
+```mermaid
+flowchart TD
+    Start[backup-restore init, multi-node] --> Data{Data dir has member/wal/snap tree?}
```
ishan16696:
```suggestion
    Start[backup-restore init, multi-node] --> Data{Data dir has valid structure + valid files }
```
this will include bootstrap + single member restoration case.

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+    Data -->|No| Fresh[Fresh PVC: skip guard, continue]
```
ishan16696: wdyt mean by fresh PVC here ? In case of single member restoration there will not be any fresh PVC.

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+    Data -->|Yes| Lease{Member lease has ID?}
+    Lease -->|Yes| Have[Local member ID]
+    Lease -->|No / unavailable| File{member-id file exists?}
+    File -->|No| NoID[No ID: treat as fresh join, continue]
+    File -->|Yes| Have
+    Have --> DB{boltdb backend exists?}
```
ishan16696: This check shouldn't exist here as you have already checked the validity of data-dir structure above ?

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+    DB -->|No| Failed[Partial deletion: ErrMembershipCheckFailed, fail closed]
```
ishan16696: what is Partial deletion, if "db file can't be opened or got corrupted" then it's single-member restoration case in HA and restoration case in case of non-HA.

docs/concepts/anti-rejoin-guard.md
```diff
@@ -0,0 +1,61 @@
...
+    Open -->|OK| Removed{Own ID in members_removed?}
+    Removed -->|No| Continue[Not removed, continue normal init]
+    Removed -->|Yes| Stop[ErrMemberPermanentlyRemoved, stop]
+```
+
+## Design decisions
+
+The check is deliberately conservative and **fails closed** — when it cannot be sure a member is safe to re-add, it stops startup rather than risk re-joining a removed member:
+
+- If the data directory state **cannot be determined** (the `member/`, `wal/`, `snap/` structure check itself errors), the guard fails closed.
```
ishan16696, line 43: can you also mentioned that once the guard fails closed then normal initialization process continues as it is.

ishan16696 (changes requested):

docs/concepts/anti-rejoin-guard.md
seshachalam-yv: Single-member restoration from an HA cluster where the data directory has been wiped/restored. there is nothing to check, so the guard skips and normal initialization proceeds.

pkg/etcdutil/dir.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
```
ishan16696, line 3: FYI: because of this PR: https://github.com/gardener/etcd-backup-restore/pull/1060, this CopyrightText has been changed now, please run `make add-license-headers`.

pkg/health/heartbeat/heartbeat.go
```diff
@@ -141,6 +141,17 @@ func (hb *Heartbeat) RenewMemberLease(ctx context.Context) error {
 		renewedMemberLease.Annotations[k] = v
 	}
 
+	// Persist the member ID to the PV before updating the lease so the file is
+	// at least as fresh as the lease and the guard can resolve the ID even when
+	// the lease API is temporarily unreachable.
```
ishan16696, line 146: I as a consumer of backup-restore can turned this off or decrease the freq. of such writes ? 
I feel the side effect of such writes can be increase in disk usage as backup-restore also start making freq disk writes which are as freq as lease update. I'm not very comfortable with this.

pkg/member/antirejoin.go
```diff
@@ -0,0 +1,67 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
```
ishan16696, line 3: same as above and rest of places.

docs/concepts/anti-rejoin-guard.md
ishan16696: this is not yet resolved.

ishan16696 (changes requested):
overall looks good, I will do the testing now with druid setup.

docs/concepts/anti-rejoin-guard.md
seshachalam-yv: > If **no local member ID** can be resolved (neither lease nor member-id file), there is nothing to check, so normal initialization continues.

https://github.com/gardener/etcd-backup-restore/pull/1058/changes#diff-7222d19469feab2afb36615e8428d39ebe77aad28325b23d054cb0c907a88d36R43-R44

Already mentioned in next line.

pkg/health/heartbeat/heartbeat.go
seshachalam-yv: It’s a very small write, the file is only ~30 bytes and is overwritten atomically, not appended to. With the current 30s lease-renewal interval, that’s at most 2 writes/minute, or roughly 60 bytes/minute.

Given the size and frequency, I don’t expect this to have any noticeable impact on backup-restore disk usage or I/O.

pkg/member/antirejoin.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-backup-restore/pull/1058/commits/eac7b5733e5ca7786a3f7f925105c8423920fe33 .

pkg/etcdutil/dir.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-backup-restore/pull/1058/commits/eac7b5733e5ca7786a3f7f925105c8423920fe33.

pkg/health/heartbeat/heartbeat.go
ishan16696: I'm thinking can we do it in a exponential way ? because scale-in feature is not a very frequently used feature, that's why I'm thinking we can even avoid these extra writes or we can reduce them using exponential backoff. wdyt ?

pkg/member/identity.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/sirupsen/logrus"
+	coordinationv1 "k8s.io/api/coordination/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// WriteMemberIDFile atomically writes the holder identity string to the
+// member-id file in the data directory.
+func WriteMemberIDFile(dataDir, holderIdentity string) error {
+	path := etcdutil.MemberIDFilePath(dataDir)
```
ishan16696, line 25: Why we are creating a `member-id` file inside dir: "data/new.etcd", why not we are creating `member-id` file on the same path as `safe_guard` file and `temp` dir exists ? I would prefer the `member-id` file shouldn't be inside dir: `new.etcd`.
```
└── data
    ├── new.etcd
    │   ├── member
    │   │   ├── snap
    │   │   │   └── db
    │   │   └── wal
    │   │       ├── 0000000000000000-0000000000000000.wal
    │   │       └── 0.tmp
    │   └── member-id
    ├── safe_guard
    └── temp
```

pkg/health/heartbeat/heartbeat.go
seshachalam-yv: We should support the self-hosted case as well. In that setup, the member Lease may be unavailable for some time. If the guard depends on the Lease during that period, the etcd member could get stuck and the self-hosted cluster could end up in a deadlock. Therefore, the guard should rely on the persisted `member-id` file as the source of truth for resolving the local member ID.

ishan16696 (commented):
while testing this PR, I have this doubt:

pkg/initializer/validator/datavalidator.go
```diff
@@ -237,6 +235,13 @@ func (d *DataValidator) hasEtcdDirectoryStructure() (bool, error) {
 	return memberExist && snapExist && walExist, nil
 }
 
+// HasPriorData reports whether the data directory contains an etcd member tree
+// (member/, wal/, snap/ sub-directories), indicating this PV was previously used
+// by a running etcd member.
+func (d *DataValidator) HasPriorData() (bool, error) {
+	return d.hasEtcdDirectoryStructure()
+}
```
ishan16696: what's the point of this wrapper function, why not just directly call has dir: `hasEtcdDirectoryStructure()`

pkg/initializer/initializer.go
```diff
@@ -56,6 +56,24 @@ func (e *EtcdInitializer) Initialize(mode validator.Mode) error {
 
 		m := member.NewMemberControl(e.Config.EtcdConnectionConfig)
 
+		// Always run the anti-rejoin guard when prior data is present on the PV,
+		// regardless of whether the member lease is still alive. A stale lease
+		// must not bypass the tombstone check.
+		hasData, err := e.Validator.HasPriorData()
```
ishan16696: hasPriorData isn't checking the inside db file and all, it's checking checking the validates the data-dir structure, so name is kind-of misleading to me.

pkg/member/member_control.go
```diff
@@ -384,10 +390,43 @@ func (m *memberControl) WasMemberInCluster(ctx context.Context, clientSet client
...
 	}
 
-	if memberLease.Spec.HolderIdentity == nil {
-		return false
+	return memberLease.Spec.HolderIdentity != nil
+}
+
+// WasPermanentlyRemoved checks whether this member was permanently removed from
+// the etcd cluster during a prior scale-in by inspecting the boltdb
+// members_removed tombstone bucket. It resolves the local member ID from the
```
ishan16696:
```suggestion
// "members_removed" tombstone bucket. It resolves the local member ID from the
```

ishan16696 (changes requested):
few more nits:

pkg/member/identity.go
ishan16696: this comment has been addressed now.
```
├── member-id
├── new.etcd
│   └── member
│       ├── snap
│       │   └── db
│       └── wal
│           ├── 0000000000000000-0000000000000000.wal
│           └── 0.tmp
├── safe_guard
└── temp
```

ishan16696 (approved):
Thanks @seshachalam-yv for the PR and addressing the comments.

PR #1062, opened 2026-08-31
fix: defragments the leader blocking all cluster writes for the duration

### LINKED ISSUE #1061: long defragment causing controllers failed to obtain theirs leases
/area etcd
/kind bug

**What happened**:
A scheduled etcd storage defragmentation on our cluster caused a ~40-second etcd stall. All Kubernetes controllers lost their leader election leases simultaneously, resulting in a brief but complete control-plane disruption. The cluster self-recovered after the defragmentation completed.
During this process:
	etcd-main-2 member taken 21s to defrag DB from 3.84 GB to 2.12 GB.
	then etcd-main-1 taken 17s to defrag DB from 3.84 GB to 2.12 GB.
         last leader etcd-main-0 taken 17s and finished defragmentation.

It should not breaking our controller pods to obtain theirs leases during defragmentation.

	1. Etcd with large DB (maybe around 4GB).
	2. Start defragment.
        3. Pods begin require lease during defragment.

yuehaii:

/area etcd
/kind bug

When DefragmentData defragments a multi-member etcd cluster, it defragments the leader, blocking all cluster writes for the duration. The new --move-leader flag was added to transfer leadership to a follower before defragmenting the former leader.

Fixes #1061

> % go build ./... 2>&1 && go test ./pkg/etcdutil/... 2>&1
> ok      github.com/gardener/etcd-backup-restore/pkg/etcdutil    2.177s
> ?       github.com/gardener/etcd-backup-restore/pkg/etcdutil/client     [no test files]
> 

**Release note**:

```other operator
add move-leader parameter to transfer etcd leadership before defragmenting the leader member to avoid cluster-wide write stalls.
```

### DIFF
--- a/pkg/etcdutil/etcdutil.go
+++ b/pkg/etcdutil/etcdutil.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
@@ -224,57 +225,8 @@ func DefragmentData(defragCtx context.Context, clientMaintenance client.Maintena
 	}
 
 	// Optionally transfer leadership before defragmenting the leader to keep writes flowing.
-	// MoveLeader must be sent to the current leader, so use the first known leader endpoint.
-	leaderTransferred := false
-	if moveLeader && len(followerEtcdEndpoints) > 0 && len(leaderEtcdEndpoints) > 0 {
-		func() {
-			ctx, cancel := context.WithTimeout(defragCtx, brtypes.DefaultEtcdConnectionTimeout)
-			defer cancel()
-			membersInfo, err := clientCluster.MemberList(ctx)
-			if err != nil {
-				logger.Warnf("failed to list members for leadership transfer: %v, defragmenting leader in place", err)
-				return
-			}
-			// Pick the first voting non-learner follower as the transfer target.
-			var transfereeID uint64
-			for _, m := range membersInfo.Members {
-				if m.GetID() != leaderID && !m.GetIsLearner() {
-					transfereeID = m.GetID()
-					break
-				}
-			}
-			if transfereeID == 0 {
-				logger.Warnf("no eligible voting follower found for leadership transfer, defragmenting leader in place")
-				return
-			}
+	transferLeadership(defragCtx, clientMaintenance, clientCluster, clientFactory, leaderEtcdEndpoints, followerEtcdEndpoints, leaderID, logger, moveLeader)
 
-			leaderEndpoint := leaderEtcdEndpoints[0]
-			func() {
-				ctx, cancel := context.WithTimeout(defragCtx, brtypes.DefaultLeaderTransferTimeout)
-				defer cancel()
-				// clientFactory is used to create a maintenance client pinned to the leader endpoint for the
-				// MoveLeader call, ensuring the request is never rejected with ErrNotLeader.
-				leaderMaintenance, err := clientFactory.NewMaintenanceForEndpoint(leaderEndpoint)
-				if err != nil {
-					logger.Warnf("failed to create maintenance client for leader endpoint %s: %v, defragmenting leader in place", leaderEndpoint, err)
-					return
-				}
-				defer leaderMaintenance.Close()
-				if _, err := leaderMaintenance.MoveLeader(ctx, transfereeID); err != nil {
-					logger.Warnf("failed to transfer etcd leadership from %s to member %d: %v, defragmenting leader in place", leaderEndpoint, transfereeID, err)
-				} else {
-					logger.Infof("etcd leadership transferred from %s to member %d, will now defragment the former leader", leaderEndpoint, transfereeID)
-					leaderTransferred = true
-				}
-			}()
-		}()
-	}
-
-	if leaderTransferred {
-		logger.Info("Starting the defragmentation on former etcd leader (now follower)")
-	} else {
-		logger.Info("Starting the defragmentation on etcd leader")
-	}
 	// Perform the defragmentation on etcd leader (or former leader after transfer).
 	for _, ep := range leaderEtcdEndpoints {
 		if err := func() error {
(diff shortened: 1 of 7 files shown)


pkg/etcdutil/etcdutil.go
```diff
@@ -214,8 +223,59 @@ func DefragmentData(defragCtx context.Context, clientMaintenance client.Maintena
...
+	if moveLeader && len(followerEtcdEndpoints) > 0 && len(leaderEtcdEndpoints) > 0 {
+		func() {
+			ctx, cancel := context.WithTimeout(defragCtx, brtypes.DefaultEtcdConnectionTimeout)
+			defer cancel()
+			membersInfo, err := clientCluster.MemberList(ctx)
+			if err != nil {
+				logger.Warnf("failed to list members for leadership transfer: %v, defragmenting leader in place", err)
+				return
+			}
+			// Pick the first voting non-learner follower as the transfer target.
```
ishan16696: why are you picking the first voting member to transfer leadership ? 
Is that follower(first voting member) will always be ahead of the other follower ?

pkg/etcdutil/etcdutil.go
```diff
@@ -214,8 +223,59 @@ func DefragmentData(defragCtx context.Context, clientMaintenance client.Maintena
 		}
 	}
 
-	logger.Info("Starting the defragmentation on etcd leader")
-	// Perform the defragmentation on etcd leader.
+	// Optionally transfer leadership before defragmenting the leader to keep writes flowing.
+	// MoveLeader must be sent to the current leader, so use the first known leader endpoint.
+	leaderTransferred := false
+	if moveLeader && len(followerEtcdEndpoints) > 0 && len(leaderEtcdEndpoints) > 0 {
```
ishan16696: Please create a separate function for all these logic, let's not mix all actions here in `func DefragmentData`

ishan16696 (changes requested):

pkg/etcdutil/etcdutil.go
yuehaii: they all client URLs of the same leader node and belong to the same single leader member. 
let me try to pickup the member with highest [RaftIndex](https://etcd.io/docs/v3.3/dev-guide/api_reference_v3/) which is the current raft committed index of the responding member.

ishan16696: > let me try to pickup the member with highest [RaftIndex](https://etcd.io/docs/v3.3/dev-guide/api_reference_v3/) which is the current raft committed index of the responding member

yes, I will also suggest that it will minimize the data loss.

ishan16696:
Hi @yuehaii ,
This moveLeader approach is fundamentally incompatible with how backup-restore orchestrates defragmentation in an HA setup. Please refer to this [doc](https://github.com/gardener/etcd-backup-restore/blob/99399fce25f7f9903d2237a1d635ae6b7667cdc1/docs/operations/leader_election.md#defragmentation)
Basically, defragmentation process for all etcd cluster members is orchestrated by the `backup-restore leader`, which is always the `sidecar` co-located with the `etcd leader`. Calling a `MoveLeader` transfers etcd leadership to a different member, which simultaneously transfers backup-restore leadership away from the current instance. That cancels the parent-context under which the defrag loop is running, because of this the former etcd leader, the very member that still needs to be defragmented, never gets defragmented. The fix would defeat itself.

Let me thought through some work around if possible without any big change as for the long-run we already have a plan to fix the entire design of defrag.

yuehaii:
> Hi @yuehaii , This moveLeader approach is fundamentally incompatible with how backup-restore orchestrates defragmentation in an HA setup. Please refer to this [doc](https://github.com/gardener/etcd-backup-restore/blob/99399fce25f7f9903d2237a1d635ae6b7667cdc1/docs/operations/leader_election.md#defragmentation) Basically, defragmentation process for all etcd cluster members is orchestrated by the `backup-restore leader`, which is always the `sidecar` co-located with the `etcd leader`. Calling a `MoveLeader` transfers etcd leadership to a different member, which simultaneously transfers backup-restore leadership away from the current instance. That cancels the parent-context under which the defrag loop is running, because of this the former etcd leader, the very member that still needs to be defragmented, never gets defragmented. The fix would defeat itself.
> 
> Let me thought through some work around if possible without any big change as for the long-run we already have a plan to fix the entire design of defrag.

Do you mean that after the leader transition, the previous leader(now-follower) will cancel [leCtx](https://github.com/gardener/etcd-backup-restore/blob/99399fce25f7f9903d2237a1d635ae6b7667cdc1/pkg/leaderelection/leaderelection.go#L92C42-L92C47) which is needed by the subsequent defragment loop?
This is a known [issue](https://github.com/etcd-io/etcd/discussions/20961) in the etcd community. Their solution uses a standalone external process (a pure gRPC [client](https://github.com/ahrtr/etcd-defrag/blob/main/utils.go#L11C2-L11C10)) that has zero dependency on which node is the etcd leader. When [MoveLeader](https://github.com/ahrtr/etcd-defrag/blob/6863205df6b8d6eced2489984710d98990e62332/agent.go#L257C13-L257C23) returns, the external tool keeps running and issues Defragment against the now-follower, its execution context is never affected. I hope this helps with refactoring the defragment loop.

ishan16696:
> Do you mean that after the leader transition, the previous leader(now-follower) will cancel [leCtx](https://github.com/gardener/etcd-backup-restore/blob/99399fce25f7f9903d2237a1d635ae6b7667cdc1/pkg/leaderelection/leaderelection.go#L92C42-L92C47) which is needed by the subsequent defragment loop?

yes, it cancels the parent context which closes all the loops started by backup-restore leader.

yuehaii:
> yes, it cancels the parent context which closes all the loops started by backup-restore leader.

I [refactored](https://github.com/gardener/etcd-backup-restore/pull/1062/changes/4fac0507117af69eba0dbae1e9b3e941c2e75371) the defrag loop. to solve this issue, transferLeadership issues all RPCs on a context.WithoutCancel derived context so a concurrent leCtx cancellation cannot abort the transfer mid-flight. 
back in DefragmentData, the former-leader defrag step is also detached via context.WithoutCancel, but only when the transfer actually succeeded, preserving normal shutdown behaviour on all other paths.

yuehaii:
hi @ishan16696 , good day. could you please review the pull request again? I have refactor the code to prevent the context been canceled during leader transfer and drag loop.

